PR #1, opened 2016-01-21
Device simulator ui

michahirsch:
* Extend device simulator to create DDI and DMF simulated devices
* Enhance simulator with an Vaadin UI to manage simulated devices


michahirsch:
close pull request due wrong commit between

PR #10, opened 2016-01-25
Reonnect queues to vhost

SirWayne:
If the RabbitMQ goes down, all queues will loose connection. The queue will stop trying to reconnect after 3 failures. Now the queues will try to reconnect all the time. To Disable the feature set the property missing.queues.fatal=true


kaizimmerm:
Please add a description what this pull request all about. Internal ticket numbers should be removed from pull request (title)

PR #11, opened 2016-01-25
Mecs-1440_target_count_not_displayed

asharani-murugesh:


kaizimmerm:
Target count not displayed when user has no update and delete target permission
Expected behavior : Display target count label if user has target read permissions

PR #12, opened 2016-01-26
Create Unit Test for all filter enum fields

SirWayne:
Create Unit tests for the new operation and filter fields in RSQL search query language:

### DIFF
--- a/hawkbit-core/src/main/java/org/eclipse/hawkbit/repository/ActionStatusFields.java
+++ b/hawkbit-core/src/main/java/org/eclipse/hawkbit/repository/ActionStatusFields.java
@@ -18,10 +18,6 @@ package org.eclipse.hawkbit.repository;
 public enum ActionStatusFields implements FieldNameProvider {
 
     /**
-     * The type field.
-     */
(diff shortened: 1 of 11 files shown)


kaizimmerm:
Please fix the merge conflicts first.

kaizimmerm:
@michahirsch any comments?

michahirsch:
As I can see the tests only verifying the count of the result of the RSQL query language. I'm not sure if this is totally sufficient. The count might be correct but the actual content of the result could be wrong. 

Maybe we could add like when we are searching e.g. for action with a specific status that the actions in the result only containing this status?

What you think?

SirWayne:
Yes that is right, that could be extend as well as a combination of different operators, but this tests should first check the enum search fields.If someone adjusts the entity model that immediately notice, that the search query is broken e.g. ipAdress.

michahirsch:
@kaizimmerm  then I would merge these tests as it is. This covers already more than we have now and we can extend them step by step. The pull-request already have 855 additions.
Thanks

PR #17, opened 2016-02-01
Wrong predicate for boolean filter

SirWayne:
- Fix conversion for boolean for sql query. If boolean is not wellformed a rsql syntax error is thrown
- Escape % for sql query#
- Remove operator like ("=li=")
- Extend not equal operator with not like jpa query


michahirsch:
@SirWayne  there are test failures of this pull-request.

michahirsch:
+1, @kaizimmerm can you double check and merge it.

PR #19, opened 2016-02-02
Fix overwrite anonymous property

michahirsch:
- removing the `@Value` annotations from the property bean and replace with inner classes and `@ConfigurationProperties` annotation


michahirsch:
@kaizimmerm  please give +1 and merge this.

PR #21, opened 2016-02-04
NPE_when_user_tries_to_create_simple_targetfilter.

venu1278:
Display target tag add icon only when user has sufficient permissions.
Signed-off-by: venu1278 <venugopal.boodidadinne@in.bosch.com>


kaizimmerm:
@venu1278 : please remove internal ticket numbers from pull request title and describe in title and description what the pull request is all about,

michahirsch:
One questions, there `config` component button is created nevertheless if the user has the permission we just don't add it to the layout. 
```
 config = SPUIComponentProvider.getButton(SPUIComponetIdProvider.ADD_TARGET_TAG, "", "", "", true,FontAwesome.COG, SPUIButtonStyleSmallNoBorder.class);
```
Questions if this is necessary to create vaadin-components if we will never "use" or show up? Maybe would be more efficient if only the vaadin-components are created which are really will be added to the UI?

PR #25, opened 2016-02-04
Fix mysql empty in statement

michahirsch:
- adding empty list check when creating targets. MySql does not handly empty in-statements in SQL
- adding flush on entity manager because the persisted Rollout entity needs the ID set, because creating the rolloutgroup and reference the not yet persisted Rollout because entitymanager is in auto-flushmode and will flush automatically during the `findAllTargets`

This should fix `RolloutManagamentTest` which creates empty targets in tests.


michahirsch:
:+1:

PR #32, opened 2016-02-10
Drag and drop handles wrong item

venu1278:
Say an item-A is selected in a list (e.g. softwareModules in the View:Upload). Then user drags item-B with 
- MouseSelectDown(item-B),
- MouseDragTo(DeleteSoftware) - _one can see item-B moving!_
- MouseSelectRelease

After pressing button /actions/ the item-A is presented to be deleted (expected was: item-B).


michahirsch:
closing due IP Validation fails. Please check and re-create pull-request again.

PR #35, opened 2016-02-11
Test fix

asharani-murugesh:
Safely getting access to the UI thread to modify the target count label

asharani-murugesh <asharani.murugesh@in.bosch.com>


michahirsch:
already on master, closing.

PR #41, opened 2016-02-13
Custom target filter bug fixes

asharani-murugesh:
Bug fixes:
* Custom filter detail view : Using =IN= parameter throws 'Unsupported operator' exception and breaks the UI
* Custom filter search : type any query.Search never gets finished.
* IN and OUT parameter : case insensitive search


hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/management/dstable/DistributionAddUpdateWindowLayout.java
```diff
@@ -344,22 +344,21 @@ private void setDistributionValues(final DistributionSet distributionSet, final
...
-            notificationMessage.displayValidationError(
-                    i18n.get("message.duplicate.dist", new Object[] { existingDs.getName(), existingDs.getVersion() }));
 
-            return false;
-        } else {
+        if (existingDs == null) {
             return true;
         }
+
+        if (editDistribution && !existingDs.getId().equals(editDistId)) {
```
michahirsch: Is this really correct, I've seen it this way already because this duplicate check method is used not only for editing also for creating so I guess we don't need to check the boolean _editDistribution_ here?

hawkbit-repository/src/main/java/org/eclipse/hawkbit/repository/TargetManagement.java
```diff
@@ -205,7 +205,7 @@ public Long countTargetsAll() {
     @PreAuthorize(SpringEvalExpressions.HAS_AUTH_READ_TARGET)
     public Slice<Target> findTargetsAll(@NotNull final TargetFilterQuery targetFilterQuery,
             @NotNull final Pageable pageable) {
-        return findTargetsAll(RSQLUtility.parse(targetFilterQuery.getQuery(), TargetFields.class), pageable);
+        return findTargetsAll(RSQLUtility.parse(targetFilterQuery.getQuery().toLowerCase(), TargetFields.class), pageable);
```
michahirsch: I would prefer to put the target filter string  `toLowerCase()` to the `RSQLUtility`, then we don't need to call the `toLowerCase()` in each method, because I think all queries by filter string are case-insensitive isn't it?

PR #45, opened 2016-02-17
DistributionSetAssignmentResult consistency fix

kaizimmerm:
* DistributionSetAssignmentResult contains actions that do not reflect the status on the DB. Needs to be avoided by returning actions IDs only and force the user to query an updated action entity.
* In addition we saw that we have a bug in Action repository as the entitygraph does not collect target.assignedDs. That was fixed also.
* while at it I also removed some expensive entitygraph usage that is in fact not necessary.


kaizimmerm:
@michahirsch please merge

hawkbit-repository/src/main/java/org/eclipse/hawkbit/repository/DeploymentManagement.java
```diff
@@ -392,7 +394,6 @@ private DistributionSetAssignmentResult assignDistributionSetToTargets(@NotNull
         entityManager.detach(set);
 
         // send distribution set assignment event
-
         targets.stream().filter(t -> !!!targetIdsCancellList.contains(t.getId()))
```
kaizimmerm: refactor: extract to method with meaning full name instead of comment

kaizimmerm:
You are right, is fixed.

PR #46, opened 2016-02-18
removing spring-context-support dependency because it's not necessary.

michahirsch:
* remove spring-context-support dependency because the dependency seems not necessary for hawkbit.


michahirsch:
closing pull-request due the spring-context-support is necessary because of the `GuavaCacheManager` it's working only because of transitive dependency to `spring-data-redis`

PR #48, opened 2016-02-21
Removed runtime dependencies to sass compiler package

kaizimmerm:
Fixed accidental introduced runtime dependencies to sass compiler. Allows applications to remove sass compiler from delivery.
Fixed a few sonar issues on the way.


hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/components/SPUIComponentProvider.java
```diff
@@ -109,7 +109,7 @@ public static HorizontalLayout getHeaderLayout(
         // Do we really need this???
         HorizontalLayout hLayout = getHorizontalLayout(new SPUIHorizontalLayout().getUiHorizontalLayout().getClass());
         try {
-            SPUIHeaderLayoutDecorator layoutDecorator = null;
+            SPUIHeaderLayoutDecorator layoutDecorator;
```
SirWayne: early return

PR #58, opened 2016-02-29
Add description for asserts in test cases

SirWayne:
Add description for asserts in test cases and use dimaond operator


michahirsch:
:+1: @kaizimmerm  can be merged, this are only test adaptions for better assertion readability.

PR #59, opened 2016-03-02
introduce a push strategy to the HawkbitUI

michahirsch:
* introduced a push strategy interface to the hawkBitUI which is initialized and cleanup by the life-cycle of the Vaadin UI
* an implementation for delayed event push which retrieves repository events from the guava-eventbus and dispatch them to the vaadin-session-event-bus every 2 seconds. The events are dispatched in the Vaadin access-thread and so there are not 1:1 dispatched directly otherwise this will overload the Vaadin access-thread and will freeze the UI. So the `DelayedEventBusPushStrategy` is consuming the events into a queue and dispatch them every 2 seconds at once.

### DIFF
--- a/hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/push/DelayedEventBusPushStrategy.java
+++ b/hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/push/DelayedEventBusPushStrategy.java
@@ -1,0 +1,245 @@
+/**
+ * Copyright (c) 2015 Bosch Software Innovations GmbH and others.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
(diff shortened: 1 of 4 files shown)


hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/push/DelayedEventBusPushStrategy.java
```diff
@@ -0,0 +1,245 @@
...
+     * @param userContext
+     *            the security context of the current session
+     * @param event
+     *            the event to dispatch to the UI
+     * @return {@code true} if the event can be dispatched to the UI otherwise
+     *         {@code false}
+     */
+    protected boolean eventSecurityCheck(final SecurityContext userContext,
+            final org.eclipse.hawkbit.eventbus.event.Event event) {
+        if (userContext != null && userContext.getAuthentication() != null) {
```
SirWayne: early return

michahirsch:
@kaizimmerm  please have a look and merge it.

PR #62, opened 2016-03-03
fix retrieving the wrong message converter

michahirsch:
*  The BaseAmqpService retrieves the wrong message converter `SimpleMessageConverter` and cannot convert messages containing JSON payload


michahirsch:
@kaizimmerm please have a look and merge it.

PR #64, opened 2016-03-03
Fixed wrong status content generation in mgmt API.

kaizimmerm:
* Action status entries in management API show wrong ID and reportedAt. Currently we present ID and reportedAt not of the status entry itself, but of the related action which is obviously wrong.

* Fixed sorting of action status by ID. We used sorting params but actually fed this into a repo method with fixed desc sorting.

* While at it I added also sorting by reportAt.

* Changed default sort of actions in target resource also to descending. Makes more sense from usability point of view and matches behaviour of mgmt. UI.

### DIFF
--- a/hawkbit-rest-resource/src/test/java/org/eclipse/hawkbit/rest/resource/TargetResourceTest.java
+++ b/hawkbit-rest-resource/src/test/java/org/eclipse/hawkbit/rest/resource/TargetResourceTest.java
@@ -887,7 +887,7 @@ public class TargetResourceTest extends AbstractIntegrationTest {
         // descending order
         mvc.perform(get(RestConstants.TARGET_V1_REQUEST_MAPPING + "/" + knownTargetId + "/"
                 + RestConstants.TARGET_V1_ACTIONS + "/" + action.getId() + "/" + RestConstants.TARGET_V1_ACTION_STATUS)
-                        .param(RestConstants.REQUEST_PARAMETER_SORTING, "REPORTED_AT:DESC"))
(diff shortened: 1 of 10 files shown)


hawkbit-rest-resource/src/test/java/org/eclipse/hawkbit/rest/resource/TargetResourceTest.java
```diff
@@ -920,6 +1026,8 @@ private String generateStatusreferenceLink(final String knownTargetId, final Act
         final List<Target> updatedTargets = deploymentManagement.assignDistributionSet(one, targets)
                 .getAssignedTargets();
         // 2nd update
+        // sleep 1ms to ensure that we can sort by reportedAt
+        Thread.sleep(1);
```
michahirsch: This sleep could also be removed by using the `AuditingHandler`

By setting an `DateTimeProvider` on the `AuditingHandler` for this test and switch back to the old one again
```
  @Autowired
    private AuditingHandler auditingHandler;
    // set for this test a new date time provider
    auditingHandler.setDateTimeProvider(myDateTimeProvider);
   // switch back the old one again
   auditingHandler.setDateTimeProvider(CurrentDateTimeProvider.INSTANCE);
```
I guess a 1ms sleep is sufficient just  worry that maybe a 1ms is not enough when using `System.currentMillis` is not updated every millisecond maybe.
```
public static long currentTimeMillis()
Returns the current time in milliseconds. Note that while the unit of time of the return value is a millisecond, the granularity of the value depends on the underlying operating system and may be larger. For example, many operating systems measure time in units of tens of milliseconds
```

PR #75, opened 2016-03-08
Create unit test for the json amqp message conversion

SirWayne:
Create unit test for the json amqp message conversion


Jonkno:
+1

PR #78, opened 2016-03-09
Fixed broken sorting for SM/DS type in mgmt API

kaizimmerm:
* Sorting for the two types did not work at all…blame on me it seems I messed that up somewhere last year :smile: 
* SM type field MAXASSIGNMENTS was in filter/sort actually called differently ("MAX"). I fixed that also as sort filter params should always have the same name as the field in the mgmt. API payloads. I mean how should an API use ever figure this out if we use different names.

### DIFF
--- a/hawkbit-rest-resource/src/main/java/org/eclipse/hawkbit/rest/resource/PagingUtility.java
+++ b/hawkbit-rest-resource/src/main/java/org/eclipse/hawkbit/rest/resource/PagingUtility.java
@@ -12,10 +12,12 @@ import org.eclipse.hawkbit.repository.ActionFields;
 import org.eclipse.hawkbit.repository.ActionStatusFields;
 import org.eclipse.hawkbit.repository.DistributionSetFields;
 import org.eclipse.hawkbit.repository.DistributionSetMetadataFields;
+import org.eclipse.hawkbit.repository.DistributionSetTypeFields;
 import org.eclipse.hawkbit.repository.RolloutFields;
 import org.eclipse.hawkbit.repository.RolloutGroupFields;
 import org.eclipse.hawkbit.repository.SoftwareModuleFields;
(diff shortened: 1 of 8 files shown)


michahirsch:
code seems ok but the circle CI build failed because of test-failure.

kaizimmerm:
Circle is green?

hawkbit-rest-resource/src/main/java/org/eclipse/hawkbit/rest/resource/PagingUtility.java
```diff
@@ -69,6 +71,17 @@ static Sort sanitizeSoftwareModuleSortParam(final String sortParam) {
         return sorting;
     }
 
+    static Sort sanitizeSoftwareModuleTypeSortParam(final String sortParam) {
+        final Sort sorting;
+        if (sortParam != null) {
```
SirWayne: if you early return you need no local variable and no else

   if (sortParam == null) {
         return new Sort(Direction.ASC, SoftwareModuleTypeFields.NAME.getFieldName());
      } 
        return  new Sort(SortUtility.parse(SoftwareModuleTypeFields.class, sortParam));

hawkbit-rest-resource/src/main/java/org/eclipse/hawkbit/rest/resource/PagingUtility.java
```diff
@@ -80,6 +93,17 @@ static Sort sanitizeDistributionSetSortParam(final String sortParam) {
         return sorting;
     }
 
+    static Sort sanitizeDistributionSetTypeSortParam(final String sortParam) {
+        final Sort sorting;
+        if (sortParam != null) {
```
SirWayne: see above

SirWayne:
+1

PR #79, opened 2016-03-10
Optimize simulator throughput

kaizimmerm:
- Removed the old 1 sec sleep and instead raised the scheduler random time range a bit


SirWayne:
+1

PR #81, opened 2016-03-11
Stabilize rollout tests

kaizimmerm:
- Reduced issues around the async rollout mgmt. API test on slow CI infrastructure.
- Ensures that also rollout repos are cleared after every test.


SirWayne:
+1

PR #84, opened 2016-03-12
Migrated hardcoded doc links to configurable setup.

kaizimmerm:
Currently we have a hardcoded legacy for documentation links offered from within Management UI.

- I migrated this to a configurable setup that is now also optional on the login view and in the navigation.
- Example app shows how this can be leveraged.
- While at it I also changed the defaults for the demo link (now disabled) but showed in the example app how this can be used.

However, we definitely have to invest more work in the management UI to provide more help links.


hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/tenantconfiguration/TenantConfigurationDashboardView.java
```diff
@@ -101,7 +104,9 @@ private HorizontalLayout saveConfigurationButtonsLayout() {
         undoConfigurationBtn.addClickListener(event -> undoConfiguration());
         hlayout.addComponent(undoConfigurationBtn);
 
-        final Link linkToSystemConfigHelp = DocumentationPageLink.SYSTEM_CONFIGURATION_VIEW.getLink();
+        final Link linkToSystemConfigHelp = SPUIComponentProvider
+                .getHelpLink(uiProperties.getLinks().getDocumentation().getSystemConfigurationView());
```
SirWayne: 2 * ;

PR #95, opened 2016-03-16
fix wrong '-' character signs

kartben:
Sorry, there was one more fix it seems!


michahirsch:
Thanks @kartben for fixing this, happend probably due copy and paste from console or somewhere.

kartben:
Yup. I hate when that happens!

And you're welcome :) Trying to get some time to go through the device
simulator demo and stuff...!

PR #104, opened 2016-03-23
Remove boilerplate code

kaizimmerm:
- Removed REST model boilerplate that was historically used for Swagger generation.
- Tenant aware and non tenant aware entities are now based on the same BaseEntity.


michahirsch:
Ok a lot of changes, but in general looks ok removing all the extra object response bodies and use a `List`or `PagedList` instead. Have you tried the `examples/hawkbit-mgmt-api-client` if it still works, should be, but...

SirWayne:
:+1:

kaizimmerm:
I tested the mgmt api client scenarios. Thx for the hint.

PR #105, opened 2016-03-24
Bug/target table perfomance

SirWayne:
1. Add a event provider which hold the repository events which are provided by the ui.

2. Fix the performance issue. Issue was the selection of targets. The table
needs very long to calculate the scroll position. After deletion of a target the table row the scroll postion is not set anymore.


hawkbit-autoconfigure/src/main/java/org/eclipse/hawkbit/autoconfigure/ui/UIAutoConfiguration.java
```diff
@@ -37,4 +36,14 @@ public DistributedResourceBundleMessageSource messageSource() {
         return new DistributedResourceBundleMessageSource();
     }
 
+    /**
+     * A event provider bean which hold the supported events for the UI.
+     * 
+     * @return the provider bean
+     */
+    @Bean
```
michahirsch: Why we adding the bean into the `AutoConfiguration` if you create the `HawkbitEventProvider` by our own with `new` in the constructor of the example app? I guess we don't need to provide the autoconfiguration for this, do we? Or otherwise we should annotate it with `@ConditionalMissingBean`

SirWayne: You're right. thanks

michahirsch:
please resolve the conflicts so the branch can be merged with the master, thanks.

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/HawkbitEventProvider.java
```diff
@@ -0,0 +1,59 @@
...
+import org.eclipse.hawkbit.eventbus.event.TargetTagCreatedBulkEvent;
+
+/**
+ * The default hawkbit event provider.
+ */
+public class HawkbitEventProvider implements UIEventProvider {
+
+    private static final Set<Class<? extends Event>> SINGLE_EVENTS = new HashSet<>();
+
+    static {
```
kaizimmerm: 1 static block
fix hashset size

kaizimmerm:
I have taken the liberty to merge this for you.

PR #112, opened 2016-03-30
Reduced log severity of invalid incoming messages to warning.

kaizimmerm:
We should not log on error what cannot be influenced by us.


michahirsch:
I restarted the circleCI build due a test failure which is not related to this pull-request, weird.

PR #114, opened 2016-03-31
Fixed mixed polling time configuration.

kaizimmerm:


hawkbit-core/src/main/java/org/eclipse/hawkbit/tenancy/configuration/DurationHelper.java
```diff
@@ -37,7 +37,7 @@ private DurationRangeValidator(final Duration min, final Duration max) {
         }
 
         public boolean isWithinRange(final Duration duration) {
```
michahirsch: Please add javadoc to the method and describe the behavior about greater and not greaterthanequal

PR #119, opened 2016-04-04
Reduce some duplicate code

SirWayne:
Reduce duplicate code


michahirsch:
Like the abstract autowired things, this was not possible in early times, cannot remember why either the beans are not correctly instantiated in the UI or the spring-framework didn't do it before when having abstract class to call the `@PostConstruct` and autowiring fields. :+1:

PR #120, opened 2016-04-05
DDI supports actifact download through http and anonyomous

michahirsch:
* fix typo of class _TenantSecurityToken_
* extend _TenantSecurityToken_ to allow to set the file-resource send by the download-server how to lookup the file on update-server side.
* introduce permission and configuration to download anonymously in the tenant configuration
* extend system configuration UI to enable/disable anonymous download configuration


hawkbit-dmf-amqp/src/main/java/org/eclipse/hawkbit/amqp/AmqpMessageHandlerService.java
```diff
@@ -207,6 +213,22 @@ private Message handleAuthentifiactionMessage(final Message message) {
         return getMessageConverter().toMessage(authentificationResponse, messageProperties);
     }
 
+    private LocalArtifact findLocalArtifactByFileResource(final FileResource fileResource) {
+        LocalArtifact localArtifact = null;
```
SirWayne: I prefer direct return without a local variable. You're doing nothing with the local variable, so you can remove it

PR #132, opened 2016-04-08
Eclipselink tenancy

kaizimmerm:
I removed the the multi tenancy exceptions. Non of them was actually needed anymore. One exception was the the transaction that was opened by the tenant cache key gen Bean but that one should not have a transaction in the first place. I removed it.

### DIFF
--- a/hawkbit-repository/src/main/java/org/eclipse/hawkbit/MultiTenantJpaTransactionManager.java
+++ b/hawkbit-repository/src/main/java/org/eclipse/hawkbit/MultiTenantJpaTransactionManager.java
@@ -38,7 +38,7 @@ public class MultiTenantJpaTransactionManager extends JpaTransactionManager {
         super.doBegin(transaction, definition);
 
         // ignore transactions on tenant independent calls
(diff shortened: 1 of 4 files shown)


kaizimmerm:
Ups by accident I added a commit that has nothing to do with the topic. Mea culpa but I think it does not hurt to have this fixed also.

michahirsch:
`rolloutScheduler` is not necessary anymore. We are iterating over all tenants because we are not able to retrieve data for multiple tenants at once. So this is ok to remove.

hawkbit-repository/src/main/java/org/eclipse/hawkbit/MultiTenantJpaTransactionManager.java
```diff
@@ -38,38 +37,31 @@
     protected void doBegin(final Object transaction, final TransactionDefinition definition) {
         super.doBegin(transaction, definition);
 
+        // ignore transactions on tenant independent calls
```
SirWayne: For this comment you can write a method, which explains the code e.g. `isTenantIndependent`

PR #139, opened 2016-04-12
fix the lookup of artifacts which exists multiple times with same lookup key e.g. SHA1

michahirsch:
Artifacts can assigned to multiple software modules which leads that when we trying to find a artifact e.g. based on filename or SHA1 hashes which are not unique in the system we cannot determine the correct software module to it to verify if the current target is permitted to download this artifact or not. 

Current code just takes the first found `LocalArtifact` and checks against this `SoftwareModule` which will lead in case the artifact exists multiple times in different `SoftwareModule`s the current check to verify if this `SoftwareModule` is assigned to the target will fail, because it could be another `SoftwareModule` but with the same artifact.

To avoid this check, we can check if the requested artifact had been ever assigned to the target based on the action history and check all actions and their `SoftwareModule`s and if in these `SoftwareModule`s exists the request artifact.

### DIFF
--- a/hawkbit-dmf-amqp/src/main/java/org/eclipse/hawkbit/amqp/AmqpMessageHandlerService.java
+++ b/hawkbit-dmf-amqp/src/main/java/org/eclipse/hawkbit/amqp/AmqpMessageHandlerService.java
@@ -166,7 +166,7 @@ public class AmqpMessageHandlerService extends BaseAmqpService {
             final LocalArtifact localArtifact = findLocalArtifactByFileResource(fileResource);
 
             if (localArtifact == null) {
-                LOG.info("target {} requested file resource which does not exists to download",
+                LOG.info("target {} requested file resource {} which does not exists to download",
                         secruityToken.getControllerId(), fileResource);
                 throw new EntityNotFoundException();
             }
--- a/hawkbit-dmf-amqp/src/main/java/org/eclipse/hawkbit/amqp/AmqpMessageHandlerService.java
+++ b/hawkbit-dmf-amqp/src/main/java/org/eclipse/hawkbit/amqp/AmqpMessageHandlerService.java
@@ -205,23 +205,32 @@ public class AmqpMessageHandlerService extends BaseAmqpService {
         return getMessageConverter().toMessage(authentificationResponse, messageProperties);
     }
 
+    /**
+     * check action for this download purposes, the method will throw an
+     * EntityNotFoundException in case the controller is not allowed to download
+     * this file because it's not assigned to an action and not assigned to this
+     * controller. Otherwise no controllerId is set = anonymous download
+     * 
+     * @param secruityToken
+     *            the security token which holds the target ID to check on
+     * @param localArtifact
+     *            the local artifact to verify if the given target is allowed to
+     *            download this artifact
+     */
     private void checkIfArtifactIsAssignedToTarget(final TenantSecurityToken secruityToken,
             final LocalArtifact localArtifact) {
-        // check action for this download purposes, the method will throw an
-        // EntityNotFoundException in case the controller is not allowed to
-        // download this file because it's not assigned to an action and not
-        // assigned to this controller. Otherwise no controllerId is set =
-        // anonymous download
-        if (secruityToken.getControllerId() != null) {
-            LOG.debug("no anonymous download request, doing authentication check for target {} and artifact {}",
-                    secruityToken.getControllerId(), localArtifact);
-            if (!controllerManagement.hasTargetArtifactAssigned(secruityToken.getControllerId(), localArtifact)) {
-                LOG.info("target {} tried to download artifact {} which is not assigned to the target");
-                throw new EntityNotFoundException();
-            }
-            LOG.info("download security check for target {} and artifact {} granted", secruityToken.getControllerId(),
-                    localArtifact);
+        final String controllerId = secruityToken.getControllerId();
+        if (controllerId == null) {
+            LOG.info("anonymous download no authentication check for artifact {}", localArtifact);
+            return;
         }
+        LOG.debug("no anonymous download request, doing authentication check for target {} and artifact {}",
+                controllerId, localArtifact);
+        if (!controllerManagement.hasTargetArtifactAssigned(controllerId, localArtifact)) {
+            LOG.info("target {} tried to download artifact {} which is not assigned to the target");
+            throw new EntityNotFoundException();
+        }
+        LOG.info("download security check for target {} and artifact {} granted", controllerId, localArtifact);
     }
 
     private LocalArtifact findLocalArtifactByFileResource(final FileResource fileResource) {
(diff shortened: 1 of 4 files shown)


hawkbit-dmf-amqp/src/main/java/org/eclipse/hawkbit/amqp/AmqpMessageHandlerService.java
```diff
@@ -213,6 +205,25 @@ private Message handleAuthentifiactionMessage(final Message message) {
...
     }
 
+    private void checkIfArtifactIsAssignedToTarget(final TenantSecurityToken secruityToken,
+            final LocalArtifact localArtifact) {
+        // check action for this download purposes, the method will throw an
+        // EntityNotFoundException in case the controller is not allowed to
+        // download this file because it's not assigned to an action and not
+        // assigned to this controller. Otherwise no controllerId is set =
+        // anonymous download
+        if (secruityToken.getControllerId() != null) {
```
SirWayne: early return

hawkbit-dmf-amqp/src/main/java/org/eclipse/hawkbit/amqp/AmqpMessageHandlerService.java
```diff
@@ -213,6 +205,25 @@ private Message handleAuthentifiactionMessage(final Message message) {
         return getMessageConverter().toMessage(authentificationResponse, messageProperties);
     }
 
+    private void checkIfArtifactIsAssignedToTarget(final TenantSecurityToken secruityToken,
+            final LocalArtifact localArtifact) {
+        // check action for this download purposes, the method will throw an
```
SirWayne: In my opionion is javadoc here better than a inline comment

hawkbit-dmf-amqp/src/main/java/org/eclipse/hawkbit/amqp/AmqpMessageHandlerService.java
```diff
@@ -213,6 +205,25 @@ private Message handleAuthentifiactionMessage(final Message message) {
...
+        // EntityNotFoundException in case the controller is not allowed to
+        // download this file because it's not assigned to an action and not
+        // assigned to this controller. Otherwise no controllerId is set =
+        // anonymous download
+        if (secruityToken.getControllerId() != null) {
+            LOG.debug("no anonymous download request, doing authentication check for target {} and artifact {}",
+                    secruityToken.getControllerId(), localArtifact);
```
SirWayne: save controller id in a local variable

michahirsch: :+1:  for better readability I agree

hawkbit-dmf-amqp/src/main/java/org/eclipse/hawkbit/amqp/AmqpMessageHandlerService.java
```diff
@@ -213,6 +205,34 @@ private Message handleAuthentifiactionMessage(final Message message) {
...
+            final LocalArtifact localArtifact) {
+        final String controllerId = secruityToken.getControllerId();
+        if (controllerId == null) {
+            LOG.info("anonymous download no authentication check for artifact {}", localArtifact);
+            return;
+        }
+        LOG.debug("no anonymous download request, doing authentication check for target {} and artifact {}",
+                controllerId, localArtifact);
+        if (!controllerManagement.hasTargetArtifactAssigned(controllerId, localArtifact)) {
+            LOG.info("target {} tried to download artifact {} which is not assigned to the target");
```
schabdo: You define two placeholders, but actually do not provide any parameters for them.

Jonkno: Add the two missing parameter {} in Log output

hawkbit-dmf-amqp/src/main/java/org/eclipse/hawkbit/amqp/AmqpMessageHandlerService.java
```diff
@@ -166,20 +166,12 @@ private Message handleAuthentifiactionMessage(final Message message) {
             final LocalArtifact localArtifact = findLocalArtifactByFileResource(fileResource);
 
             if (localArtifact == null) {
+                LOG.info("target {} requested file resource which does not exists to download",
+                        secruityToken.getControllerId(), fileResource);
```
Jonkno: Remove fileResource parameter or add {} to Log info

hawkbit-dmf-amqp/src/test/java/org/eclipse/hawkbit/amqp/AmqpMessageHandlerServiceTest.java
```diff
@@ -313,11 +313,9 @@ public void authenticationRequestAllowedForArtifactWhichExistsAndAssignedToTarge
 
         // mock
         final LocalArtifact localArtifactMock = mock(LocalArtifact.class);
-        final Action actionMock = mock(Action.class);
         final DbArtifact dbArtifactMock = mock(DbArtifact.class);
         when(artifactManagementMock.findFirstLocalArtifactsBySHA1(anyString())).thenReturn(localArtifactMock);
-        when(controllerManagementMock.getActionForDownloadByTargetAndSoftwareModule(anyObject(), anyObject()))
-                .thenReturn(actionMock);
+        when(controllerManagementMock.hasTargetArtifactAssigned(anyObject(), anyObject())).thenReturn(true);
```
schabdo: Why you don't use `localArtifactMock` and the targetId used by the test?
So you only mock expected behavior and don't return anything by accident.

hawkbit-repository/src/main/java/org/eclipse/hawkbit/repository/specifications/ActionSpecifications.java
```diff
@@ -0,0 +1,60 @@
...
+     * 
+     * @param target
+     *            the target to verfiy if the given artifact is currently
+     *            assigned or had been assigned
+     * @param localArtifact
+     *            the local artifact to check wherever the target had ever been
+     *            assigned
+     * @return a specification to use with spring JPA
+     */
+    public static Specification<Action> hasTargetArtifactAssigned(final Target target,
```
schabdo: What about `hasTargetAssignedArtifact(...)`? Sounds better.

hawkbit-repository/src/main/java/org/eclipse/hawkbit/repository/specifications/ActionSpecifications.java
```diff
@@ -0,0 +1,60 @@
...
+            final LocalArtifact localArtifact) {
+        return (actionRoot, query, cb) -> {
```
schabdo: Is `cb` really a no-brainer here without any type?

PR #124, opened 2016-04-07
Reduce duplicate code

SirWayne:
Refactoring Table, TableHeader and Detail Layout components  …
 - Add interfaces to reduce boiler plate code
 - Add generic to reuse code
 - Try to remove duplicate code

PR #140, opened 2016-04-12
Bug/upload artifact

SirWayne:
Update the entity before call the super method, because they use the same getter for 

### DIFF
--- a/hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/artifacts/smtable/SoftwareModuleTable.java
+++ b/hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/artifacts/smtable/SoftwareModuleTable.java
@@ -9,6 +9,7 @@
 package org.eclipse.hawkbit.ui.artifacts.smtable;
 
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 
 import org.eclipse.hawkbit.repository.SoftwareManagement;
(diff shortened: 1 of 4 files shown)


michahirsch:
Can you explain why it is a different to call the ` super.updateEntity(baseEntity, item);` after setting the properties? When I look into the `AbstractTable#updateEntity(final E baseEntity, final Item item)` implementation it seems that the abstract classes only setting properties as well. Seems odd that changing the order of setting the properties is fixing something :)

SirWayne:
If you do it after `super.updateEntity(baseEntity, item);` then you have to set the `VAR_NAME` again. I don't know why, i guess the entity proxy do something special with the name and version (`item.getItemProperty(SPUILabelDefinitions.NAME_VERSION).setValue(swNameVersion);` there happen some calls via reflection).

michahirsch:
Is it only because of the `NAME_VERSION` property? Looks very strange that this is working and another way not? I guess even a better approach would not to set the property at all manually 
```
final String swNameVersion = HawkbitCommonUtil.concatStrings(":", baseEntity.getName(),		
baseEntity.getVersion());		
item.getItemProperty(SPUILabelDefinitions.NAME_VERSION).setValue(swNameVersion);
```

because this could also be implemented in the getter of the class `org.eclipse.hawkbit.ui.artifacts.smtable.ProxyBaseSoftwareModuleItem.getNameAndVersion()`

put the logic into the item bean like this?
```
    public String getNameAndVersion() {
        return HawkbitCommonUtil.concatStrings(":", getName(), getVersion());
    }
```

SirWayne:
it just happen with the 'VAR_NAME' property. If you set the update before 'item.getItemProperty(SPUILabelDefinitions.NAME_VERSION).setValue(swNameVersion);'. Then you have to set the 'VAR_NAME' again.

'baseEntity' is not the proxy. I don't know the sequence of events with the LazyQueryContainer and reflection. Now the properties and the 'setValues' will set in the same order before this pull requests:
https://github.com/eclipse/hawkbit/pull/124. I adapt this order of events.

michahirsch:
ok discussed and is working like this

PR #144, opened 2016-04-16
change transaction handling of creating groups and check running Rollouts

michahirsch:
* change transaction behavior of `checkRunningRollouts` to `READ_UNCOMITTED` so the transaction is not blocked by the transaction of creating large rollouts which leads into a dead lock scenario when acquire lock for checking but large rollout is creating and locking the table.
```
Internal Exception: java.sql.BatchUpdateException: Deadlock found when trying to get lock; try restarting transaction
Call: UPDATE sp_rollout SET optlock_revision = (optlock_revision + ?), last_check = ? WHERE (((last_check < (? - ?)) AND (status = ?)) AND (tenant = ?))
``` 


hawkbit-repository/src/main/java/org/eclipse/hawkbit/repository/RolloutManagement.java
```diff
@@ -308,7 +313,8 @@ private Rollout createRolloutGroups(final int amountGroup, final RolloutGroupCon
         int groupIndex = 0;
         final Long totalCount = savedRollout.getTotalTargets();
         final int groupSize = (int) Math.ceil((double) totalCount / (double) amountGroup);
-        // validate if the amount of groups that will be created are the amount
+        // validate if the amount of groups that will be created are the
+        // amount
```
Jonkno: Move comment to have only two lines

hawkbit-repository/src/main/java/org/eclipse/hawkbit/repository/RolloutManagement.java
```diff
@@ -288,6 +284,15 @@ private void verifyRolloutGroupParameter(final int amountGroup) {
         }
     }
 
+    private Rollout createRolloutGroupsInNewTransaction(final int amountGroup, final RolloutGroupConditions conditions,
```
Jonkno: Use a more meaningful name = amountOfGroups

PR #162, opened 2016-05-03
Create filter has different layout than rollout and the other widgets

melanieretter:
The "create filter" icon in the target filter management view did not have the same look like in other views. Fixed that UI problem. In addition I added a constant for the header caption.

create_filter_icon

### DIFF
--- a/hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/filtermanagement/TargetFilterHeader.java
+++ b/hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/filtermanagement/TargetFilterHeader.java
@@ -78,9 +78,8 @@ public class TargetFilterHeader extends VerticalLayout {
     }
 
     private Label createHeaderCaption() {
-        final Label captionLabel = SPUIComponentProvider.getLabel(getHeaderCaption(),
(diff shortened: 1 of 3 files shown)


hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/filtermanagement/TargetFilterHeader.java
```diff
@@ -187,4 +185,11 @@ protected void resetSearchText() {
         eventBus.publish(this, CustomFilterUIEvent.FILTER_BY_CUST_FILTER_TEXT_REMOVE);
     }
 
+    protected String getAddIconId() {
```
SirWayne: Use private scope

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/filtermanagement/TargetFilterHeader.java
```diff
@@ -187,4 +185,11 @@ protected void resetSearchText() {
...
+        return SPUIComponetIdProvider.TARGET_FILTER_ADD_ICON_ID;
+    }
+
+    protected String getHeaderCaption() {
```
SirWayne: Use private scope

PR #163, opened 2016-05-03
Filter icon (succes/error) not centered

melanieretter:
The error/success icon in the "target filter management - create filter" view is centered now.
filter_icon_not_centered

Furthermore I changed the position of some table headers because they were in the center neither.


hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/common/detailslayout/AbstractTableDetailsLayout.java
```diff
@@ -152,18 +152,18 @@ private void buildLayout() {
         final HorizontalLayout nameEditLayout = new HorizontalLayout();
         nameEditLayout.setWidth(100.0f, Unit.PERCENTAGE);
```
Jonkno: Use F

PR #164, opened 2016-05-03
Reflect tag colour change immediatly after popup is closed. 

melanieretter:
After changing the colour of a tag it is displayed immediately without refreshing the page. (deployment view)
Furthermore I corrected a spelling mistake.


hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/artifacts/smtype/CreateUpdateSoftwareTypeLayout.java
```diff
@@ -612,7 +612,7 @@ private void discard() {
      */
     private void crateNewSWModuleType() {
```
michahirsch: There is also a typo in the method name `crateNewSWModuleType() ` => `createNewSWModuleType()`

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/management/targettag/CreateUpdateTargetTagLayout.java
```diff
@@ -183,7 +183,7 @@ private Boolean mandatoryValuesPresent() {
      * Create new tag.
      */
     private void crateNewTag() {
```
Jonkno: There is another typo in crateNewTag

michahirsch:
No exactly sure about this change because the `TargetTagUpdateEvent` is an repository event I think is not nice about firing this event within the UI code. Another thing is that the `TargetTagUpdateEvent` is and should be fired in the `org.eclipse.hawkbit.repository.TagManagement.updateTargetTag(TargetTag)`  https://github.com/eclipse/hawkbit/pull/164/files#diff-8c4c87bbd03cdd4a44df5f98a8344a71R218

Is the immediate color change working for the `DistributionSet` ? Because the code of the `CreateUpdateDistributionTagLayoutWindow` looks very similar to the `CreateUpdateTargetTagLayout` but it does not fire any repository event.

michahirsch:
@melanieretter  can you merge this branch so it's mergeable with the master, then I'm going to merge it. Thanks :+1:

Issue #100, opened 2016-03-21
Provide abstact interface for hawkBits repository implementation 


kaizimmerm:
Currently hawkBits interfaces interact directly with the repository implementation. That makes it very hard to evaluate/implement alternative implementations on different persistence providers (e.g. MongoDB, Geode, Cassandra etc.).

What we should do:
* Provide interfaces for the repository services.
* Migrate current repository to be a JPA implementation.

Tasks:
- [x]  Split management methods (interface and JPA impl)
- [x]  Split model POJOs (interface and JPA imply)
- [x]  Remove org.springframework.data.jpa.domain.Specification from API
- [x]  Split maven modules into API and JPA
- [x]  Refactor TestDataUtils to use new API.
- [x]  Refactor default data generation to use new API and create more reasonable type setup.
- [x]  Document new setup in wiki.
- [x]  Remove dependencies on JPA part from the other modules (especially the tests) and use new API instead
- [x]  Provide auto configuration for implementations

kaizimmerm:
Moved the interface refactoring part to https://github.com/eclipse/hawkbit/issues/197

PR #174, opened 2016-05-13
Provide abstact interface for hawkBits repository implementation

kaizimmerm:
Note: work in progress, this pull request is for discussion.

Implements #100.

* Java Interfaces for the repository that are now used in the API/UI modules.
* TestDataUtil migrated to a proper repository implementation independent service
* While at it i changed the default SM/DS type generation for new accounts/tenants to a more generic model. Removed the old model from the initial contribution.


kaizimmerm:
create a new because of failed IP check

PR #175, opened 2016-05-13
allow the getTargetSecurityToken can be called as system code

michahirsch:
Allow that the `org.eclipse.hawkbit.repository.model.Target.getSecurityToken()` can be called as system-code. 

Because since we are sending the target-security-token to the DMF-API, it shouldn't be up-to the current security-context of the user which does the assignment to read the token. 

I'm explicit against the idea to move the security check out of the `Target` entity because then you'll need to be aware of the security check in every API which exposes the security-token as well as on the UI. This will lead to distributed security checks through the code.

### DIFF
--- a/hawkbit-security-core/src/main/java/org/eclipse/hawkbit/security/SystemSecurityContext.java
+++ b/hawkbit-security-core/src/main/java/org/eclipse/hawkbit/security/SystemSecurityContext.java
@@ -49,6 +49,21 @@ public class SystemSecurityContext {
         this.tenantAware = tenantAware;
     }
 
+    /**
+     * Runs a given {@link Callable} within a system security context, which is
+     * permitted to call secured system code. Often the system needs to call
+     * secured methods by it's own without relying on the current security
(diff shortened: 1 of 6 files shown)


hawkbit-security-core/src/main/java/org/eclipse/hawkbit/security/SystemSecurityContext.java
```diff
@@ -68,6 +83,17 @@ public SystemSecurityContext(final TenantAware tenantAware) {
         }
     }
 
+    /**
+     * @return {@code true} if the current running code is running as system
+     *         code block.
+     */
+    public boolean isCurrentThreadSystemCode() {
+        if (SecurityContextHolder.getContext().getAuthentication() instanceof SystemCodeAuthentication) {
```
SirWayne: You can return the result of the if statement ```return SecurityContextHolder.getContext().getAuthentication() instanceof SystemCodeAuthentication```

PR #181, opened 2016-05-17
Feature/add rest api with java client

### LINKED ISSUE #99: Refactor module structure for DDI and mgmt. API
Currently Artifact download, DDI and mgmt. API are mixed in the the two "rest" modules. We should split this in DDI and mgmt. API and in addition name the modules correctly to a avoid confusing our users.

Module naming proposal:
- hawkbit-ddi-api
- hawkbit-ddi-resource
- hawkbit-ddi-dl-api (for the download part)
- hawkbit-ddi-dl-resource (for the download part)
- hawkbit-mgmt-api
- hawkbit-mgmt-resource

While at it we should rename the UI module to management-ui.

SirWayne:
Fix  #99

- Split rest modules in DDI and Mgmt api and resource modules 
- Add DDI client resources
- Add DDI example client
- Rename all Ddi* and Mgmt* rest resources classes.


ddewaele:
This PR looks really interesting. haven't looked at all the commits but was wondering if this is also taken into account : 

- API versioning (Some way to version the mgmt apis)
- ISO 8601 standard for JSON dates (instead of custom string based formats)
- Feign client : Using Jackson encoder / decoder instead of working with Strings (as currently being done in the ControllerResource interface in the simulator .

michahirsch:
@ddewaele 

This PR is intend to introduce interfaces for the REST-API which can be used e.g. with a feign-client to generate a stub and to split up the different APIs into different maven-modules.

* API versioning:
 * The packages are currently not versioned, the server API uses a version identifier in the URL, not sure if versioning the packages yet, maybe is better to introduce a new maven module for a another version, so there no versioning in the package structure yet, but we'll think about it.

*  ISO 8601 standard
  * Yes as discussed also in Gitter I think this will make sense instead of a custom string, but this pull-request is splitting up modules and refactoring so for the time I've created an Github issue https://github.com/eclipse/hawkbit/issues/191

* Feign client
  * Yes you are right, the refactoring and introducing the DDI interfaces for using feign was also intend that the device-simulator can use then these interfaces to generate a feign-client stub instead using a plain feign-client JSON based interface. This will be done in further improvement of the device-simulator

PR #183, opened 2016-05-18
Switch back Spring Boots default logger implementation logback

kaizimmerm:
The log4j2 usage was introduced in the past as we expected better JUL integration. However, turned out logback is doing that as well as log4j2. In addition logback allows easier configuration through boots property mechanism.

I fixed also a small simulator glitch and added a few test asserts into DMF tests.


michahirsch:
The logback.xml is missing the license header.

PR #184, opened 2016-05-18
Fix display ds name version actionhistory table

venu1278:
i)The truncation of DistributionSet Name/Version is removed , as user unable to see even expand the column Deployment management, in action-history table.

Signed-off-by: B. Venugopal venugopal.boodidadinne@in.bosch.com


kaizimmerm:
Looks good, thanks.

PR #185, opened 2016-05-18
fix_flashing_whitepage_on_login_application

venu1278:
fix for flashing UI on log in application.

Signed-off-by: venu1278 <venugopal.boodidadinne@in.bosch.com>


michahirsch:
Tried it out and looks good to me, thanks :+1:

PR #203, opened 2016-06-07
Upgraded Spring AMQP. Added task scheduler. Improved executor shutdown.

kaizimmerm:
Small things, but they make our live easier. Includes upgrade to Spring AMQP 1.6 and more robustness on DMF rabbit connection.

Especially with the task scheduler and Spring AMQP 1.6 you can do wonderful asynchronous stuff.


hawkbit-dmf-amqp/src/main/java/org/eclipse/hawkbit/amqp/AmqpProperties.java
```diff
@@ -39,6 +41,11 @@
     private boolean missingQueuesFatal = false;
 
     /**
+     * Requested heartbeat interval from broker in {@link TimeUnit#SECONDS}.
+     */
+    private int requestedHeartBeat = 60;
```
SirWayne: Is it better to use `(int) TimeUnit.SECONDS.toSeconds(60)`. We use this in some other configurations

PR #205, opened 2016-06-10
Fixed bug where DS could not be deleted as it was assigned to a rollout.

kaizimmerm:
If a DS is assigned to a rollout it has to be soft deleted and not hard which in fact will fail anyhow as we have foreign key constrain there.


hawkbit-repository/hawkbit-repository-jpa/src/test/java/org/eclipse/hawkbit/repository/jpa/DistributionSetManagementTest.java
```diff
@@ -777,36 +784,55 @@ public void findAllDistributionSetMetadataByDsId() {
...
     }
 
+    private Rollout createRolloutByVariables(final String rolloutName, final String rolloutDescription,
+            final int groupSize, final String filterQuery, final DistributionSet distributionSet,
+            final String successCondition, final String errorCondition) {
+        final RolloutGroupConditions conditions = new RolloutGroupConditionBuilder()
+                .successCondition(RolloutGroupSuccessCondition.THRESHOLD, successCondition)
+                .errorCondition(RolloutGroupErrorCondition.THRESHOLD, errorCondition)
+                .errorAction(RolloutGroupErrorAction.PAUSE, null).build();
+        final Rollout rolloutToCreate = new JpaRollout();
```
SirWayne: Is it possible to use the `entityFactory`, or?

kaizimmerm: Possible but not needed in the JPA repo itself.

PR #210, opened 2016-06-15
Refactoring

SirWayne:
1. Refactor component scan for all managment jpa services. The beans for the jpa services are created by the RepositoryApplicationConfiguration and not by the spring component scan.
Therefore it is necessary, to create a service for SystemManagementCacheKeyGenerator.

2. Split InMemoryUserManagementConfiguration from the SecurityAutoConfiguration. 
If no userdetail service is avilable, the configuration will created,
3. Add a util class to collect permission easily + unit test


hawkbit-security-core/src/main/java/org/eclipse/hawkbit/security/SpringSecurityAuditorAware.java
```diff
@@ -29,16 +25,21 @@ public String getCurrentAuditor() {
 
         final Authentication authentication = SecurityContextHolder.getContext().getAuthentication();
 
-        if (authentication == null || !authentication.isAuthenticated()) {
+        if (isAuthentifactionInvalid(authentication)) {
```
melanieretter: isAuthenti**ca**tionInvalid

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/RepositoryApplicationConfiguration.java
```diff
@@ -173,4 +202,94 @@ protected AbstractJpaVendorAdapter createJpaVendorAdapter() {
     public PlatformTransactionManager transactionManager() {
         return new MultiTenantJpaTransactionManager();
     }
+
+    @Bean
```
Jonkno: Missing javadoc for all methods

PR #215, opened 2016-06-17
Feature manage sw ds metadata

venu1278:
Feature of UI for cretae, update and delete metadata for Distibutions & SoftwareModules.

Signed-off-by: venugopal<venugopal.boodidadinne@in.bosch.com>


kaizimmerm:
Broken commit history

PR #225, opened 2016-06-24
Fix wrong multi part exception message in response.

SirWayne:
Fix wrong multi part exception message in response.
Additional hardening all remaining exception message and classes.

### DIFF
--- a/hawkbit-core/src/main/java/org/eclipse/hawkbit/exception/SpServerError.java
+++ b/hawkbit-core/src/main/java/org/eclipse/hawkbit/exception/SpServerError.java
@@ -90,10 +90,6 @@ public enum SpServerError {
     /**
     *
     */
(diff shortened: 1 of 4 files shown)


Jonkno:
Did you also add a unit test for this bug?

SirWayne:
Don't know how, because the `FileUploadException` message is generic.

michahirsch:
Î tried it out, doing upload without `boundary` containing in the `Content-Type` header `multipart/form-data` the exception looks correct now.

 ```
{
  "exceptionClass": "org.eclipse.hawkbit.repository.exception.MultiPartFileUploadException",
  "errorCode": "hawkbit.server.error.artifact.uploadFailed",
  "message": "the request was rejected because no multipart boundary was found"
}
```

The file size exceeded exception looks then like 

```
{
  "exceptionClass": "org.eclipse.hawkbit.repository.exception.MultiPartFileUploadException",
  "errorCode": "hawkbit.server.error.artifact.uploadFailed",
  "message": "Could not parse multipart servlet request; nested exception is java.lang.IllegalStateException: org.apache.tomcat.util.http.fileupload.FileUploadBase$FileSizeLimitExceededException: The field file exceeds its maximum permitted size of 1048576 bytes."
}
```

PR #226, opened 2016-06-24
Management UI DS and SM metadata mgmt

kaizimmerm:
Management UI handles Software Module and Distribution Set Metadata.

This has been in Management API for quite some time but was still missing in UI.


hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/common/detailslayout/DistributionSetMetadatadetailslayout.java
```diff
@@ -0,0 +1,190 @@
...
+    
+
+    /**
+     * Populate software module metadata.
+     * 
+     * @param distributionSet
+     */
+    public void populateDSMetadata(final DistributionSet distributionSet) {
+        removeAllItems();
+        if (null != distributionSet) {
```
Jonkno: I would prefer a early return here

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/common/detailslayout/SoftwareModuleMetadatadetailslayout.java
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    /**
+     * Populate software module metadata table.
+     * 
+     * @param swModule
+     */
+    public void populateSMMetadata(final SoftwareModule swModule) {
+        removeAllItems();
+        if (null != swModule) {
```
Jonkno: I would prefer a early return here

kaizimmerm:
Can you please solve the merge conflicts?

PR #227, opened 2016-06-24
Fix executor service and created extra exchange for authentication requests.

kaizimmerm:
I am still on my mission to optimise performance and robustness on DMF.

One problem that we saw during load tests is that a crowded DMF queue with update feedback actually blocks authentication requests coming through. I created an extra non durable exchange/queue setup for these.

Using device simulator for load testing is getting more fun along the way. However, originally it was not created for that purpose so there is always stuff to do. I am using now the poll scheduler also for DMF and had to fix a couple of things there.

I fixed also the mess in the central executor services. We have now a clean setup for standard executors and scheduled ones.


michahirsch:
looks ok :+1:

PR #229, opened 2016-06-27
Return last throwable in response

SirWayne:
Return last throwable in response for MultiPartException


hawkbit-rest-core/src/main/java/org/eclipse/hawkbit/rest/exception/ResponseExceptionHandler.java
```diff
@@ -37,6 +42,9 @@
     private static final Map<SpServerError, HttpStatus> ERROR_TO_HTTP_STATUS = new EnumMap<>(SpServerError.class);
     private static final HttpStatus DEFAULT_RESPONSE_STATUS = HttpStatus.INTERNAL_SERVER_ERROR;
 
+    @Autowired
+    private ResourceLoader resourceLoader;
```
michahirsch: The `ResourceLoader` is not used, does we need it?

michahirsch:
ok :+1:

PR #232, opened 2016-06-29
Made queue declaration retry count configurable.

kaizimmerm:
We had the need to configure queue declaration retry count. I added an new RabbitListenerContainerFactory implementation to make this configurable.

While at it I fixed also a wrong exception, byte to string conversion and fixed a sonar issue.

### DIFF
--- a/hawkbit-dmf-amqp/src/main/java/org/eclipse/hawkbit/amqp/AmqpMessageHandlerService.java
+++ b/hawkbit-dmf-amqp/src/main/java/org/eclipse/hawkbit/amqp/AmqpMessageHandlerService.java
@@ -408,12 +408,16 @@ public class AmqpMessageHandlerService extends BaseAmqpService {
 
         if (ArrayUtils.isNotEmpty(message.getMessageProperties().getCorrelationId())) {
             actionStatus.addMessage(RepositoryConstants.SERVER_MESSAGE_PREFIX + "DMF message correlation-id "
-                    + new String(message.getMessageProperties().getCorrelationId()));
+                    + convertCorrelationId(message));
         }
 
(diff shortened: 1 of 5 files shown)


hawkbit-dmf-amqp/src/main/java/org/eclipse/hawkbit/amqp/AmqpMessageHandlerService.java
```diff
@@ -414,7 +408,7 @@ private ActionStatus createActionStatus(final Message message, final ActionUpdat
 
         if (ArrayUtils.isNotEmpty(message.getMessageProperties().getCorrelationId())) {
             actionStatus.addMessage(RepositoryConstants.SERVER_MESSAGE_PREFIX + "DMF message correlation-id "
-                    + message.getMessageProperties().getCorrelationId());
+                    + new String(message.getMessageProperties().getCorrelationId()));
```
schabdo: Why you're using `new String(...)`?

hawkbit-dmf-amqp/src/test/java/org/eclipse/hawkbit/amqp/AmqpMessageHandlerServiceTest.java
```diff
@@ -141,7 +141,7 @@ public void testWrongContentType() {
         try {
             amqpMessageHandlerService.onMessage(message, MessageType.THING_CREATED.name(), TENANT, "vHost");
             fail("IllegalArgumentException was excepeted due to worng content type");
```
Jonkno: Also adjust comment -> AmqpRejectAndDontRequeueException now

PR #238, opened 2016-07-06
Traffic data in DownloadProgress event

kaizimmerm:
Added Traffic data to DownloadProgressEvent. That is in many use cases more helpful than the percentage which can be calculated by the receiver quite easily.

Fixed a bug that the DownloadProgressEvent is based on Action Id which is wrong. It should be ActionStatus as one of those is created per download. I guess this got messed up when we gave up the UpdateActionStatus concept.

### DIFF
--- a/hawkbit-cache-redis/src/test/java/org/eclipse/hawkbit/cache/eventbus/EventDistributorTest.java
+++ b/hawkbit-cache-redis/src/test/java/org/eclipse/hawkbit/cache/eventbus/EventDistributorTest.java
@@ -56,7 +56,7 @@ public class EventDistributorTest {
     @Test
     public void distributeDistributedEventSendsToRedis() {
 
-        final DownloadProgressEvent event = new DownloadProgressEvent("tenant", 123L, 10);
+        final DownloadProgressEvent event = new DownloadProgressEvent("tenant", 123L, 10, 100L);
         underTest.distribute(event);
 
         // origin node ID should be set by distributing the event
@@ -67,7 +67,7 @@ public class EventDistributorTest {
     @Test
     public void dontDistributeDistributedEventIfSameNode() {
         final String knownNodeId = EventDistributor.getNodeId();
-        final DownloadProgressEvent event = new DownloadProgressEvent("tenant", 123L, 10);
+        final DownloadProgressEvent event = new DownloadProgressEvent("tenant", 123L, 10, 100L);
         event.setNodeId(knownNodeId);
 
         // test
@@ -79,7 +79,7 @@ public class EventDistributorTest {
 
     @Test
     public void handleDistributedMessageFromRedis() {
-        final DownloadProgressEvent event = new DownloadProgressEvent("tenant", 123L, 10);
+        final DownloadProgressEvent event = new DownloadProgressEvent("tenant", 123L, 10, 100L);
         final String knownChannel = "someChannel";
 
         underTest.handleMessage(event, knownChannel);
@@ -90,7 +90,7 @@ public class EventDistributorTest {
 
     @Test
     public void handleDistributedMessageFilteredIfSameNodeId() {
-        final DownloadProgressEvent event = new DownloadProgressEvent("tenant", 123L, 10);
+        final DownloadProgressEvent event = new DownloadProgressEvent("tenant", 123L, 10, 100L);
         final String knownChannel = "someChannel";
         event.setOriginNodeId(EventDistributor.getNodeId());
 
--- a/hawkbit-cache-redis/src/test/java/org/eclipse/hawkbit/cache/eventbus/EventDistributorTest.java
+++ b/hawkbit-cache-redis/src/test/java/org/eclipse/hawkbit/cache/eventbus/EventDistributorTest.java
@@ -56,7 +56,7 @@ public class EventDistributorTest {
     @Test
     public void distributeDistributedEventSendsToRedis() {
 
-        final DownloadProgressEvent event = new DownloadProgressEvent("tenant", 123L, 10, 100L);
+        final DownloadProgressEvent event = new DownloadProgressEvent("tenant", 123L, 10, 100L, 200L);
         underTest.distribute(event);
 
         // origin node ID should be set by distributing the event
@@ -67,7 +67,7 @@ public class EventDistributorTest {
     @Test
     public void dontDistributeDistributedEventIfSameNode() {
         final String knownNodeId = EventDistributor.getNodeId();
-        final DownloadProgressEvent event = new DownloadProgressEvent("tenant", 123L, 10, 100L);
+        final DownloadProgressEvent event = new DownloadProgressEvent("tenant", 123L, 10, 100L, 200L);
         event.setNodeId(knownNodeId);
 
         // test
@@ -79,7 +79,7 @@ public class EventDistributorTest {
 
     @Test
     public void handleDistributedMessageFromRedis() {
-        final DownloadProgressEvent event = new DownloadProgressEvent("tenant", 123L, 10, 100L);
+        final DownloadProgressEvent event = new DownloadProgressEvent("tenant", 123L, 10, 100L, 200L);
         final String knownChannel = "someChannel";
 
         underTest.handleMessage(event, knownChannel);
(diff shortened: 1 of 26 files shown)


hawkbit-cache-redis/src/test/java/org/eclipse/hawkbit/cache/eventbus/EventDistributorTest.java
```diff
@@ -56,7 +56,7 @@ public void before() {
     @Test
     public void distributeDistributedEventSendsToRedis() {
```
Jonkno: Missing description for all tests

hawkbit-repository/hawkbit-repository-api/src/main/java/org/eclipse/hawkbit/repository/SystemManagement.java
```diff
@@ -63,13 +63,18 @@
     /**
      * @return {@link TenantMetaData} of {@link TenantAware#getCurrentTenant()}
      */
+    // @PreAuthorize(SpringEvalExpressions.HAS_AUTH_READ_REPOSITORY +
```
Jonkno: Remove

hawkbit-repository/hawkbit-repository-api/src/main/java/org/eclipse/hawkbit/repository/SystemManagement.java
```diff
@@ -79,6 +84,7 @@
      *            to retrieve data for
      * @return {@link TenantMetaData} of given tenant
      */
+    // @PreAuthorize(SpringEvalExpressions.IS_SYSTEM_CODE)
```
Jonkno: Remove

hawkbit-repository/hawkbit-repository-api/src/main/java/org/eclipse/hawkbit/repository/SystemManagement.java
```diff
@@ -88,6 +94,7 @@
      *            to update
      * @return updated {@link TenantMetaData} entity
      */
+    // @PreAuthorize(SpringEvalExpressions.HAS_AUTH_TENANT_CONFIGURATION)
```
Jonkno: Remove

hawkbit-cache-redis/src/test/java/org/eclipse/hawkbit/cache/eventbus/EventDistributorTest.java
kaizimmerm: This test is not from me.

michahirsch: This code has to be re-think when hawkBit learns really cluster communication, this code was prototyping of using redis as inter-cluster-communication to distribute events.

michahirsch:
The `JpaAction` still defines the `downloadProgressPercent` which probably now needs to be moved to the `JpaActionStatus` entity
```
    /**
     * Note: filled only in {@link Status#DOWNLOAD}.
     */
    @Transient
    @CacheField(key = CacheKeys.DOWNLOAD_PROGRESS_PERCENT)
    private int downloadProgressPercent;
```

This field is automatically be read by the `org.eclipse.hawkbit.repository.jpa.model.CacheFieldEntityListener`

hawkbit-repository/hawkbit-repository-test/src/main/java/org/eclipse/hawkbit/repository/test/util/AbstractIntegrationTest.java
```diff
@@ -94,6 +95,9 @@
     protected TargetManagement targetManagement;
 
     @Autowired
+    protected TenantStatsManagement tenantStatsManagement;
```
michahirsch: This field is never used in test, I guess we can just remove it and add it if it's used

hawkbit-repository/hawkbit-repository-api/src/main/java/org/eclipse/hawkbit/repository/TenantStatsManagement.java
```diff
@@ -16,19 +16,17 @@
...
-     * to collect data of all tenants in the system.
+     * Service for stats of a single tenant.
      *
-     * @param tenant
-     *            to collect for
      * @return collected statistics
      */
-    @PreAuthorize(SpringEvalExpressions.HAS_AUTH_SYSTEM_ADMIN)
-    TenantUsage getStatsOfTenant(String tenant);
+    @PreAuthorize(SpringEvalExpressions.HAS_AUTH_READ_REPOSITORY + SpringEvalExpressions.HAS_AUTH_OR
```
michahirsch: Is there a reason why we need so many different permissions for this? Need some clarification. I'd like to add in hawkBit that system code is generally allowed to call management APIs.

hawkbit-repository/hawkbit-repository-api/src/main/java/org/eclipse/hawkbit/repository/TenantStatsManagement.java
```diff
@@ -16,19 +16,17 @@
...
+            + SpringEvalExpressions.HAS_AUTH_READ_TARGET + SpringEvalExpressions.HAS_AUTH_OR
+            + SpringEvalExpressions.HAS_AUTH_TENANT_CONFIGURATION + SpringEvalExpressions.HAS_AUTH_OR
+            + SpringEvalExpressions.IS_SYSTEM_CODE)
+    TenantUsage getStatsOfTenant();
```
michahirsch: I've liked the way with the `tenant` parameter more then using the `TenantAware.getCurrentTenant()` within this method. Is there a reason not using the `tenant` parameter anymore? :) Otherwise it would maybe make sense to write in the javadoc that this method returns the stats of the current tenant.

hawkbit-repository/hawkbit-repository-api/src/main/java/org/eclipse/hawkbit/repository/TenantStatsManagement.java
kaizimmerm: I think there is no reason why only the system admin should have access to the stats. You can retrieve the data with the other permissions manually as well, so there is no need to hide the simplifying stats method.

hawkbit-repository/hawkbit-repository-api/src/main/java/org/eclipse/hawkbit/repository/TenantStatsManagement.java
kaizimmerm: We never use tenant in the signature of a repository method.

michahirsch:
I need one explanation on this code https://github.com/eclipse/hawkbit/pull/238/files#diff-77dc4c2f75458d290461d921f2bea4faR306
```
long toRead = length;
boolean toContinue = true;
long shippedSinceLastEvent = 0;

while (toContinue) {
    final int r = from.read(buf);
    if (r == -1) {
        break;
    }

toRead -= r;
if (toRead > 0) {
    to.write(buf, 0, r);
    total += r;
    shippedSinceLastEvent += r;
} else {
    to.write(buf, 0, (int) toRead + r);
    total += toRead + r;
    shippedSinceLastEvent += toRead + r;
    toContinue = false;
}
```
I know this is there already for quite some time but I don't get the else block. The variable `toRead` will be never negative only positive or zero, right? So the else block adds the `toRead` to the variable `r` which is the current read size. I guess the else block can be totally removed? Or am I wrong by doing so

```
int r;
while ((r = from.read(buf)) > 0) {
    to.write(buf, 0, r);
    total += r;
    if (controllerManagement != null) {
        final int newPercent = DoubleMath.roundToInt(total * 100.0 / length, RoundingMode.DOWN);

        // every 10 percent an event
        if (newPercent == 100 || newPercent > progressPercent + 10) {
            progressPercent = newPercent;
            controllerManagement.downloadProgress(statusId, length, (long) r, total);
        }
    }
}
return total;
```

Correct? Or do I missing something?

kaizimmerm:
'while ((r = from.read(buf)) > 0) {' is not gonna work as the 'length' parameter from the range request might be smaller that the bytes you can read. However, instead of cutting down on the write below you could calculate the bytes to be read before the read request, i.e. if toRead smaller than buffer read only that.

PR #243, opened 2016-07-14
Correct colour of cancel/save button in DialogWindows

melanieretter:
The colour of the cancel and save button was different as the other colours in the theme. Now it fits to the theme.

borderless_button

### DIFF
--- a/hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/decorators/SPUIButtonStyleNoBorderWithIcon.java
+++ b/hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/decorators/SPUIButtonStyleNoBorderWithIcon.java
@@ -15,7 +15,7 @@ import com.vaadin.ui.Button;
 import com.vaadin.ui.themes.ValoTheme;
 
 /**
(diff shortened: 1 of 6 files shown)


hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/decorators/SPUIButtonStyleNoBorderWithIcon.java
```diff
@@ -25,18 +26,33 @@ public Button decorate(final Button button, final String style, final boolean se
...
+        button.addStyleName(ValoTheme.LABEL_SMALL);
+        button.addStyleName(ValoTheme.BUTTON_BORDERLESS_COLORED);
+
         setButtonStyle(style, setStyle);
         setButtonIcon(icon);
 
-        button.addStyleName(ValoTheme.LABEL_SMALL);
         button.setSizeFull();
 
+        setButtonStyle(style, setStyle);
```
schabdo: You are calling `setButtonStyle(style, setStyle);` already in line 32, so why once more?

PR #257, opened 2016-07-28
sonar pom hygiene

michahirsch:
* sonar pom hygiene
* remove the aspectJ weaver from argline


pom.xml
```diff
@@ -312,36 +292,21 @@
                      <exclude>**/Abstract*.java</exclude>
                   </excludes>
                </configuration>
-               <dependencies>
-                  <dependency>
-                     <groupId>org.aspectj</groupId>
-                     <artifactId>aspectjweaver</artifactId>
-                     <version>${aspectj.version}</version>
```
schabdo: Is property `aspectj.version` still needed? If not, remove it as well

pom.xml
```diff
@@ -127,30 +127,11 @@
       <sonar.java.coveragePlugin>jacoco</sonar.java.coveragePlugin>
```
schabdo: Still required? I think you can remove it!

pom.xml
michahirsch: removed

PR #259, opened 2016-07-29
Beautify error notifications

melanieretter:
Introduce a close button in the right upper corner of the error notification.  In case an exception occures the exception message is stored and can be displayed. The standard ErrorHandler of Hawkbit is only assigned if there is no other ErrorHandler assigned yet.


hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/HawkbitUI.java
```diff
@@ -180,7 +180,10 @@ public void afterViewChange(final ViewChangeEvent event) {
         final String locale = getLocaleId(SPUIDefinitions.getAvailableLocales());
         setLocale(new Locale(locale));
 
-        UI.getCurrent().setErrorHandler(new SPUIErrorHandler());
+        if (UI.getCurrent().getErrorHandler() == null) {
```
schabdo: Is there a case were you won't set the handler, because it is already set?

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/components/HawkbitUIErrorHandler.java
```diff
@@ -37,29 +32,55 @@
...
+    private String exceptionMessage;
+    protected Component errorOrigin;
 
     @Override
     public void error(final ErrorEvent event) {
-        final SPNotificationMessage notification = new SPNotificationMessage();
+        notification = new HawkbitNotificationMessage();
         // Build error style
-        final StringBuilder style = new StringBuilder(ValoTheme.NOTIFICATION_FAILURE);
+        style = new StringBuilder(ValoTheme.NOTIFICATION_FAILURE);
```
schabdo: You can use the fluent API of `StringBuilder` to concat all calls

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/components/HawkbitUIErrorHandler.java
```diff
@@ -37,29 +32,55 @@
...
-        final Component errorOrgin = findAbstractComponent(event);
-        if (null != errorOrgin && errorOrgin.getUI() != null) {
-            notification.showNotification(style.toString(), i18n.get("caption.error"),
-                    i18n.get("message.error.temp", new Object[] { exceptionName }), false,
-                    errorOrgin.getUI().getPage());
-        }
+        return exceptionName;
     }
+
+    protected Boolean getAndCheckErrorOrigin(final ErrorEvent event) {
```
schabdo: Method name is confusing

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/HawkbitUI.java
melanieretter: Yes. It is necessary if you have a different UI class with a specific errorHandler. The HawkbitUI always overwrites the errorHandler then.

schabdo: I see. Just asking because old code don't do that

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/components/SPUIErrorHandler.java
```diff
@@ -39,27 +34,56 @@
...
-        if (null != errorOrgin && errorOrgin.getUI() != null) {
-            notification.showNotification(style.toString(), i18n.get("caption.error"),
-                    i18n.get("message.error.temp", new Object[] { exceptionName }), false,
-                    errorOrgin.getUI().getPage());
+        return exceptionName;
+    }
+
+    protected Boolean getAndCheckErrorOrigin(final ErrorEvent event) {
+        errorOrigin = findAbstractComponent(event);
+        if (errorOrigin != null && errorOrigin.getUI() != null) {
```
schabdo: Better: `return (errorOrigin != null && errorOrigin.getUI() != null)`

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/components/HawkbitUIErrorHandler.java
```diff
@@ -0,0 +1,101 @@
...
+/**
+ * Default handler for SP UI.
+ */
+public class HawkbitUIErrorHandler extends DefaultErrorHandler {
+
+    /**
+     * Comment for <code>serialVersionUID</code>.
+     */
+    private static final long serialVersionUID = 1877326479308824191L;
+    /**
```
schabdo: Remove odd comment

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/components/HawkbitUIErrorHandler.java
```diff
@@ -0,0 +1,101 @@
...
+import com.vaadin.ui.Component;
+import com.vaadin.ui.themes.ValoTheme;
+
+/**
+ * Default handler for SP UI.
+ */
+public class HawkbitUIErrorHandler extends DefaultErrorHandler {
+
+    /**
+     * Comment for <code>serialVersionUID</code>.
```
schabdo: Remove odd comment

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/components/HawkbitUIErrorHandler.java
```diff
@@ -0,0 +1,101 @@
...
+     *            ErrorEvent
+     * @return name of the thrown exception
+     */
+    protected String getExceptionName(final ErrorEvent event) {
+        for (Throwable error = event.getThrowable(); error != null; error = error.getCause()) {
+            exceptionName = HawkbitCommonUtil.getLastSequenceBySplitByDot(error.getClass().getName());
+            exceptionMessage = error.getMessage();
+            LOG.error("Error in SP-UI:", error);
+        }
+        return exceptionName;
```
schabdo: You do not return the name from `ErrorEvent` every time! So why I have to provide it for the method call?

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/components/HawkbitUIErrorHandler.java
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    /**
+     * Determines the error origin and checks if it is not null
+     * 
+     * @param event
+     *            ErrorEvent
+     * @return boolean if error Origin is not null
+     */
+    protected Boolean getAndCheckErrorOrigin(final ErrorEvent event) {
```
schabdo: Name is confusing. Can you find a better one?

michahirsch:
What's the different between the `HawkbitNotificationMessage` class and the `NotificationMessage` seems like there are doing the same things, right? The `decorate` method is exactly the same.

As I understand the `HawkbitNotificationMessage` is only used for showing error notifications. So either we can declare one notification only to error like `HawkbitErrorNotificationMessage`.

But anyway I don't like the way that the only have a `decorate` method. Why the business logic like showing the notification has been removed from the `HawkbitNotificationMessage` ? 

I think it makes sense to instantiate thy type of notification I need, like `new ErrorNotification()` and the class knows it's style etc.

michahirsch:
What has been really changed on style to before? The styles used for the error notification has been the same, don't they?

melanieretter:
I renamed `HawkbitNotificationMessage `to `HawkbitErrorNotificationMessage`. The `HakwbitErrorNotificationMessage `is used in the standard errorHandler of the UI. It is used when an unexpected exception occures. The `NotificationMessage `is used when there is an expected (validation) error or notification like saving successful. I removed the decorate method in `HawkbitErrorNotificationMessage `and expand the constructor. The properties style and autoclose has to be in the constructor so that it is possible to overwrite them in case there is a different look desired.

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/components/HawkbitErrorNotificationMessage.java
```diff
@@ -0,0 +1,43 @@
...
+
+    /**
+     * ID.
+     */
+    private static final long serialVersionUID = -6512576924243195753L;
+
+    /**
+     * Constructor.
+     */
+    public HawkbitErrorNotificationMessage(final String style, final String caption, final String description,
```
michahirsch: Missing javadoc here.

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/components/HawkbitErrorNotificationMessage.java
```diff
@@ -0,0 +1,43 @@
...
+
+import com.vaadin.shared.Position;
+import com.vaadin.ui.Notification;
+
+/**
+ * Notification message component.
+ */
+public class HawkbitErrorNotificationMessage extends Notification {
+
+    /**
```
michahirsch: This javadoc can be removed.

PR #260, opened 2016-07-29
Allow to set error handler or take default one

Jonkno:
With this feature it is now possible to set a custom error handler. If no error hander is set the default one will be used. 

### DIFF
--- a/hawkbit-dmf-amqp/src/main/java/org/eclipse/hawkbit/amqp/AmqpConfiguration.java
+++ b/hawkbit-dmf-amqp/src/main/java/org/eclipse/hawkbit/amqp/AmqpConfiguration.java
@@ -267,7 +267,9 @@ public class AmqpConfiguration {
 
     /**
      * Returns the Listener factory.
-     *
+     * 
+     * @param errorHandler
+     *            the error hander
      * @return the {@link SimpleMessageListenerContainer} that gets used receive
      *         AMQP messages
      */
--- a/hawkbit-dmf-amqp/src/main/java/org/eclipse/hawkbit/amqp/AmqpConfiguration.java
+++ b/hawkbit-dmf-amqp/src/main/java/org/eclipse/hawkbit/amqp/AmqpConfiguration.java
@@ -60,8 +60,8 @@ public class AmqpConfiguration {
     @Autowired
     private ConnectionFactory rabbitConnectionFactory;
 
-    @Autowired
-    private ErrorHandler errorHandler;
+    // @Autowired
+    // private ErrorHandler errorHandler;
 
     @Configuration
     @ConditionalOnMissingBean(ConnectionFactory.class)
@@ -273,7 +273,8 @@ public class AmqpConfiguration {
      *         AMQP messages
      */
     @Bean(name = { "listenerContainerFactory" })
-    public RabbitListenerContainerFactory<SimpleMessageListenerContainer> listenerContainerFactory() {
+    public RabbitListenerContainerFactory<SimpleMessageListenerContainer> listenerContainerFactory(
+            final ErrorHandler errorHandler) {
         return new ConfigurableRabbitListenerContainerFactory(amqpProperties, rabbitConnectionFactory, errorHandler);
     }
 
--- a/hawkbit-dmf-amqp/src/main/java/org/eclipse/hawkbit/amqp/AmqpConfiguration.java
+++ b/hawkbit-dmf-amqp/src/main/java/org/eclipse/hawkbit/amqp/AmqpConfiguration.java
@@ -26,7 +26,6 @@ import org.springframework.amqp.rabbit.connection.CachingConnectionFactory;
 import org.springframework.amqp.rabbit.connection.ConnectionFactory;
 import org.springframework.amqp.rabbit.core.RabbitAdmin;
 import org.springframework.amqp.rabbit.core.RabbitTemplate;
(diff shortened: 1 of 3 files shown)


hawkbit-dmf-amqp/src/main/java/org/eclipse/hawkbit/amqp/AmqpConfiguration.java
```diff
@@ -58,6 +60,9 @@
     @Autowired
     private ConnectionFactory rabbitConnectionFactory;
 
+    @Autowired
+    private ErrorHandler errorHandler;
```
schabdo: Are you really need this variable? Just call

`return new ConfigurableRabbitListenerContainerFactory(amqpProperties, rabbitConnectionFactory, errorHandler());`

Jonkno: Yes I need it, if I would use the errorHandler() method the RabbitListenerContainerFactory would always use the default error handler (ConditionalRejectingErrorHandler) and we could not provide a custom one. By autowiring the error handler Spring will take the one in the application context which is either a custom one that is been created or the default one if none has been created.

SirWayne: @schabdo this was my first thought when I saw the row. It's difficult to understand and confusing.

hawkbit-dmf-amqp/src/main/java/org/eclipse/hawkbit/amqp/ConfigurableRabbitListenerContainerFactory.java
```diff
@@ -30,8 +31,9 @@
      *            to configure the container factory
      */
     public ConfigurableRabbitListenerContainerFactory(final AmqpProperties amqpProperties,
-            final ConnectionFactory rabbitConnectionFactory) {
+            final ConnectionFactory rabbitConnectionFactory, final ErrorHandler errorHandler) {
```
SirWayne: Miss javadoc change

hawkbit-dmf-amqp/src/main/java/org/eclipse/hawkbit/amqp/AmqpConfiguration.java
michahirsch: Calling the private method won't work yes, because there is no spring-proxy around which could check if there is already a bean `errorHandler` 

One way could also be setting the parameter to the method 
```
@Bean(name = { "listenerContainerFactory" })
    public RabbitListenerContainerFactory<SimpleMessageListenerContainer> listenerContainerFactory(
            final ErrorHandler errorHandler) {
        return new ConfigurableRabbitListenerContainerFactory(amqpProperties, rabbitConnectionFactory, errorHandler);
    }
```

but in general it looks weird having the default `errorHandler` within the AmqpConfiguration, what about moving the 
```
    @Bean
    @ConditionalOnMissingBean
    public ErrorHandler errorHandler() {
        return new ConditionalRejectingErrorHandler();
    }

```
to the autoconfiguration or just handle `null` because `null` is already handled in the `AbstractRabbitListenerContainerFactory`

```
@Override
	public C createListenerContainer(RabbitListenerEndpoint endpoint) {
		C instance = createContainerInstance();

		if (this.connectionFactory != null) {
			instance.setConnectionFactory(this.connectionFactory);
		}
		if (this.errorHandler != null) {
			instance.setErrorHandler(this.errorHandler);
		}
```

And if the `errorHandler` is not set then the default is used 
```
private ErrorHandler errorHandler = new ConditionalRejectingErrorHandler();
```

SirWayne: ok my proposal is:
```
    @Bean(name = { "listenerContainerFactory" })
    public RabbitListenerContainerFactory<SimpleMessageListenerContainer> listenerContainerFactory(
            final ErrorHandler errorHandler) {
        return new ConfigurableRabbitListenerContainerFactory(amqpProperties, rabbitConnectionFactory, errorHandler);
    }
```

schabdo: @SirWayne I like your approach :+1:

hawkbit-dmf-amqp/src/main/java/org/eclipse/hawkbit/amqp/AmqpConfiguration.java
```diff
@@ -37,7 +38,8 @@
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.retry.backoff.ExponentialBackOffPolicy;
-import org.springframework.retry.support.RetryTemplate;;
+import org.springframework.retry.support.RetryTemplate;
+import org.springframework.util.ErrorHandler;;
```
melanieretter: remove double semicolon

hawkbit-dmf-amqp/src/main/java/org/eclipse/hawkbit/amqp/AmqpConfiguration.java
```diff
@@ -58,6 +59,9 @@
     @Autowired
     private ConnectionFactory rabbitConnectionFactory;
 
+    // @Autowired
```
SirWayne: remove

PR #262, opened 2016-08-04
Fix enable showing errors during bulk upload

Jonkno:
With this fix errors triggered by RuntimeExceptions will be shown on UI during bulk upload. If such an error is thrown the bulk upload dialog will return to a valid state:

### DIFF
--- a/hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/management/targettable/BulkUploadHandler.java
+++ b/hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/management/targettable/BulkUploadHandler.java
@@ -22,6 +22,7 @@ import java.util.ArrayList;
 import java.util.Date;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.concurrent.Executor;
 
 import org.eclipse.hawkbit.repository.DeploymentManagement;
@@ -56,6 +57,7 @@ import com.vaadin.ui.HorizontalLayout;
 import com.vaadin.ui.Label;
 import com.vaadin.ui.ProgressBar;
 import com.vaadin.ui.TextArea;
+import com.vaadin.ui.UI;
 import com.vaadin.ui.Upload;
 import com.vaadin.ui.Upload.FailedEvent;
 import com.vaadin.ui.Upload.FailedListener;
@@ -99,7 +101,7 @@ public class BulkUploadHandler extends CustomComponent
     private final transient Executor executor;
     private transient EventBus.SessionEventBus eventBus;
 
-    final TargetBulkUpdateWindowLayout targetBulkUpdateWindowLayout;
+    TargetBulkUpdateWindowLayout targetBulkUpdateWindowLayout;
 
     private transient EntityFactory entityFactory;
 
@@ -139,7 +141,7 @@ public class BulkUploadHandler extends CustomComponent
     public void buildLayout() {
         final HorizontalLayout horizontalLayout = new HorizontalLayout();
         upload = new Upload();
-        upload.setEnabled(false);
+        upload.setEnabled(true);
         upload.setButtonCaption("Bulk Upload");
         upload.setReceiver(this);
         upload.setImmediate(true);
@@ -178,6 +180,7 @@ public class BulkUploadHandler extends CustomComponent
     }
 
     class UploadAsync implements Runnable {
+
         final SucceededEvent event;
 
         /**
@@ -193,7 +196,6 @@ public class BulkUploadHandler extends CustomComponent
             if (tempFile == null) {
                 return;
             }
-
             try (InputStream tempStream = new FileInputStream(tempFile)) {
                 readFileStream(tempStream);
             } catch (final FileNotFoundException e) {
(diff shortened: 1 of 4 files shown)


hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/management/targettable/BulkUploadHandler.java
```diff
@@ -218,18 +216,24 @@ private void readFileStream(final InputStream tempStream) {
...
-                doAssignments();
-                eventBus.publish(this, new TargetTableEvent(TargetComponentEvent.BULK_UPLOAD_COMPLETED));
 
-                // Clearing after assignments are done
-                managementUIState.getTargetTableFilters().getBulkUpload().getTargetsCreated().clear();
             } catch (final IOException e) {
                 LOG.error("Error reading file {}", tempFile.getName(), e);
+            } catch (final RuntimeException e) {
+                Optional.ofNullable(UI.getCurrent()).ifPresent(error -> UI.getCurrent().getErrorHandler()
+                        .error(new ConnectorErrorEvent(targetTableHeader, e)));
```
michahirsch: Does it need to be a `ConnectorEvent`? And way we giving the `targetTableHeader` in there, this looks a little bit strange to me.

michahirsch:
The dependencies between the classes `TargetTableHeader` and `TargetBulkUpdateWindowLayout` is weird. These classes inject each other and holding each other as fields.

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/management/targettable/BulkUploadHandler.java
Jonkno: Yes, our current error handler identifies that component in a first step and based on the component the right page to show error.
`
 public static AbstractComponent findAbstractComponent(
            com.vaadin.server.ErrorEvent event) {
        if (event instanceof ConnectorErrorEvent) {
            Component c = findComponent(((ConnectorErrorEvent) event)
                    .getConnector());
            if (c instanceof AbstractComponent) {
                return (AbstractComponent) c;
            }
        }
        return null;
    }
`
It has to be a component that is not part of the bulk upload dialog so that the error still emerges even if the bulk upload dialog has been minimized.

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/management/targettable/BulkUploadHandler.java
```diff
@@ -218,18 +216,24 @@ private void readFileStream(final InputStream tempStream) {
...
                 }
-                doAssignments();
-                eventBus.publish(this, new TargetTableEvent(TargetComponentEvent.BULK_UPLOAD_COMPLETED));
 
-                // Clearing after assignments are done
-                managementUIState.getTargetTableFilters().getBulkUpload().getTargetsCreated().clear();
             } catch (final IOException e) {
                 LOG.error("Error reading file {}", tempFile.getName(), e);
+            } catch (final RuntimeException e) {
+                Optional.ofNullable(UI.getCurrent()).ifPresent(error -> UI.getCurrent().getErrorHandler()
```
schabdo: I appreciate the use of `Òptional` as a return value of a function. But I'd prefer the use of a simple if-statement here:
```
if (UI.getCurrent() != null) {
    UI.getCurrent().getErrorHandler().error(new ConnectorErrorEvent(targetTableHeader, e));
}
```

SirWayne: And save the UI in a local variable

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/management/targettable/BulkUploadHandler.java
```diff
@@ -218,18 +216,24 @@ private void readFileStream(final InputStream tempStream) {
...
+                        .error(new ConnectorErrorEvent(targetTableHeader, e)));
+
             } finally {
+                updateBulkUpload();
+                doAssignments();
```
schabdo: This method call looks strange to me. In case an exception occured, you're doing the assinment anyway?

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/management/targettable/BulkUploadHandler.java
```diff
@@ -283,13 +287,16 @@ private void readEachLine(final String line, final double innerCounter, final do
...
-                managementUIState.getTargetTableFilters().getBulkUpload().setFailedUploadCount(failedTargetCount);
-                eventBus.publish(this, new TargetTableEvent(TargetComponentEvent.BULK_TARGET_CREATED));
+                updateBulkUpload();
             }
         }
 
+        private void updateBulkUpload() {
+            managementUIState.getTargetTableFilters().getBulkUpload().setSucessfulUploadCount(successfullTargetCount);
+            managementUIState.getTargetTableFilters().getBulkUpload().setFailedUploadCount(failedTargetCount);
+            eventBus.publish(this, new TargetTableEvent(TargetComponentEvent.BULK_TARGET_CREATED));
```
schabdo: How do you now here, that a target was created. This will be called within a `finally`-block (see line 232)

Jonkno:
Fixed

Jonkno:
@michahirsch fixed your last comment

PR #269, opened 2016-08-11
introduce property to disable amqp interface in simulator

michahirsch:
Introducing a configuration parameter to allow to disable the AMQP interface if the AMQP interface is not necessary. In case there is not RabbitMQ message broker running the device simulator tries to connect to the default configuration on localhost. This will log connection reset exception on each try and so having a lot of exception in the log. 

* Introducing the `hawkbit.device.simulator.amqp.enabled=true` configuration which can be set via properties, environment etc.
* Added exception handling when trying to create DMF simulated devices via rest 
`The AMQP interface has been disabled, to use DMF protocol you need to enable the AMQP interface via 'hawkbit.device.simulator.amqp.enabled=true'`
* Handling the disabling of the AMQP interface in the simulator UI

image

* Updated `README.MD`

### DIFF
--- a/examples/hawkbit-device-simulator/README.md
+++ b/examples/hawkbit-device-simulator/README.md
@@ -24,6 +24,12 @@ The simulator has user authentication enabled in **cloud profile**. Default cred
 
 This can be configured/disabled by spring boot properties
 
+## hawkBit APIs
+
+The simulator supports `DDI` as well as the `DMF` integration APIs.
+In case there is no AMQP message broker (like rabbitMQ) running, you can disable the AMQP support for the device simulator, so the simulator is not trying to connect to a amqp message broker. 
+Configuration property `hawkbit.device.simulator.amqp.enabled=true` 
+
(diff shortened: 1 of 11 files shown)


examples/hawkbit-device-simulator/README.md
```diff
@@ -24,6 +24,12 @@ The simulator has user authentication enabled in **cloud profile**. Default cred
 
 This can be configured/disabled by spring boot properties
 
+## hawkBit APIs
+
+The simulator supports `DDI` as well as the `DMF` integration APIs.
+In case there is no AMQP message broker (like rabbitMQ) running, you can disable the AMQP support for the device simulator, so the simulator is not trying to connect to a amqp message broker.
```
melanieretter: **an** amqp

Jonkno:
Please also update the manifest.yml in the cf folder (remove SPRING_PROFILES_ACTIVE)

PR #270, opened 2016-08-11
Show error notification needs a lock when run in separate thread

melanieretter:
Extend the HawkbitUiErrorHandler, so it is possible to show an error notification if it runs in a separate thread. Furthermore the error notification is shown in the appropriate browser window (in case several browser windows are open)

### DIFF
--- a/hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/components/HawkbitUIErrorHandler.java
+++ b/hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/components/HawkbitUIErrorHandler.java
@@ -40,27 +40,26 @@ public class HawkbitUIErrorHandler extends DefaultErrorHandler {
     @Override
     public void error(final ErrorEvent event) {
 
-        final Optional<Page> originError = getPageOriginError(event);
         final HawkbitErrorNotificationMessage message = buildNotification(getRootExceptionFrom(event));
-
-        if (!originError.isPresent()) {
-            HawkbitErrorNotificationMessage.show(message.getCaption(), message.getDescription(),
-                    Type.HUMANIZED_MESSAGE);
-            return;
-        }
-
         if (event instanceof ConnectorErrorEvent) {
             final Connector connector = ((ConnectorErrorEvent) event).getConnector();
             if (connector instanceof UI) {
-                ((UI) connector).access(() -> message.show(originError.get()));
+                final UI uiInstance = (UI) connector;
+                uiInstance.access(() -> message.show(uiInstance.getPage()));
                 return;
             }
         }
-        message.show(originError.get());
+
+        final Optional<Page> originError = getPageOriginError(event);
+        if (originError.isPresent()) {
+            message.show(originError.get());
+            return;
+        }
+
+        HawkbitErrorNotificationMessage.show(message.getCaption(), message.getDescription(), Type.HUMANIZED_MESSAGE);
     }
 
     private static Throwable getRootExceptionFrom(final ErrorEvent event) {
-
         return getRootCauseOf(event.getThrowable());
     }
 
@@ -84,16 +83,18 @@ public class HawkbitUIErrorHandler extends DefaultErrorHandler {
         return Optional.absent();
     }
 
+    /**
+     * Method to build a notification based on an exception.
+     * 
+     * @param ex
+     *            the throwable
+     * @return a hawkbit error notification message
+     */
     protected HawkbitErrorNotificationMessage buildNotification(final Throwable ex) {
-
-        log(ex);
-
+        LOG.error("Error in UI: ", ex);
         final I18N i18n = SpringContextHelper.getBean(I18N.class);
         return new HawkbitErrorNotificationMessage(STYLE, i18n.get("caption.error"),
                 i18n.get("message.error.temp", ex.getClass().getSimpleName()), false);
     }
 
-    protected void log(final Throwable exception) {
-        LOG.error("Error in UI: ", exception);
-    }
 }
--- a/hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/components/HawkbitUIErrorHandler.java
+++ b/hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/components/HawkbitUIErrorHandler.java
@@ -83,6 +83,10 @@ public class HawkbitUIErrorHandler extends DefaultErrorHandler {
 
     protected HawkbitErrorNotificationMessage buildNotification(final Throwable exception) {
         LOG.error("Error in UI: ", exception);
+        return createHawkbitErrorNotificationMessage(exception);
+    }
+
+    protected HawkbitErrorNotificationMessage createHawkbitErrorNotificationMessage(final Throwable exception) {
         final I18N i18n = SpringContextHelper.getBean(I18N.class);
         return new HawkbitErrorNotificationMessage(STYLE, i18n.get("caption.error"),
                 i18n.get("message.error.temp", exception.getClass().getSimpleName()), false);
--- a/hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/components/HawkbitUIErrorHandler.java
+++ b/hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/components/HawkbitUIErrorHandler.java
@@ -18,10 +18,15 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.google.common.base.Optional;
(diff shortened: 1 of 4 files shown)


hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/management/targettable/BulkUploadHandler.java
```diff
@@ -111,8 +112,9 @@
      */
     public BulkUploadHandler(final TargetBulkUpdateWindowLayout targetBulkUpdateWindowLayout,
             final TargetManagement targetManagement, final ManagementUIState managementUIState,
-            final DeploymentManagement deploymentManagement, final I18N i18n) {
+            final DeploymentManagement deploymentManagement, final I18N i18n, final UI uiInstance) {
```
schabdo: The value is always `UI.getCurrent()` do you realy need this parameter?

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/components/HawkbitUIErrorHandler.java
```diff
@@ -38,10 +42,21 @@ public void error(final ErrorEvent event) {
...
 
+        final HawkbitErrorNotificationMessage message = buildNotification(getRootExceptionFrom(event));
+
         if (originError.isPresent()) {
-            final HawkbitErrorNotificationMessage message = buildNotification(getRootExceptionFrom(event));
+            final Connector connector = ((ConnectorErrorEvent) event).getConnector();
+            // in case of BulkUpload: BulkUpload needs a lock to show the
+            // notification
+            if (connector instanceof UI) {
+                ((UI) (((ConnectorErrorEvent) event).getConnector())).access(() -> message.show(originError.get()));
```
schabdo: Why are you casting twice in one line? This is hard to understand/read. BTW you casted to `ConnectorErrorEvent` already in line 48!

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/components/HawkbitUIErrorHandler.java
```diff
@@ -38,10 +42,21 @@ public void error(final ErrorEvent event) {
 
         final Optional<Page> originError = getPageOriginError(event);
 
+        final HawkbitErrorNotificationMessage message = buildNotification(getRootExceptionFrom(event));
+
         if (originError.isPresent()) {
-            final HawkbitErrorNotificationMessage message = buildNotification(getRootExceptionFrom(event));
+            final Connector connector = ((ConnectorErrorEvent) event).getConnector();
+            // in case of BulkUpload: BulkUpload needs a lock to show the
```
schabdo: Do not mention a specific view here. How do you know in that line that the view even exists?

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/components/HawkbitUIErrorHandler.java
schabdo: I don't get the difference between line 52 and 55. Where is the mentioned lock? You are showing the message anyway !?!?

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/components/HawkbitUIErrorHandler.java
```diff
@@ -38,10 +42,21 @@ public void error(final ErrorEvent event) {
 
         final Optional<Page> originError = getPageOriginError(event);
 
+        final HawkbitErrorNotificationMessage message = buildNotification(getRootExceptionFrom(event));
+
         if (originError.isPresent()) {
-            final HawkbitErrorNotificationMessage message = buildNotification(getRootExceptionFrom(event));
+            final Connector connector = ((ConnectorErrorEvent) event).getConnector();
```
schabdo: How do you know that the event is from type `ConnectorErrorEvent`?

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/components/HawkbitUIErrorHandler.java
```diff
@@ -38,10 +42,21 @@ public void error(final ErrorEvent event) {
...
+            // in case of BulkUpload: BulkUpload needs a lock to show the
+            // notification
+            if (connector instanceof UI) {
+                ((UI) (((ConnectorErrorEvent) event).getConnector())).access(() -> message.show(originError.get()));
+                return;
+            }
             message.show(originError.get());
+            return;
         }
+        // Default
```
schabdo: remove => superfluous

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/components/HawkbitUIErrorHandler.java
```diff
@@ -38,10 +42,21 @@ public void error(final ErrorEvent event) {
 
         final Optional<Page> originError = getPageOriginError(event);
 
+        final HawkbitErrorNotificationMessage message = buildNotification(getRootExceptionFrom(event));
```
schabdo: `buildNotification(...)` only adds a log statement can't you move the log-statement some where else and remove the seperate method?

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/components/HawkbitUIErrorHandler.java
melanieretter: It has to be, because otherwise the originError would be null. Have a look in the method getPageOriginError and inside that in findAbstractComponent...

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/components/HawkbitUIErrorHandler.java
melanieretter: The method access locks the session

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/management/targettable/BulkUploadHandler.java
melanieretter: Yes, because otherwise you would not see the notification in the appropriate browser window if you are working with more than one browser in the same session.

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/components/HawkbitUIErrorHandler.java
schabdo: Well, I see. But I'd not rely on that knowlege. This dependency is hard to figure out and if you'll only look at class level `ConnectorErrorEvent` comes never into play. `findAbstractComponent(...)` is returning a `Component`. JavaDoc is also not pointing out that contract:

```
final Component errorOrigin = findAbstractComponent(event);

if (errorOrigin != null && errorOrigin.getUI() != null) {
     return Optional.fromNullable(errorOrigin.getUI().getPage());
}

return Optional.absent();  
```

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/components/HawkbitUIErrorHandler.java
Jonkno: The '''buildNotification''' method logs the error and creates the notification message. If we don't want this behavior the method can be overwritten (e. g. creating a notification message but not logging the error). Default is logging the error. If I remove the log somewhere else, we cannot overwrite it anymore in the current implementation.

michahirsch:
What about changing the `error` method like this? I haven't tried it but it seems more readable than the other by using this `orginError.get()` when the `connector instanceof UI` then we should retrieve the page from the `uiInstance` why taking it from the `orginError`? 

```
 @Override
    public void error(final ErrorEvent event) {
        final HawkbitErrorNotificationMessage message = buildNotification(getRootExceptionFrom(event));
        if (event instanceof ConnectorErrorEvent) {
            final Connector connector = ((ConnectorErrorEvent) event).getConnector();
            if (connector instanceof UI) {
                final UI uiInstance = (UI) connector;
                uiInstance.access(() -> message.show(uiInstance.getPage()));
                return;
            }
        }

        final Optional<Page> originError = getPageOriginError(event);
        if (originError.isPresent()) {
            message.show(originError.get());
            return;
        }
        HawkbitErrorNotificationMessage.show(message.getCaption(), message.getDescription(), Type.HUMANIZED_MESSAGE);
    }
```

PR #276, opened 2016-08-18
Removed unnecessary target info startus persistence.

kaizimmerm:
Is actually not needed. Target is at that point of the state machine already in PENDING.

While at it I fixed two problems on the way:
* DMF is not updating last target poll in case of EVENT topic.
* DMF is sending internal DS assignment event duplicated. 

### DIFF
--- a/hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaControllerManagement.java
+++ b/hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaControllerManagement.java
@@ -299,8 +299,6 @@ public class JpaControllerManagement implements ControllerManagement {
         case CANCELED:
         case WARNING:
         case RUNNING:
-            handleIntermediateFeedback(mergedAction, mergedTarget);
-            break;
         default:
             break;
         }
@@ -310,16 +308,6 @@ public class JpaControllerManagement implements ControllerManagement {
         LOG.debug("addUpdateActionStatus {} for target {} is finished.", action.getId(), mergedTarget.getId());
 
         return actionRepository.save(mergedAction);
-    }
-
-    private void handleIntermediateFeedback(final JpaAction mergedAction, final JpaTarget mergedTarget) {
-        // we change the target state only if the action is still running
-        // otherwise this is considered as late feedback that does not have
-        // an impact on the state anymore.
-        if (mergedAction.isActive()) {
-            DeploymentHelper.updateTargetInfo(mergedTarget, TargetUpdateStatus.PENDING, false, targetInfoRepository,
-                    entityManager);
-        }
     }
 
     private void handleErrorOnAction(final JpaAction mergedAction, final JpaTarget mergedTarget) {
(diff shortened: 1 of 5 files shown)


hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaControllerManagement.java
```diff
@@ -349,15 +337,17 @@ private void handleFinishedAndStoreInTargetStatus(final JpaTarget target, final
         action.setStatus(Status.FINISHED);
         final JpaTargetInfo targetInfo = (JpaTargetInfo) target.getTargetInfo();
         final JpaDistributionSet ds = (JpaDistributionSet) entityManager.merge(action.getDistributionSet());
+
         targetInfo.setInstalledDistributionSet(ds);
-        if (target.getAssignedDistributionSet() != null && targetInfo.getInstalledDistributionSet() != null && target
-                .getAssignedDistributionSet().getId().equals(targetInfo.getInstalledDistributionSet().getId())) {
+        targetInfo.setInstallationDate(System.currentTimeMillis());
+
+        // check if the assigned set is equal no to the installed set (not
```
SirWayne: I would prefer a methode for that instead of a comment

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaControllerManagement.java
```diff
@@ -349,15 +337,17 @@ private void handleFinishedAndStoreInTargetStatus(final JpaTarget target, final
...
+        // necessarily the case as another update might be pending already).
+        if (target.getAssignedDistributionSet() != null && target.getAssignedDistributionSet().getId()
+                .equals(targetInfo.getInstalledDistributionSet().getId())) {
```
SirWayne: The `targetInfo.getInstalledDistributionSet()` cannot bet null?

kaizimmerm: No, its set two lines above.

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaControllerManagement.java
kaizimmerm: How do you do the braces in a method name? 😄

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaControllerManagement.java
SirWayne: Proposal:

```
    private boolean isUpdateStatusInSync(...){
     boolean isAssignedDsEqualToInstalledDs = target.getAssignedDistributionSet() != null && target.getAssignedDistributionSet().getId()
             .equals(targetInfo.getInstalledDistributionSet().getId());
        return isAssignedDsEqualToInstalledDs ;
    }
```

PR #279, opened 2016-08-23
Fix negative max assignments of softwareModuleTypes

melanieretter:
Currently you are able to create a software module type via REST API with a negative number or zero of assignments. This doesn't make sense. Thus the request will be rejected and returns a response of 400 Bad Request.


SirWayne:
I would add a constraint on the database column.

SirWayne:
Please add a unit test for SoftwareManagement#createSoftwareModuleType

hawkbit-repository/hawkbit-repository-jpa/src/main/resources/db/migration/MYSQL/V1_9_0__software_module_type_constraint__MYSQL.sql
```diff
@@ -0,0 +1,2 @@
+alter table sp_software_module_type
```
SirWayne: Why version 1_9_0? The next version is 1_7_0 or?

hawkbit-repository/hawkbit-repository-api/src/main/java/org/eclipse/hawkbit/repository/exception/ConstraintViolationException.java
```diff
@@ -0,0 +1,49 @@
...
+     * 
+     * @param ex
+     *            the javax.validation.ConstraintViolationException which is
+     *            thrown
+     */
+    public ConstraintViolationException(final javax.validation.ConstraintViolationException ex) {
+        super(setExceptionMessage(ex), SpServerError.SP_REPO_CONSTRAINT_VIOLATION);
+    }
+
+    public static String setExceptionMessage(final javax.validation.ConstraintViolationException ex) {
```
SirWayne: Why is it a setter? It's a getter or?
A setter with a return value looks weird

hawkbit-repository/hawkbit-repository-api/src/main/java/org/eclipse/hawkbit/repository/exception/ConstraintViolationException.java
```diff
@@ -0,0 +1,49 @@
...
+        final Set<ConstraintViolation<?>> violations = ex.getConstraintViolations();
+        final List<String> messages = new ArrayList<>();
+        violations.stream()
+                .forEach(violation -> messages.add(violation.getPropertyPath() + " " + violation.getMessage() + "."));
```
SirWayne: Use a constant or local variable for `" "`. It's duplicated

PR #283, opened 2016-09-05
Fixed DMF sending out wrong message if action on CANCELING

kaizimmerm:
DMF is sending out an UPDATE message even when the action is on CANCELING. Fixes this.

While at it I also optimised the same functionality in DDI by means of letting the database filtering out the oldest action instead of getting all active actions and let the JVM filter them. Upgrade to newest MariaDB driver was done as well.

### DIFF
--- a/hawkbit-repository/hawkbit-repository-api/src/main/java/org/eclipse/hawkbit/repository/ControllerManagement.java
+++ b/hawkbit-repository/hawkbit-repository-api/src/main/java/org/eclipse/hawkbit/repository/ControllerManagement.java
@@ -11,6 +11,7 @@ package org.eclipse.hawkbit.repository;
 import java.net.URI;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 
 import javax.validation.constraints.NotNull;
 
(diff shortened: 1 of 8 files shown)


hawkbit-repository/hawkbit-repository-api/src/main/java/org/eclipse/hawkbit/repository/ControllerManagement.java
```diff
@@ -114,6 +115,17 @@
...
     /**
+     * Retrieves oldest {@link Action} that is active and assigned to a
+     * {@link Target}.
+     *
+     * @param target
+     *            the target to retrieve the actions from
+     * @return a list of actions assigned to given target which are active
+     */
+    @PreAuthorize(SpringEvalExpressions.IS_CONTROLLER)
+    Optional<Action> findOldestActionByTargetAndActive(@NotNull Target target);
```
schabdo: A better name would be `findOldestActiveActionByTarget`.  You don't have to follow the Spring-Data naming conventions here. Take also `findActionByTargetAndActive` into account => `findActiveActionByTarget`

PR #284, opened 2016-09-05
Fixed entity content validation.

kaizimmerm:
Most entity properties where not validated and as a result the error was catched on DB layer which resulted in a wrong error message to the user.

- I fixed this by introducing Java Bean validation where missing.
- While at it I fixed also an issue with far to big columns for String based enumerations and set not null constraint on DB for tenant config the same way as it is for metadata.


hawkbit-mgmt-resource/src/main/java/org/eclipse/hawkbit/mgmt/rest/resource/MgmtDistributionSetTypeMapper.java
```diff
@@ -54,26 +55,30 @@ static DistributionSetType fromRequest(final EntityFactory entityFactory,
                 smsRest.getName(), smsRest.getDescription());
 
         // Add mandatory
-        smsRest.getMandatorymodules().stream().map(mand -> {
-            final SoftwareModuleType smType = softwareManagement.findSoftwareModuleTypeById(mand.getId());
+        if (!CollectionUtils.isEmpty(smsRest.getMandatorymodules())) {
```
schabdo: Better extract this block to a private method and remove the odd comment `Add mandatory`

hawkbit-mgmt-resource/src/main/java/org/eclipse/hawkbit/mgmt/rest/resource/MgmtDistributionSetTypeMapper.java
```diff
@@ -54,26 +55,30 @@ static DistributionSetType fromRequest(final EntityFactory entityFactory,
...
-            return smType;
-        }).forEach(result::addMandatoryModuleType);
+                return smType;
+            }).forEach(result::addMandatoryModuleType);
+        }
 
         // Add optional
-        smsRest.getOptionalmodules().stream().map(opt -> {
-            final SoftwareModuleType smType = softwareManagement.findSoftwareModuleTypeById(opt.getId());
+        if (!CollectionUtils.isEmpty(smsRest.getOptionalmodules())) {
```
schabdo: Same here

PR #286, opened 2016-09-07
DMF can be disabled at runtime by property

kaizimmerm:
Allows to disable DMF by property as alternative to auto configuration.

Removed a few service annotations to get this done.

### DIFF
--- a/hawkbit-dmf-amqp/src/main/java/org/eclipse/hawkbit/amqp/AmqpConfiguration.java
+++ b/hawkbit-dmf-amqp/src/main/java/org/eclipse/hawkbit/amqp/AmqpConfiguration.java
@@ -53,7 +53,7 @@ import org.springframework.util.ErrorHandler;
  *
  */
 @EnableConfigurationProperties({ AmqpProperties.class, AmqpDeadletterProperties.class })
-@ConditionalOnProperty(prefix = "hawkbit.dmf.rabbitmq", name = "enabled")
+@ConditionalOnProperty(prefix = "hawkbit.dmf.rabbitmq", name = "enabled", matchIfMissing = true)
 public class AmqpConfiguration {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(AmqpConfiguration.class);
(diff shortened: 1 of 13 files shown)


hawkbit-dmf-amqp/src/test/java/org/eclipse/hawkbit/amqp/AmqpMessageDispatcherServiceTest.java
```diff
@@ -79,17 +79,16 @@ public void before() throws Exception {
...
-        amqpMessageDispatcherService.setAmqpSenderService(senderService);
 
         final ArtifactUrlHandler artifactUrlHandlerMock = Mockito.mock(ArtifactUrlHandler.class);
         when(artifactUrlHandlerMock.getUrl(anyString(), anyLong(), anyString(), anyString(), anyObject()))
                 .thenReturn("http://mockurl");
 
-        amqpMessageDispatcherService.setArtifactUrlHandler(artifactUrlHandlerMock);
+        amqpMessageDispatcherService = new AmqpMessageDispatcherService(rabbitTemplate, senderService,
+                artifactUrlHandlerMock);
+        amqpMessageDispatcherService = spy(amqpMessageDispatcherService);
```
schabdo: Why do you spy the `amqpMessageDispatcherService` here? Is it really required?

hawkbit-dmf-amqp/src/main/java/org/eclipse/hawkbit/amqp/AmqpConfiguration.java
```diff
@@ -61,6 +69,7 @@
 
     @Configuration
     @ConditionalOnMissingBean(ConnectionFactory.class)
+    @ConditionalOnProperty(prefix = "hawkbit.dmf.rabbitmq", name = "enabled", matchIfMissing = true)
```
michahirsch: I haven't tried that, but I think it's not necessary to use `@ConditionalOnProperty` on the inner class configuration because the outer configuration already has the same conditional, maybe I'm wrong but I thought it's not necessary.

kaizimmerm: Nope, its necessary.

michahirsch: ok :)

PR #289, opened 2016-09-12
Fix Mgmt Client Netflix Feign usage

kaizimmerm:
It seems the Boot 1.3 upgrade broke the Spring Cloud Netflix integration as they dependend on different Spring Framework versions.

I upgraded to the newest Spring Cloud version.


examples/hawkbit-example-core-feign-client/src/main/java/org/eclipse/hawkbit/feign/core/client/IgnoreMultipleConsumersProducersSpringMvcContract.java
```diff
@@ -38,6 +40,7 @@ protected void processAnnotationOnMethod(final MethodMetadata data, final Annota
             // multiple consumers and produces, see
             // https://github.com/spring-cloud/spring-cloud-netflix/issues/808
             LOGGER.trace(e.getMessage(), e);
+            data.indexToExpander(new LinkedHashMap<Integer, Param.Expander>());
```
schabdo: Can you explain/document why exactly this line from super-method is necessary? What about `parseHeaders(data, method, methodMapping);`?

PR #294, opened 2016-09-20
Performance and memory optimisation

kaizimmerm:
I looked primarily into our _Collections_ usage and found some room for improvement. 

While stress testing the system to get results I also stumbled upon a memory leak in the system to UI eventBus bridge and a performance problem with spring data JPA (https://jira.spring.io/browse/DATAJPA-841) I fixed the first and worked around the second to be able finally run my tests. Those two fixed actually had a bigger impact than my original cause 😄 

Test setup:
- Start example app with default settings with MySQL/RabbitMQ/MongoDB
- Open all UI views in one session (i.e. one browser tab each).
- Run Mgmt API simulator with 10k targets and 300 DS
- Run device Simulator with DMF and 100k targets
- Collect heapdump

Results (master/this branch):
- Heap in MB: 95.3/92.6
- ArrayLists in heap: 52.032/46.583
- HashMaps in heap: 52.696/43.628


hawkbit-mgmt-resource/src/test/java/org/eclipse/hawkbit/mgmt/rest/resource/MgmtDistributionSetTypeResourceTest.java
```diff
@@ -554,7 +555,7 @@ public void invalidRequestsOnDistributionSetTypesResource() throws Exception {
                 .contentType(MediaType.APPLICATION_JSON)).andDo(MockMvcResultPrinter.print())
                 .andExpect(status().isBadRequest());
 
-        final DistributionSetType missingName = entityFactory.generateDistributionSetType("test123", null, "Desc123");
+        final DistributionSetType missingName = new JpaDistributionSetType("test123", null, "Desc123");
```
michahirsch: why using here `new JpaDistributionSetType` and not like before the `entityFactory.generateDistributionSetType`

hawkbit-mgmt-resource/src/test/java/org/eclipse/hawkbit/mgmt/rest/resource/MgmtSoftwareModuleResourceTest.java
```diff
@@ -457,7 +458,7 @@ public void invalidRequestsOnSoftwaremodulesResource() throws Exception {
                 .contentType(MediaType.APPLICATION_JSON)).andDo(MockMvcResultPrinter.print())
                 .andExpect(status().isBadRequest());
 
-        final SoftwareModule missingName = entityFactory.generateSoftwareModule(osType, null, "version 1", null, null);
+        final SoftwareModule missingName = new JpaSoftwareModule(osType, null, "version 1", null, null);
```
michahirsch: why using here `new JpaDistributionSetType` and not like before the `entityFactory.generateDistributionSetType`?

hawkbit-mgmt-resource/src/test/java/org/eclipse/hawkbit/mgmt/rest/resource/MgmtSoftwareModuleTypeResourceTest.java
```diff
@@ -337,7 +338,7 @@ public void invalidRequestsOnSoftwaremoduleTypesResource() throws Exception {
                 .contentType(MediaType.APPLICATION_JSON)).andDo(MockMvcResultPrinter.print())
                 .andExpect(status().isBadRequest());
 
-        final SoftwareModuleType missingName = entityFactory.generateSoftwareModuleType("test123", null, "Desc123", 5);
+        final SoftwareModuleType missingName = new JpaSoftwareModuleType("test123", null, "Desc123", 5);
```
michahirsch: why using here `new JpaDistributionSetType` and not like before the `entityFactory.generateDistributionSetType`?

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaTargetManagement.java
```diff
@@ -342,7 +343,7 @@ public TargetTagAssignmentResult toggleTagAssignment(final Collection<String> ta
 
         // all are already assigned -> unassign
         if (alreadyAssignedTargets.size() == allTargets.size()) {
-            alreadyAssignedTargets.forEach(target -> target.getTags().remove(tag));
+            alreadyAssignedTargets.forEach(target -> ((JpaTarget) target).removeTag(tag));
```
michahirsch: If you change the return type of the query `org.eclipse.hawkbit.repository.jpa.TargetRepository.findByTagNameAndControllerIdIn(String, Collection<String>)`  from `Target` to `JpaTarget` you don't need to cast here.

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/model/JpaDistributionSetType.java
```diff
@@ -108,37 +110,74 @@ public void setDeleted(final boolean deleted) {
...
+        if (!(dsType instanceof JpaDistributionSetType) || isOneModuleListEmpty(dsType)) {
+            return false;
+        } else if (areBothModuleListsEmpty(dsType)) {
+            return true;
+        }
+
         return new HashSet<DistributionSetTypeElement>(((JpaDistributionSetType) dsType).elements).equals(elements);
     }
 
+    private boolean isOneModuleListEmpty(final DistributionSetType dsType) {
```
michahirsch: little bit confused about the method name, is not about if one list is empty, is about if one is empty and the other not, or if the other is empty and the one not :) Just the name and the complex if was little bit hard to read.

Before there was tested if the `elements` were equal is this the same check by checking if there are empty?
`HashSet<DistributionSetTypeElement>(((JpaDistributionSetType) dsType).elements).equals(elements);`

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/management/targettable/TargetTable.java
```diff
@@ -140,11 +135,14 @@ public void onEvents(final List<?> events) {
         if (TargetCreatedEvent.class.isInstance(firstEvent)) {
             onTargetCreatedEvents();
         } else if (TargetInfoUpdateEvent.class.isInstance(firstEvent)) {
-            onTargetInfoUpdateEvents((List<TargetInfoUpdateEvent>) events);
+            onTargetUpdateEvents(((List<TargetInfoUpdateEvent>) events).stream()
```
michahirsch: This pull-request is about memory consumption and lists, but here you are copying the event list into a new list, I guess I know why. Probably because of using the same method `onTargetUpdateEvents` and the same logic.  I guess here it's OK to map the lists to another list of `List<Target>` but it would also work if you have a method 3 methods like :
```
void onEvent(List<TargetInfoUpdateEvent> events)
{
   events.foreach(event->onEvent(event.getEntity())):
}

void onEvent(List<TargetUpdateEvent> events)
{
   events.foreach(event->onEvent(event.getEntity())):
}

void onEvent(Target t)
{
  // do update table
}
```

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/push/DelayedEventBusPushStrategy.java
```diff
@@ -163,38 +173,43 @@ private DispatchRunnable(final UI ui, final VaadinSession session) {
...
+            }
+
+            final List<org.eclipse.hawkbit.eventbus.event.Event> events = new ArrayList<>(size);
+            for (int i = 0; i < size; i++) {
                 final org.eclipse.hawkbit.eventbus.event.Event pollEvent = queue.poll();
                 if (pollEvent == null) {
                     continue;
                 }
-                events.add(pollEvent);
+                events.add(i, pollEvent);
```
michahirsch: There is no difference in using `events.add(pollEvent);` or using `events.add(i, pollEvent);` in this case, but when I'm thinking about it we actually changing the order of the events and maybe you were trying to fix this but then you would need to change for for-loop. The `queue.poll()` is given you a oldest event and this we add currently to the beginning of the list.

In my opinion the for-loop should like this
```
for(int i = size ; i >=0 ; i--)
{
   final org.eclipse.hawkbit.eventbus.event.Event pollEvent = queue.poll();
   if (pollEvent == null) {
      continue;
   }
   events.add(i, pollEvent);
}
```

schabdo: If the order is correct better go for `queue.drainTo(...)` due to:
> This operation may be more efficient than repeatedly polling this queue

If you'll change the order as mentioned by @michahirsch take into account to use a `Deque` instead of a handcrafted reverse-loop approach.

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/push/DelayedEventBusPushStrategy.java
```diff
@@ -57,26 +57,32 @@
     private static final Logger LOG = LoggerFactory.getLogger(DelayedEventBusPushStrategy.class);
 
     private static final int BLOCK_SIZE = 10_000;
-    private final ScheduledExecutorService executorService = Executors.newSingleThreadScheduledExecutor();
+    private final ScheduledExecutorService executorService;
     private final BlockingDeque<org.eclipse.hawkbit.eventbus.event.Event> queue = new LinkedBlockingDeque<>(BLOCK_SIZE);
     private final EventBus.SessionEventBus eventBus;
     private final com.google.common.eventbus.EventBus systemEventBus;
+    private int uiid = -1;
```
schabdo: The use of `uiid` is not thread-safe. I suppose this was intended so `DispatchRunnable` is finished with the id it was  started with. Maybe you can use `vaadinUI.getUIId()` as it is already a instance variable of `DispatchRunnable` ?

hawkbit-mgmt-resource/src/test/java/org/eclipse/hawkbit/mgmt/rest/resource/MgmtDistributionSetTypeResourceTest.java
kaizimmerm: The entity factory ensures that no illegal entities are created. However, in this case the whole idea is to create an illegal one.

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/model/JpaDistributionSetType.java
kaizimmerm: Because they can also be null

hawkbit-mgmt-resource/src/test/java/org/eclipse/hawkbit/mgmt/rest/resource/MgmtSoftwareModuleResourceTest.java
kaizimmerm: The entity factory ensures that no illegal entities are created. However, in this case the whole idea is to create an illegal one.

hawkbit-mgmt-resource/src/test/java/org/eclipse/hawkbit/mgmt/rest/resource/MgmtSoftwareModuleTypeResourceTest.java
kaizimmerm: The entity factory ensures that no illegal entities are created. However, in this case the whole idea is to create an illegal one.

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/push/DelayedEventBusPushStrategy.java
```diff
@@ -163,38 +173,43 @@ private DispatchRunnable(final UI ui, final VaadinSession session) {
...
-            final List<org.eclipse.hawkbit.eventbus.event.Event> events = new LinkedList<>();
-            for (int i = 0; i < BLOCK_SIZE; i++) {
+
+            final int size = queue.size();
+            if (size <= 0) {
+                LOG.debug("UI EventBus aggregator for UI {} has nothing to do.", uiid);
+                return;
+            }
+
+            final List<org.eclipse.hawkbit.eventbus.event.Event> events = new ArrayList<>(size);
```
schabdo: remove full qualified name, it's not required

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/push/DelayedEventBusPushStrategy.java
```diff
@@ -138,7 +148,7 @@ public void clean() {
      * @return {@code true} if the event can be dispatched to the UI otherwise
      *         {@code false}
      */
-    protected boolean eventSecurityCheck(final SecurityContext userContext,
+    protected static boolean eventSecurityCheck(final SecurityContext userContext,
```
schabdo: remove full qualified name of `Event ` it's not required

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaTargetManagement.java
kaizimmerm: @michahirsch makes no sense, create a cast, generates a another one
@schabdo nope, that means a lot more casts for the _Management_ services

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/model/JpaTarget.java
```diff
@@ -161,9 +158,37 @@ public String getControllerId() {
...
+
+    public boolean addTag(final TargetTag tag) {
+        if (tags == null) {
+            tags = new HashSet<>();
+        }
+
+        return tags.add(tag);
+    }
+
+    public boolean removeTag(final TargetTag tag) {
```
kaizimmerm: Why???

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/model/JpaTarget.java
```diff
@@ -161,9 +158,37 @@ public String getControllerId() {
...
 
+    public List<RolloutTargetGroup> getRolloutTargetGroup() {
+        if (rolloutTargetGroup == null) {
+            return Collections.emptyList();
+        }
+
+        return Collections.unmodifiableList(rolloutTargetGroup);
+    }
+
+    public boolean addTag(final TargetTag tag) {
```
kaizimmerm: Why???

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/management/targettable/TargetTable.java
kaizimmerm: That won't work as we need to work on the entire list on on every target but I will check if I can optimise this a bit.

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/model/JpaDistributionSetType.java
```diff
@@ -108,37 +110,74 @@ public void setDeleted(final boolean deleted) {
...
+        return (!CollectionUtils.isEmpty(((JpaDistributionSetType) dsType).elements)
+                && CollectionUtils.isEmpty(elements))
```
kaizimmerm: Guava does not check for null here

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/push/DelayedEventBusPushStrategy.java
kaizimmerm: Fixed, thanks for the hint.

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaControllerManagement.java
```diff
@@ -163,12 +165,13 @@ public boolean hasTargetArtifactAssigned(final String controllerId, final LocalA
 
     @Override
     public Optional<Action> findOldestActiveActionByTarget(final Target target) {
-        return actionRepository.findFirstByTargetAndActiveOrderByIdAsc((JpaTarget) target, true);
+        return actionRepository.findFirstByTargetAndActive(new Sort(Direction.ASC, "id"), (JpaTarget) target, true);
```
kaizimmerm: Fixed, thanks for the hint.

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaRolloutGroupManagement.java
```diff
@@ -114,7 +114,7 @@ public RolloutGroup findRolloutGroupById(final Long rolloutGroupId) {
 
         for (final RolloutGroup rolloutGroup : rolloutGroups) {
             final TotalTargetCountStatus totalTargetCountStatus = new TotalTargetCountStatus(
-                    allStatesForRollout.get(rolloutGroup.getId()), rolloutGroup.getTotalTargets());
+                    allStatesForRollout.get(rolloutGroup.getId()), new Long(rolloutGroup.getTotalTargets()));
```
kaizimmerm: Fixed, thanks for the hint.

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaRolloutGroupManagement.java
```diff
@@ -128,7 +128,7 @@ public RolloutGroup findRolloutGroupWithDetailedStatus(final Long rolloutGroupId
                 .getStatusCountByRolloutGroupId(rolloutGroupId);
 
         final TotalTargetCountStatus totalTargetCountStatus = new TotalTargetCountStatus(rolloutStatusCountItems,
-                rolloutGroup.getTotalTargets());
+                new Long(rolloutGroup.getTotalTargets()));
```
kaizimmerm: Fixed, thanks for the hint.

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaSoftwareManagement.java
```diff
@@ -185,22 +189,22 @@ public SoftwareModule createSoftwareModule(final SoftwareModule swModule) {
...
             final Pageable pageable) {
-        return new PageImpl<>(new ArrayList<>(findAll.getContent()), pageable, findAll.getTotalElements());
+        return new PageImpl<>(Collections.unmodifiableList(findAll.getContent()), pageable, findAll.getTotalElements());
     }
 
     @Override
     public Long countSoftwareModulesByType(final SoftwareModuleType type) {
 
-        final List<Specification<JpaSoftwareModule>> specList = new ArrayList<>();
+        final List<Specification<JpaSoftwareModule>> specList = new LinkedList<>();
```
kaizimmerm: Fixed, thanks for the hint.

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/model/JpaActionStatus.java
```diff
@@ -136,13 +137,20 @@ public void setOccurredAt(final Long occurredAt) {
     @Override
     public final void addMessage(final String message) {
         if (message != null) {
+            if (messages == null) {
+                messages = new ArrayList<>((message.length() / 512) + 1);
```
kaizimmerm: Fixed, thanks for the hint.

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/push/DelayedEventBusPushStrategy.java
kaizimmerm: Fixed, thanks for the hint.

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/push/DelayedEventBusPushStrategy.java
kaizimmerm: Fixed, thanks for the hint.

hawkbit-mgmt-resource/src/test/java/org/eclipse/hawkbit/mgmt/rest/resource/MgmtDistributionSetTypeResourceTest.java
michahirsch: just thought that actually the mgmt-resource module shouldn't know about `JPA` so this was actually hided behind the `entityFactory`, now this test depends on a JPA implementation would it make sense to not use the Jpa-Entity instead a JSON string for the mgmt-resource tests

kaizimmerm: Yeah, I can use a static JSON string. Not easy to read but it does the trick.

hawkbit-dmf-amqp/src/main/java/org/eclipse/hawkbit/amqp/AmqpDeadletterProperties.java
```diff
@@ -9,24 +9,26 @@
...
 @ConfigurationProperties("hawkbit.dmf.rabbitmq.deadLetter")
 public class AmqpDeadletterProperties {
+    private static final int THREE_WEEKS = 21;
 
     /**
      * Message time to live (ttl) for the deadletter queue. Default ttl is 3
      * weeks.
      */
-    private long ttl = Duration.ofDays(21).toMillis();
+    private long ttl = Duration.ofDays(THREE_WEEKS).toMillis();
```
schabdo: :+1:

Issue #197, opened 2016-05-31
Repository abstraction API


kaizimmerm:
# Refactoring:
- [x] Provide interface on top of *management implementations.
- [x] Make entities immutable and create proper update methods that state by signature what can be updated.
- [X] Introduce builder pattern for EntityFactory create methods and repository update methods.
- [x] Use IDs in signatures for searches, not complete entities
- [x] Consistent usage of Java 8 Optional in the interfaces (Repo shall never return null!)
- [x] Remove TargetInfo interfaces as this does not belong into the API. The very existence of this object is very Spring Data/JPA specific.
- [x] Consistent usage of paging (make sure that we never query everything)
- [x] Fix DMF model name clashes by means of renaming DMF model in similar pattern to the other APIs.
- [x] Check if we really always ned Page or is Slice sufficient
- [x] Split management services (e.g. DS and DS type)
- [x] Consistent naming of methods (see below)

# Remove boiler plate
- [x] Remove external artifact (concept was never finished anyway and does not fit to hawkBit as of today)

# Method naming concept (example target management)

```
// Count all targets
Long count()

// Count by filter parameter (example)
Long countByTargetFilterQuery(@NotEmpty String targetFilterQuery);

//Create entity
List<Target> create(@NotEmpty Collection<TargetCreate> create)
Target create(@NotNull TargetCreate create)

//Delete entities (throws EntityNotFoundException if one element does not exist (at least one not found in collection case))
void delete(@NotEmpty Collection<Long> targetIDs);
void delete(@NotNull Long targetID);
void deleteByControllerId(@NotEmpty String controllerId);
void deleteByControllerId(@NotEmpty Collection<String> controllerId);

//Update Target (throws EntityNotFoundException if one element does not exist (at least one not found in collection case))
List<Target> update(@NotEmpty Collection<TargetUpdate> update);
Target update(@NotNull TargetUpdate update);

//Exist
boolean exists(@NotNull Long targetId)
boolean existsByAssignedDistributionSet(@NotNull Long distributionSetID);

// Read methods
// Find one on technical ID (Optional, no EntityNotFoundException)
Optional<Target> get(@NotNull Long targetId);
List<Target> get(@NotEmpty Collection<Long> targetId);

// Find one on non-ID but unique constraint (Optional, no EntityNotFoundException)
Optional<Target> getByControllerID(@NotEmpty String controllerId);
List<Target> getByControllerID(@NotEmpty Collection<String> controllerId);

[...]

PR #322, opened 2016-10-20
Removed external artifact from repository.

kaizimmerm:
First task for https://github.com/eclipse/hawkbit/issues/197

* Started to clean up Artifact and Software Management
* Removed external repository


hawkbit-mgmt-resource/src/main/java/org/eclipse/hawkbit/mgmt/rest/resource/MgmtSoftwareModuleMapper.java
```diff
@@ -156,28 +155,20 @@ public static MgmtSoftwareModule toResponse(final SoftwareModule baseSofwareModu
      * @return
      */
     static MgmtArtifact toResponse(final Artifact artifact) {
-        final MgmtArtifact.ArtifactType type = artifact instanceof LocalArtifact ? MgmtArtifact.ArtifactType.LOCAL
-                : MgmtArtifact.ArtifactType.EXTERNAL;
-
         final MgmtArtifact artifactRest = new MgmtArtifact();
-        artifactRest.setType(type);
```
schabdo: Just to be sure: This will change the API, is it intended?

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/model/JpaArtifact.java
```diff
@@ -53,10 +53,19 @@
     @JoinColumn(name = "software_module", nullable = false, updatable = false, foreignKey = @ForeignKey(value = ConstraintMode.CONSTRAINT, name = "fk_assigned_sm"))
     private JpaSoftwareModule softwareModule;
 
+    @Column(name = "sha1_hash", length = 40, nullable = true)
```
schabdo: What about `javax.validation` annotations?

kaizimmerm: We never do them execept if we want to do a validation that is not supported by the database

hawkbit-mgmt-resource/src/main/java/org/eclipse/hawkbit/mgmt/rest/resource/MgmtSoftwareModuleMapper.java
kaizimmerm: Well, it was an unused part of the API. So I think its save to remove it. We supported only one type (local) anyway.

PR #326, opened 2016-10-21
fix NPE on color picking on dist and software types

michahirsch:


hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/distributions/disttype/CreateUpdateDistSetTypeLayout.java
```diff
@@ -121,6 +122,31 @@ protected void buildLayout() {
         mainLayout.addComponent(twinTableLayout, 2, 0);
     }
 
+    @Override
+    protected void previewButtonClicked() {
```
schabdo: This method is almost identically to `CreateUpdateSoftwareTypeLayout#previewButtonClicked()` (the one above). They only differ in the checked entity (line 132/133). So use the common super class to share the logic and only implement the query in each child class differen. You can find an detailed example here => http://refactoring.com/catalog/formTemplateMethod.html

PR #339, opened 2016-11-07
Upgrade vaadin from 7.6.8 -> 7.7.3:

MarcelMager:
upgrade vaadin.spring       from 1.0.0 -> 1.0.2
upgrade flexibleoptiongroup from 2.2.0 -> 2.3.0
upgrade dbar-addon          from 1.2.0 -> 2.0.0
exclude gwt-elemental for example app

Signed-off-by: Marcel Mager (INST-IOT/ESB) <Marcel.Mager@bosch-si.com>

### DIFF
--- a/hawkbit-rest-core/pom.xml
+++ b/hawkbit-rest-core/pom.xml
@@ -48,7 +48,6 @@
       <dependency>
          <groupId>javax.servlet</groupId>
          <artifactId>javax.servlet-api</artifactId>
-         <scope>provided</scope>
       </dependency>
       <dependency>
          <groupId>org.apache.tomcat.embed</groupId>
--- a/hawkbit-rest-core/pom.xml
+++ b/hawkbit-rest-core/pom.xml
@@ -48,6 +48,7 @@
       <dependency>
          <groupId>javax.servlet</groupId>
          <artifactId>javax.servlet-api</artifactId>
+         <scope>provided</scope>
       </dependency>
       <dependency>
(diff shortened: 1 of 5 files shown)


hawkbit-rest-core/pom.xml
```diff
@@ -48,7 +48,6 @@
       <dependency>
          <groupId>javax.servlet</groupId>
          <artifactId>javax.servlet-api</artifactId>
-         <scope>provided</scope>
```
michahirsch: why we removed the `provided` scope of the servlet-api? Can you explain the case?

MarcelMager: I removed it for symmetry reasons, as the hawkbit-security-core also uses compile scope for the javax.servlet-api.
Using provided scopes brings the risk of runtime problems regarding wrong class version (e.g. gwt-elemental brings transitive dependency on gwt-user that brings along classes of javax.servlet-api with wrong version; this is no current problem, but dependencies might change...).
On the other hand: the usage of compile scope might cause clashes in classpath.
I don't know what is better, but I think hawkbit-security-core and hawkbit-rest-core should handle it symmetrically.

michahirsch: I'd prefer a general use of `provided` scope of the servlet-api. Using `compile` scope will lead to problems when using different application-server. You can also build an hawkBit application as WAR layout and deployed in a application server your choice. This will lead to clashes as you mentioned. So I'd prefer `provided` also for other modules as well :smile:

PR #341, opened 2016-11-09
Feature assert events within tests

Jonkno:
Introduction of event count and verification mechanism

- new annotation to annotate tests with event type and expected count
- added annotations to TargetManagmentTest and DdiRootControllerTest


hawkbit-repository/hawkbit-repository-test/src/main/java/org/eclipse/hawkbit/repository/test/matcher/EventVerifier.java
```diff
@@ -0,0 +1,93 @@
...
+
+    private void removeEventListener() {
+        final ApplicationEventMulticaster multicaster = TestContextProvider.getContext()
+                .getBean(ApplicationEventMulticaster.class);
+        multicaster.removeApplicationListener(eventCaptor);
+    }
+
+    private static class EventCaptor implements ApplicationListener<RemoteApplicationEvent> {
+
+        public final Multiset<Class<?>> capturedEvents = HashMultiset.create();
```
michahirsch: This field can probably be `private`

hawkbit-repository/hawkbit-repository-test/src/main/java/org/eclipse/hawkbit/repository/test/matcher/EventVerifier.java
```diff
@@ -0,0 +1,132 @@
...
+                        equalTo(expectedEvent.count()));
+
+            } catch (final ConditionTimeoutException ex) {
+                Assert.fail("Did not receive the expected amount of events form " + expectedEvent.type() + " Expected: "
+                        + expectedEvent.count() + " but was: " + eventCaptor.getCountFor(expectedEvent.type()));
+            }
+        }
+    }
+
+    private void verifyAllEventsCounted(final Expect[] expectedEvents) {
```
michahirsch: The chance that asynchronous events are still processed in the eventbus are very unlikely, nevertheless it's possible that after all `expected` events are correct  `verifyRightCountOfEvents` still total events are missing. So I guess we'll need a `waiter` in the `verifyAllEventsCounted` as well.

hawkbit-repository/hawkbit-repository-jpa/src/test/java/org/eclipse/hawkbit/repository/jpa/TargetManagementTest.java
```diff
@@ -336,7 +364,7 @@ public void singleTargetIsInsertedIntoRepo() throws Exception {
         assertThat(target).as("Target compared with saved target").isEqualTo(savedTarget);
 
         savedTarget.setDescription("changed description");
-        Thread.sleep(1);
```
michahirsch: Oh, the sleep probably made sense because of this assertion
```
assertThat(createdAt).as("CreatedAt compared with saved modifiedAt")
                .isNotEqualTo(savedTarget.getLastModifiedAt());
        assertThat(modifiedAt).as("ModifiedAt compared with saved modifiedAt")
                .isNotEqualTo(savedTarget.getLastModifiedAt());
```
It's a check about `notEqual` and maybe on a fast machine it could be equal?

kaizimmerm:
There are still commits comming in. Is this pull request complete now and can be reviewed?

Jonkno:
Yes, you can review.

PR #344, opened 2016-11-09
Fix mayors squid:S1604 and squid:S1151

kaizimmerm:
With that our sonar quality gate should be happy again.


hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/filtermanagement/client/AutoCompleteTextFieldConnector.java
```diff
@@ -48,25 +47,20 @@
     protected void init() {
         super.init();
 
-        registerRpc(TextFieldSuggestionBoxClientRpc.class, new TextFieldSuggestionBoxClientRpc() {
```
michahirsch: You cannot use lambdas in GWT 2.7. This is client side code and will be compiled with GWT,  so this must keep anonymous inner class. see GWT 2.8 RC1 release notes: http://www.gwtproject.org/release-notes.html#Release_Notes_2_8_0_RC1

PR #348, opened 2016-11-15
fixes #347 correct extension parent

### LINKED ISSUE #347: Wrong Maven parent for Maven "extensions" modules
As the extensions modules specify hawkbit-parent as their parent POM, this fail when building hawkbit on a system that does not have it yet.

michahirsch:
correct the hawkbit-extension-uaa parent.


michahirsch:
better fix by @kartben https://github.com/eclipse/hawkbit/pull/349

PR #350, opened 2016-11-15
use ID property for sorting because its indexed on DB

michahirsch:
use the sort property `id` instead of `createdAt` because the `id` field is indexed.


hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/filtermanagement/CustomTargetBeanQuery.java
```diff
@@ -45,7 +45,7 @@
 public class CustomTargetBeanQuery extends AbstractBeanQuery<ProxyTarget> {
 
     private static final long serialVersionUID = 6490445732785388071L;
-    private Sort sort = new Sort(Direction.DESC, "createdAt");
+    private Sort sort = new Sort(Direction.ASC, "id");
```
michahirsch: I changed the order of the FilterManagementTable from latest to bottom. Before it was latest top which was the only table which behaved like this. So for consistent I changed it as well.

PR #367, opened 2016-11-29
Fix UI login and view change performance

kaizimmerm:
The UI is getting slower and slower by means of both login as well as open view performance in general.

I optimised in the following directions:
* Switched view to Ui scope to ensure that they do not get garbage collected when the user switches the view.
* Reduced the number of prototype beans drastically to reduce pressure on the bean factory. New strategy is to have primarily only the views themselves and everything in sessions cope as bean.
* A new minor tweaks on repository access to reduce DB calls.
* UI internal events switched to Ui scope to ensure that multiple UIs in the same session are not flooded by duplicate events by DelayedEventBusPushStrategy (which is running per UI and not session).
* While at it I migrated most of the remaining beans to constructor injection.

### DIFF
--- a/hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/common/grid/AbstractGrid.java
+++ b/hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/common/grid/AbstractGrid.java
@@ -8,13 +8,11 @@
  */
 package org.eclipse.hawkbit.ui.common.grid;
 
-import javax.annotation.PostConstruct;
-import javax.annotation.PreDestroy;
-
+import org.eclipse.hawkbit.repository.SpPermissionChecker;
 import org.eclipse.hawkbit.ui.utils.I18N;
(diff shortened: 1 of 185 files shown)


hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/common/grid/AbstractGrid.java
```diff
@@ -27,19 +25,18 @@
...
 
-    @Autowired
-    protected transient EventBus.SessionEventBus eventBus;
+    protected final I18N i18n;
+
+    protected final transient EventBus.UIEventBus eventBus;
 
+    protected final SpPermissionChecker permissionChecker;
+
+    protected AbstractGrid(final I18N i18n, final UIEventBus eventBus, final SpPermissionChecker permissionChecker) {
```
kaizimmerm: I think these kind of things are hard to standardise. I would guess that none of our constructors follow any pattern here.

PR #368, opened 2016-11-29
Replace server-side dnd validation with client-side validation:

MarcelMager:
Client-side validation means:
* Custom widgetset is used to introduce the client-side validation
* Accept-calculation is performed on client-side
* Drop-hints are calculated on client-side (show/hide)
* Row count decoration is calculated on client-side (multi-selection)
* Notification box is triggered from client-side

Beside the migration from server-side validation to client-side
validation there are few improvements:
* Improvement1: If drag is aborted outside the components by releasing
the mouse or the drag aborts by pressing the ESC key) -> drop area hints
are removed
* Improvement2: If multiple rows are selected in table, but drag is
performed on a non-selected single element from table -> the single
drag-element is not decorated with the row selection count
* Improvement3: Upload View: Component 'Drop files to upload' only
accepts files (and no drags from filter and software module component)

Signed-off-by: Marcel Mager (INST-IOT/ESB) <Marcel.Mager@bosch-si.com>


kaizimmerm:
Cool, I definitly have to try this out 😄

michahirsch:
Tried it out in firefox and chrome browser and definitely much smoother drag'n-drop feeling.  Also the showing drop hints as well as the drop-highlighting is much smoother and faster, thanks for that :+1:

PR #369, opened 2016-11-30
Create Rollout UI with Groups definition

dominikhb:
Extended the UI Dialog which creates a new Rollout. 
In this dialog a new Rollout can be defined either with _Number of Groups_ or with an 
_Advanced Group definition_.

In the _Advanced Group definition_ tab groups can be added and defined with a _Target filter query_ and 
_Target percentage_. The _Number of Groups_ tab defines a fixed amount of groups onto which the
targets are distributed in equal parts.

A pie chart has been added that shows the distribution of the total targets count onto the different
groups.

Reviewers: @kaizimmerm @michahirsch


hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/common/builder/ComboBoxBuilder.java
```diff
@@ -0,0 +1,53 @@
...
+
+import org.eclipse.hawkbit.ui.components.SPUIComponentProvider;
+import org.eclipse.hawkbit.ui.utils.I18N;
+import org.eclipse.hawkbit.ui.utils.SPUILabelDefinitions;
+
+import com.vaadin.data.Property;
+import com.vaadin.ui.ComboBox;
+import com.vaadin.ui.themes.ValoTheme;
+
+public class ComboBoxBuilder {
```
melanieretter: Please add JavaDoc to Class. Quick explain what the class is doing.

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/rollout/groupschart/client/GroupsPieChartWidget.java
```diff
@@ -0,0 +1,205 @@
...
+        moveSelection(infoText, x, y);
+        infoText.attr(ATTR_VISIBILITY, "visible");
+    }
+
+    private static void moveSelection(Selection sel, double x, double y) {
+        sel.attr(ATTR_TRANSFORM, "translate(" + x + ", " + y + ")");
+    }
+
+
+    private static final native double getTextWidth(Element e)/*-{
```
Jonkno: I guess you didn't want to add this comment block?

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/rollout/groupschart/client/GroupsPieChartClientRpc.java
```diff
@@ -0,0 +1,15 @@
...
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package org.eclipse.hawkbit.ui.rollout.groupschart.client;
+
+import com.vaadin.shared.communication.ClientRpc;
+
+public interface GroupsPieChartClientRpc extends ClientRpc {
```
Jonkno: Document this public interface

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/rollout/groupschart/client/GroupsPieChartServerRpc.java
```diff
@@ -0,0 +1,15 @@
...
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package org.eclipse.hawkbit.ui.rollout.groupschart.client;
+
+import com.vaadin.shared.communication.ServerRpc;
+
+public interface GroupsPieChartServerRpc extends ServerRpc {
```
Jonkno: Document this public interface

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/rollout/groupschart/client/GroupsPieChartWidget.java
```diff
@@ -0,0 +1,205 @@
...
+        private double startAngle;
+
+        private double endAngle;
+
+        public PieArc(double startAngle, double endAngle) {
+            this.startAngle = startAngle;
+            this.endAngle = endAngle;
+        }
+
+        public double getStartAngle() {
```
Jonkno: You can remove this getter since it is not needed.

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/rollout/groupschart/client/GroupsPieChartWidget.java
dominikhb: The comment block is required, because it is the implementation of the native method. See [JSNI](http://www.gwtproject.org/doc/latest/DevGuideCodingBasicsJSNI.html).

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/rollout/groupschart/client/GroupsPieChartWidget.java
```diff
@@ -0,0 +1,205 @@
...
+        sel.attr(ATTR_TRANSFORM, "translate(" + x + ", " + y + ")");
+    }
+
+
+    private static final native double getTextWidth(Element e)/*-{
+        return e.getBBox().width;
+    }-*/;
+
+    private static final native double getTextHeight(Element e)/*-{
+        return e.getBBox().height;
```
dominikhb: The comment block is required, because it is the implementation of the native method. See [JSNI](http://www.gwtproject.org/doc/latest/DevGuideCodingBasicsJSNI.html).

hawkbit-repository/hawkbit-repository-api/src/main/java/org/eclipse/hawkbit/repository/builder/RolloutUpdate.java
```diff
@@ -30,4 +35,35 @@
...
+    default RolloutUpdate set(final DistributionSet set) {
+        return set(Optional.ofNullable(set).map(DistributionSet::getId).orElse(null));
+    }
+
+    /**
+     * @param setId
+     *            for {@link Rollout#getDistributionSet()}
+     * @return updated builder instance
+     */
+    RolloutUpdate set(long setId);
```
michahirsch: The ID cannot be updated actually, right? So it shouldn't be included in the `RolloutUpdate` class

hawkbit-ui/pom.xml
```diff
@@ -252,6 +252,18 @@
          <artifactId>spring-boot-configuration-processor</artifactId>
          <optional>true</optional>
       </dependency>
+      <dependency>
+         <groupId>com.github.gwtd3</groupId>
+         <artifactId>gwt-d3-api</artifactId>
+         <version>1.2.0</version>
```
michahirsch: would be good to move this dependency into the parent-pom to the `<dependencyManagement>`  where all dependencies of hawkBit are declared

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/rollout/groupschart/GroupsPieChart.gwt.xml
```diff
@@ -0,0 +1,29 @@
...
+    Copyright (c) 2015 Bosch Software Innovations GmbH and others.
+
+    All rights reserved. This program and the accompanying materials
+    are made available under the terms of the Eclipse Public License v1.0
+    which accompanies this distribution, and is available at
+    http://www.eclipse.org/legal/epl-v10.html
+
+-->
+<module>
+	<inherits name="com.vaadin.DefaultWidgetSet" />
```
michahirsch: small one but the XML seems not to be formatted, right. There are some additional spaces in the beginning it seems.

dominikhb:
I optimized the queries that are being built to validate the groups definition.

The Create Rollouts UI runs the validation now as async task. One UI Session can only run one async validation after another. They don't run in parallel to avoid overloading the system with tasks. When a new validation is requested during a running validation task, it sets a marker to run one more validation after the current one. Once nothing has changed anymore and the validation finishes, the result is being displayed.

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/RolloutHelper.java
```diff
@@ -215,30 +239,69 @@ static String getAllGroupsTargetFilter(final List<RolloutGroup> groups) {
         if (groups.stream().anyMatch(group -> StringUtils.isEmpty(group.getTargetFilterQuery()))) {
             return "";
         }
-        return groups.stream().map(RolloutGroup::getTargetFilterQuery).collect(Collectors.joining(","));
+
+        return "(" + groups.stream().map(RolloutGroup::getTargetFilterQuery).distinct().sorted()
```
dominikhb: It probably does not matter in this case how it is used, because the groups list is always in the same order for multiple calls of the `getAllGroupsTargetFilter` method.
The result is later used for caching and thus in the key of a map, where I want't to uniquely identify the filter. I wanted to avoid two filters in the Map, that mean the same but are ordered differently. Maybe there is a corner case where the `.sorted()` call helps.

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/rsql/RSQLUtility.java
```diff
@@ -297,6 +338,20 @@ private RSQLParameterUnsupportedFieldException createRSQLParameterUnsupportedExc
 
                 final String fieldNameSplit = split[i];
                 fieldPath = (fieldPath != null) ? fieldPath.get(fieldNameSplit) : root.get(fieldNameSplit);
+                if (fieldPath instanceof PluralJoin) {
```
dominikhb: I added a JavaDoc description to the method.

PR #373, opened 2016-12-01
Introduce optional throttled DMF requeuing

kaizimmerm:
We have currently the problem that when we have to requeue a message we do this immediately. The result is that rabbit will push it back immediately again. This results in loop that consumes all available resources both on broker and server side.

The standard Rabbit recommendation to throttle this using dead letter in combination with a ttl is problematic as it does not allow to still have a real dead letter for later investigation.

In general we have to distinguish between three error cases:
* hawkBit runtime has a real problem (e.g. DB down). These message should be requed and processed later, i.e. when the problem is fixed. This is the default behaviour.
* hawkBit identifies the cause of the failure as a problem on the sender side (i.e. invalid data). This message goes into dead letter. These problems are defined by means of blacklisting the issue. This is done to ensure that we do not "loose" messages into dead letter in case of an unknown error. I call this "loose" as in practice it is really hard to find message in the haystack called dead letter.
* As a result the third category are Unknown errors which might be of either of the categories above.

No matter which cause is the reason to requeue it seems to me to be good compromise to continue the current behaviour but throttle message consumption down until the problem could be fixed on hawkBit side, i.e. either handling the message by means pushing to dead letter or fix the problem inside hawkBit that cause the issue.

I used a new custom exception handling strategy introduced with Spring AMQP 1.6.3 (and fixed in 1.6.5 😄 ) to get this done.

A new `hawkbit.dmf.rabbitmq.requeue-delay` can be configured (default: 0)

### DIFF
--- a/hawkbit-dmf-amqp/src/main/java/org/eclipse/hawkbit/amqp/AmqpConfiguration.java
+++ b/hawkbit-dmf-amqp/src/main/java/org/eclipse/hawkbit/amqp/AmqpConfiguration.java
@@ -51,7 +51,6 @@ import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.retry.backoff.ExponentialBackOffPolicy;
 import org.springframework.retry.support.RetryTemplate;
-import org.springframework.util.ErrorHandler;
 
 import com.google.common.collect.Maps;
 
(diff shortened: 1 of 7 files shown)


michahirsch:
Is it possible to make the `ErrorHandler` as a bean to be able to implement also a custom behavior.

kaizimmerm:
Done, its a bean now 😄

PR #382, opened 2016-12-07
Added mising update of DS property.

kaizimmerm:
There was one field that could be updated by UI but not API. I fixed that.

I also updated the sonar plugin due to some issues I had here, besides its sonar recommendation to fix the version.


michahirsch:
The `requiredMigrationStep` propery has been removed from the `MgmtDistributionSetRequestBodyPost` so I cannot set the field anymore when creating a new `DistributionSet`? Or do I see it wrong?

kaizimmerm:
Inheritance: Post is a child of Put, so its still there

PR #383, opened 2016-12-08
encode class type information in message payload instead of header

michahirsch:
To support rabbitMQ producer batching where messages are batched together and only the header of the first message is transfered and so different message headers are not supported, we should remove the special `__CLASS__` header from the messages.

The class information of the event-type is not encoded in the first 2 bytes of the payload. So it's included in the message.


hawkbit-repository/hawkbit-repository-core/src/test/java/org/eclipse/hawkbit/event/BusProtoStuffMessageConverterTest.java
```diff
@@ -0,0 +1,88 @@
...
+    @Mock
+    private Message<Object> messageMock;
+
+    @Before
+    public void before() {
+        when(targetMock.getId()).thenReturn(1L);
+    }
+
+    @Test
+    @Description("Verifies that the TargetCreatedEvent can be successfully seriaized and deserialized")
```
melanieretter: seria**l**ized

hawkbit-repository/hawkbit-repository-core/src/main/java/org/eclipse/hawkbit/event/BusProtoStuffMessageConverter.java
```diff
@@ -52,57 +59,96 @@ protected boolean supports(final Class<?> aClass) {
...
+            throw new MessageConversionException("Missing mapping of EventType for value " + eventType.getValue());
+        }
+        @SuppressWarnings("unchecked")
+        final Schema<Object> schema = (Schema<Object>) RuntimeSchema.getSchema(targetClass);
+        final Object deserializeEvent = schema.newMessage();
+        ProtobufIOUtil.mergeFrom(content, deserializeEvent, schema);
+        return deserializeEvent;
+    }
+
+    private static void mergeClassHeaderAndContent(final byte[] clazzHeader, final byte[] writeContent,
```
schabdo: Better create `body` within this method and return it instead of passing it around.

hawkbit-repository/hawkbit-repository-core/src/main/java/org/eclipse/hawkbit/event/BusProtoStuffMessageConverter.java
```diff
@@ -52,57 +59,96 @@ protected boolean supports(final Class<?> aClass) {
...
+            final byte[] body) {
+        System.arraycopy(clazzHeader, 0, body, 0, clazzHeader.length);
+        System.arraycopy(writeContent, 0, body, clazzHeader.length, writeContent.length);
+    }
+
+    private static void splitClazzHeaderAndContent(final byte[] payload, final byte[] clazzHeader,
```
schabdo: Same here with `content`. Rename the method to e.g. `extractContent(...)` and return the content instead of passing a empty array into this method.

hawkbit-repository/hawkbit-repository-core/src/main/java/org/eclipse/hawkbit/event/BusProtoStuffMessageConverter.java
```diff
@@ -52,57 +59,98 @@ protected boolean supports(final Class<?> aClass) {
...
-        throw new MessageConversionException(
-                "Cannot set the serialize class to message header. Need Mutable message header");
-    }
-
-    private static boolean isAccessorMutable(final MessageHeaders headers) {
-        final MessageHeaderAccessor accessor = MessageHeaderAccessor.getAccessor(headers, MessageHeaderAccessor.class);
-        return accessor != null && accessor.isMutable();
-    }
-
 }
```
SirWayne: Line ending :)

PR #394, opened 2016-12-15
Support secure AMQP connections

schabdo:
Connections to RabbitMQ can now be encrypted by setting `spring.rabbitmq.ssl.enabled=true` as property. 

Depending on your broker configuration you'll have to set the corresponding port for SSL via  `spring.rabbitmq.port` as well. At the moment the default implementation is trusting everybody out there with a valid certificate.

Reviewer: @kaizimmerm & @Jonkno

### DIFF
--- a/hawkbit-dmf-amqp/src/main/java/org/eclipse/hawkbit/amqp/AmqpConfiguration.java
+++ b/hawkbit-dmf-amqp/src/main/java/org/eclipse/hawkbit/amqp/AmqpConfiguration.java
@@ -8,6 +8,7 @@
  */
 package org.eclipse.hawkbit.amqp;
 
+import java.security.GeneralSecurityException;
 import java.time.Duration;
 import java.util.Map;
 import java.util.concurrent.Executor;
@@ -81,7 +82,7 @@ public class AmqpConfiguration {
 
     /**
      * Register the bean for the custom error handler.
-     * 
+     *
      * @return custom error handler
      */
     @Bean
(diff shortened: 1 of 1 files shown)


hawkbit-dmf-amqp/src/main/java/org/eclipse/hawkbit/amqp/AmqpConfiguration.java
```diff
@@ -109,19 +110,26 @@ public ErrorHandler errorHandler() {
...
          * @param config
          *            with standard {@link RabbitProperties}
          * @return {@link ConnectionFactory}
+         * @throws GeneralSecurityException
+         *             in case of problems with enabled SSL connections
          */
         @Bean
-        public ConnectionFactory rabbitConnectionFactory(final RabbitProperties config) {
+        public ConnectionFactory rabbitConnectionFactory(final RabbitProperties config)
+                throws GeneralSecurityException {
```
Jonkno: Why are you throwing a GeneralSecurityException and not a KeyManagementException,NoSuchAlgorithmException?

hawkbit-dmf-amqp/src/main/java/org/eclipse/hawkbit/amqp/AmqpConfiguration.java
```diff
@@ -109,19 +110,26 @@ public ErrorHandler errorHandler() {
...
             final CachingConnectionFactory factory = new CachingConnectionFactory();
             factory.setRequestedHeartBeat(amqpProperties.getRequestedHeartBeat());
             factory.setExecutor(threadPoolExecutor);
             factory.getRabbitConnectionFactory().setHeartbeatExecutor(scheduledExecutorService);
             factory.setPublisherConfirms(true);
 
+            if (config.getSsl().isEnabled()) {
```
SirWayne: At `SecurityProperties` for http it's called `requireSsl`. I would use the same naming

michahirsch: these are the configuration from the `org.springframework.boot.autoconfigure.amqp.RabbitProperties` not from hawkBit.

SirWayne: ah ok :)

PR #395, opened 2016-12-19
Custom Tenant configuration.

kaizimmerm:
Tenant configuration was so far fixed in an enum. We changed this to a property based approach that allows hawkBit developers to extend the mechanism to their own need. To complete the picture we also added an extension point for the configuration UI.

Not directly related but on the branch is a second contribution: an event that is send when a device polls.


hawkbit-repository/hawkbit-repository-api/src/main/java/org/eclipse/hawkbit/repository/TargetManagement.java
```diff
@@ -644,4 +644,11 @@ Long countAllTargetsByTargetFilterQueryAndNotInRolloutGroups(List<RolloutGroup>
             + SpringEvalExpressions.IS_CONTROLLER)
     Target updateTarget(TargetUpdate update);
 
+    /**
+     * @param id
+     * @return
+     */
+    @PreAuthorize(SpringEvalExpressions.HAS_AUTH_READ_TARGET)
```
michahirsch: javadoc on this method is not complete yet :)

SirWayne: I think, we can remove this methode. It's not used.

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/components/SPUICheckBox.java
```diff
@@ -37,11 +42,14 @@
...
      *            inputpromt of the CheckBox
      */
-    private void decorate(final String caption, final String style, final String styleName, final boolean required,
-            final String data) {
+    private void decorate(final String id, final String caption, final String style, final String styleName,
+            final boolean required, final String data) {
         // Default settings
         setRequired(required);
         addStyleName(ValoTheme.CHECKBOX_SMALL);
+        if (null != id) {
```
SirWayne: Please switch the if paramaters.

hawkbit-core/src/main/java/org/eclipse/hawkbit/tenancy/configuration/TenantConfigurationProperties.java
```diff
@@ -0,0 +1,169 @@
...
+         */
+        public static final String POLLING_OVERDUE_TIME_INTERVAL = "pollingOverdueTime";
+
+        /**
+         * See system default {@link Download#isEnabled()}.
+         */
+        public static final String ANONYMOUS_DOWNLOAD_MODE_ENABLED = "anonymous.download.enabled";
+
+        private String keyName;
+        private String defaultValue;
```
michahirsch: Maybe it makes sense to initialize the `defaultValue` as empty String. Because the default class and validator is String based. Then I could just configure the configuration for the `keyName` and it will work without problem with `null` value as default, right?

PR #397, opened 2016-12-20
Fix DMF update cancel message

kaizimmerm:
wo bugs actually:

- DMF tries to send cancelation even for DDI targets
- DMF send cancelation actually fails as target info is needed but not part of the event.


hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaDeploymentManagement.java
```diff
@@ -409,6 +409,7 @@ public Action cancelAction(final Action action, final Target target) {
      *            the action id of the assignment
      */
     private void cancelAssignDistributionSetEvent(final Target target, final Long actionId) {
+        target.getTargetInfo();
```
michahirsch: little comment on the lazy loading here would probably good.

PR #399, opened 2016-12-20
Posted feedback without "result" cause NPE in DDI API 

Jonkno:
Fix for the bug by using @Valid on member variables in nested object.
Also added junit test and done some refactoring

Review: @SirWayne and @melanieretter 

### DIFF
--- a/hawkbit-ddi-resource/src/test/java/org/eclipse/hawkbit/ddi/rest/resource/DdiDeploymentBaseTest.java
+++ b/hawkbit-ddi-resource/src/test/java/org/eclipse/hawkbit/ddi/rest/resource/DdiDeploymentBaseTest.java
@@ -939,14 +939,12 @@ public class DdiDeploymentBaseTest extends AbstractDDiApiIntegrationTest {
     }
 
     @Test
-    @Description("Various bad requests due to invalid feedback body. "
-            + "Ensures correct answering behaviour as expected to these kind of errors.")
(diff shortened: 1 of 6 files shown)


hawkbit-ddi-resource/src/test/java/org/eclipse/hawkbit/ddi/rest/resource/DdiDeploymentBaseTest.java
```diff
@@ -947,6 +938,41 @@ public void badDeplomentActionFeedback() throws Exception {
 
     }
 
+    @Test
```
michahirsch: I'd like to split these test up in two different ones because they actually testing different use-cases. One for missing the `result` in the feedback and one time the `finished` attribute. Then the test-case can have a better name e.g. `missingResultAttributeInFeedbackReturnsBadRequest`

Jonkno:
I checked the Management API but could not reproduce the same problem there because we are not using such nested objects like in the DDI API.

hawkbit-ddi-resource/src/test/java/org/eclipse/hawkbit/ddi/rest/resource/DdiDeploymentBaseTest.java
Jonkno: Fixed by splitting into three separated junit tests

PR #405, opened 2016-12-23
The unread event notification window will close when it loses focus

melanieretter:
The unread event notification window will close when it loses focus.

Reviewer: @SirWayne @schabdo 


hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/HawkbitUI.java
```diff
@@ -201,6 +204,12 @@ public void afterViewChange(final ViewChangeEvent event) {
         LOG.info("Current locale of the application is : {}", i18n.getLocale());
     }
 
+    private void closeAllWindows(final LayoutClickEvent event) {
+        for (final Window window : getWindows()) {
+            window.close();
```
SirWayne: As long i remember this will not work for every browser  e.g. Safari.
I think is better to use `UI#removeWindow(window);`

PR #416, opened 2017-01-13
Added template variables for request driven URL calculation.

kaizimmerm:
Download URL generation can now rely on host, domain and port from the DDI deploymentBase request itself.


hawkbit-core/src/main/java/org/eclipse/hawkbit/api/PropertyBasedArtifactUrlHandler.java
```diff
@@ -94,15 +107,21 @@ private static String generateUrl(final UrlProtocol protocol, final URLPlacehold
         return urlPattern;
     }
 
-    private static Map<String, String> getReplaceMap(final UrlProtocol protocol, final URLPlaceholder placeholder) {
+    private static Map<String, String> getReplaceMap(final UrlProtocol protocol, final URLPlaceholder placeholder,
+            final URI requestUri) {
         final Map<String, String> replaceMap = new HashMap<>();
```
Jonkno: I guess you cloud use  `Maps.newHashMapWithExpectedSize(12)` here

PR #417, opened 2017-01-13
Bug wrong count bulk upload dialog when runtime exception

Jonkno:
Fix for the bug by changing the way the syncCountAfterUpload method works. Now the method validates if something went wrong by comparing the successfullTargetCount/failedTargetCount with the total count. The process bar for the upload is also set to 100% if something went wrong.

Reviewer: @melanieretter and @schabdo 

### DIFF
--- a/hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/management/targettable/BulkUploadHandler.java
+++ b/hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/management/targettable/BulkUploadHandler.java
@@ -269,7 +269,7 @@ public class BulkUploadHandler extends CustomComponent
             tempFile = null;
         }
 
-        private void readEachLine(final String line, final double innerCounter, final double totalNumberOfLines) {
+        private void readEachLine(final String line, final long innerCounter, final long totalNumberOfLines) {
             final String csvDelimiter = ",";
             final String[] targets = line.split(csvDelimiter);
             if (targets.length == 2) {
@@ -281,7 +281,7 @@ public class BulkUploadHandler extends CustomComponent
             }
             final float current = managementUIState.getTargetTableFilters().getBulkUpload()
                     .getProgressBarCurrentValue();
-            final float next = (float) (innerCounter / totalNumberOfLines);
+            final float next = innerCounter / totalNumberOfLines;
             if (Math.abs(next - 0.1) < 0.00001 || current - next >= 0 || next - current >= 0.05
                     || Math.abs(next - 1) < 0.00001) {
                 managementUIState.getTargetTableFilters().getBulkUpload().setProgressBarCurrentValue(next);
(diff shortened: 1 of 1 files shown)


hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/management/targettable/BulkUploadHandler.java
```diff
@@ -207,7 +207,7 @@ private void readFileStream(final InputStream tempStream) {
                 eventBus.publish(this, new TargetTableEvent(TargetComponentEvent.BULK_UPLOAD_PROCESS_STARTED));
                 while ((line = reader.readLine()) != null) {
                     innerCounter++;
-                    readEachLine(line, innerCounter, totalFileSize);
+                    readEachLine(line, innerCounter, totalNumberOfLines);
```
schabdo: It makes sense to rename the variable to `totalNumberOfLines`, but than please rename the method parameter `totalFileSize` as well

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/management/targettable/BulkUploadHandler.java
```diff
@@ -193,11 +193,11 @@ public void run() {
 
         private void readFileStream(final InputStream tempStream) {
             String line;
+            final long totalNumberOfLines = getTotalNumberOfLines();
```
schabdo: Do you really have the necessity for type `long`? If so please be consequent and change type of parameter in `readEachLine` as well

Jonkno: Yes, the parameter type in `readEachLine` has allways been `long.` I don't know way it had been converted to double. Didn't make any sense to me.

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/management/targettable/BulkUploadHandler.java
Jonkno: fixed

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/management/targettable/BulkUploadHandler.java
```diff
@@ -269,7 +269,7 @@ private void deleteFile() {
             tempFile = null;
         }
 
-        private void readEachLine(final String line, final double innerCounter, final double totalFileSize) {
+        private void readEachLine(final String line, final double innerCounter, final double totalNumberOfLines) {
```
michahirsch: `totalNumberOfLines` does not have to be a `double`, right? maybe can change the type of the variable then it does not have to be auto-boxed to `double`.

PR #419, opened 2017-01-16
Removed deprecated GridFS column from repository

kaizimmerm:
Refactored repository ArtifactManagement by means of removing the grids column. This is something that was used in the past when MongoDb GridFS artifact binary repository was hardcoded into the system.

I refactored as well the wild mix of identifying the DbBArtifact by artfact ID or SHA1 (especially DMF got this mixed). Its now always SHA1. That is the way to identify the artifact.

Reviewers: @Jonkno @melanieretter


hawkbit-repository/hawkbit-repository-jpa/src/test/java/org/eclipse/hawkbit/repository/jpa/ControllerManagementTest.java
```diff
@@ -81,6 +86,42 @@ public void controllerAddsActionStatus() {
     }
 
     @Test
+    @Description("Verifies that assignement verfication works based on SHA1 hash. By design it is not important which artifact "
```
melanieretter: assignment verification

PR #424, opened 2017-01-20
Consistent content types produced and consistent links format in REST APIs.

kaizimmerm:
Fixes the mess of returned default content types and the accepted ones.

We generally support: application/hal+json (default) and application/json

In addition it fixes the problem that for some resources we did not return the hypermedia links in HAL format.

### DIFF
--- a/hawkbit-mgmt-api/src/main/java/org/eclipse/hawkbit/mgmt/rest/api/MgmtDistributionSetRestApi.java
+++ b/hawkbit-mgmt-api/src/main/java/org/eclipse/hawkbit/mgmt/rest/api/MgmtDistributionSetRestApi.java
@@ -87,8 +87,9 @@ public interface MgmtDistributionSetRestApi {
      *         failure the JsonResponseExceptionHandler is handling the
      *         response.
      */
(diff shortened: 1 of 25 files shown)


hawkbit-mgmt-api/src/main/java/org/eclipse/hawkbit/mgmt/rest/api/MgmtDistributionSetRestApi.java
```diff
@@ -86,7 +88,7 @@
      *         response.
      */
     @RequestMapping(method = RequestMethod.POST, consumes = { MediaType.APPLICATION_JSON_VALUE,
-            "application/hal+json" }, produces = { "application/hal+json", MediaType.APPLICATION_JSON_VALUE })
+            MediaTypes.HAL_JSON_VALUE }, produces = { MediaTypes.HAL_JSON_VALUE, MediaType.APPLICATION_JSON_VALUE })
```
schabdo: I don't think the order within `consumes` matter. Just to be sure that's the only resource I saw were JSON is before HAL.

PR #425, opened 2017-01-26
Fix tag resource for assigned targets.

kaizimmerm:
The resource on assigned targets of a tag behaves differently on the hypermedia links that the other target resources.

Reviewers: @michahirsch @Jonkno 


hawkbit-mgmt-resource/src/main/java/org/eclipse/hawkbit/mgmt/rest/resource/MgmtTargetMapper.java
```diff
@@ -168,7 +148,7 @@ public static MgmtTarget toResponse(final Target target) {
             targetRest.setInstalledAt(installationDate);
         }
 
-        targetRest.add(linkTo(methodOn(MgmtTargetRestApi.class).getTarget(target.getControllerId())).withRel("self"));
+        targetRest.add(linkTo(methodOn(MgmtTargetResource.class).getTarget(target.getControllerId())).withSelfRel());
```
Jonkno: Why are you using the `MgmtTargetResource.class` here and not the `MgmtTargetRestApi.class`?

PR #428, opened 2017-02-01
Fix bulk progress bar

kaizimmerm:
The bulk upload progress bar is not updated during the upload. It basically switches from zero to full. 

In addition the target create behaviour stresses the UI to much, switch to the event driven create behaviour.

Reviewers: @melanieretter @Jonkno 

Fixed a few sonar issues on the way.


hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/management/targettable/BulkUploadHandler.java
```diff
@@ -369,14 +362,11 @@ private void displayValidationMessage(final StringBuilder errorMessage, final St
 
         private void addNewTarget(final String controllerId, final String name) {
             final String newControllerId = HawkbitCommonUtil.trimAndNullIfEmpty(controllerId);
-            if (mandatoryCheck(newControllerId) && duplicateCheck(newControllerId)) {
-                final String newName = HawkbitCommonUtil.trimAndNullIfEmpty(name);
-                final String newDesc = HawkbitCommonUtil.trimAndNullIfEmpty(descTextArea.getValue());
-
-                final Target newTarget = targetManagement.createTarget(entityFactory.target().create()
-                        .controllerId(newControllerId).name(newName).description(newDesc));
+            if (isNotDuplicate(newControllerId)) {
```
Jonkno: Is this check necessary? The createTarget(…) method also hast the duplicate check. You cloud just catch the EntityAlreadyExistsException and adjust your count. Is this cheack neccessary?

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/management/targettable/BulkUploadHandler.java
```diff
@@ -203,46 +208,48 @@ private void readFileStream(final InputStream tempStream) {
...
-            try (InputStreamReader inputStreamReader = new InputStreamReader(new FileInputStream(tempFile),
-                    Charset.defaultCharset())) {
-                try (BufferedReader readerForSize = new BufferedReader(inputStreamReader)) {
-                    totalFileSize = readerForSize.lines().count();
-                }
-            } catch (final FileNotFoundException e) {
-                LOG.error("Error reading file {}", tempFile.getName(), e);
             } catch (final IOException e) {
-                LOG.error("Error while closing reader of file {}", tempFile.getName(), e);
+                LOG.error("Error while reading of file {}", tempFile.getName(), e);
```
Jonkno: You are using a wrong method call. Instead of the e you should provide a second argument not the exception.

PR #429, opened 2017-02-06
Fix cancel action status

kaizimmerm:
When an action is on canceling mode it is currently possible to override that status if DDI or DMF provide intermediate feedback (either for the cancelation itself or the original update).

In that case the action switches back from canceling to running or one of the other status.

This is wrong of course. The action has to stay on cancelling until the cancel operation is confirmed or rejected or the original update completes.


hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/management/actionhistory/ActionStatusIconMapper.java
```diff
@@ -0,0 +1,79 @@
...
+
+import com.google.common.collect.Maps;
+import com.vaadin.server.FontAwesome;
+
+/**
+ * Mapping utility for {@link ActionStatus} to icon in action history table.
+ *
+ */
+public final class ActionStatusIconMapper {
+    static final String STATUS_ICON_GREEN = "statusIconGreen";
```
michahirsch: There are constants for these styles `org.eclipse.hawkbit.ui.utils.SPUIDefinitions`

PR #434, opened 2017-02-10
Show violation details in case a ConstraintViolationException is raised

schabdo:
We show now the details of a constraint violation in the UI

screen shot 2017-02-10 at 16 32 13

Reviewer: @melanieretter and @kaizimmerm

### DIFF
--- a/hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/components/HawkbitUIErrorHandler.java
+++ b/hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/components/HawkbitUIErrorHandler.java
@@ -110,19 +110,17 @@ public class HawkbitUIErrorHandler extends DefaultErrorHandler {
 
     private String extractMessageFrom(final Throwable ex) {
 
-        String errorMessage = ex.getClass().getSimpleName();
-
-        if (ConstraintViolationException.class.isAssignableFrom(ex.getClass())) {
-
-            final Set<ConstraintViolation<?>> violations = ((ConstraintViolationException) ex)
-                    .getConstraintViolations();
-
-            if (violations != null) {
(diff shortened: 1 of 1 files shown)


hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/components/HawkbitUIErrorHandler.java
```diff
@@ -91,10 +98,31 @@ private static Throwable getRootCauseOf(final Throwable ex) {
...
         return new HawkbitErrorNotificationMessage(STYLE, i18n.get("caption.error"),
-                i18n.get("message.error.temp", ex.getClass().getSimpleName()), false);
+                i18n.get("message.error.temp", errorMessage), false);
     }
 
+    private String extractMessageFrom(final Throwable ex) {
+
+        String errorMessage = ex.getClass().getSimpleName();
+
+        if (ConstraintViolationException.class.isAssignableFrom(ex.getClass())) {
```
SirWayne: Would prefer early return

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/components/HawkbitUIErrorHandler.java
```diff
@@ -91,10 +98,31 @@ private static Throwable getRootCauseOf(final Throwable ex) {
...
+
+            final Set<ConstraintViolation<?>> violations = ((ConstraintViolationException) ex)
+                    .getConstraintViolations();
+
+            if (violations != null) {
+                errorMessage = violations.stream().map(v -> v.getPropertyPath() + " " + v.getMessage())
```
SirWayne: v is not a meaningful name

PR #445, opened 2017-02-27
Insert possibility to set an ID for confirmation dialog and fix redirect to rollout overview

melanieretter:
Two new constructors were inserted into ConfirmationDialog.java for enabling setting an ID for the confirmation dialog.
The redirect to the rollout overview was fixed. When the user is on the rollout group overview and clicks on the name of a group for viewing the group targets, he is redirected to the rollout overview if the rollout was deleted in the meantime.

Reviewers: @SirWayne @michahirsch 


hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/common/ConfirmationDialog.java
```diff
@@ -8,6 +8,7 @@
  */
 package org.eclipse.hawkbit.ui.common;
 
+import org.apache.commons.lang3.StringUtils;
```
michahirsch: Please use Guava's `Strings` instead of commons-lang. `Strings.nullOrEmpty(final String s)`

PR #476, opened 2017-04-10
Fix missing TargetUpdate events

kaizimmerm:
Assign DistributionSet does not send assignment events in case cancellations have to be handled first by DMF. However, the mechanism was applied as well to TargetUpdate events which is obviously wrong. The target was changed even in case of cancelations.

I applied as well a few optimisations and small fixes.
* Clear entiymanager after flush (instead of manual detachments)
* Applied standard scheme to action status created by update server.
* jave stream used for calculating cancellations.

Reviewers: @melanieretter @SirWayne

### DIFF
--- a/hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaDeploymentManagement.java
+++ b/hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaDeploymentManagement.java
@@ -222,7 +222,7 @@ public class JpaDeploymentManagement implements DeploymentManagement {
 
         if (targets.isEmpty()) {
             // detaching as it is not necessary to persist the set itself
-            entityManager.detach(set);
+            entityManager.clear();
             // return with nothing as all targets had the DS already assigned
             return new DistributionSetAssignmentResult(Collections.emptyList(), 0, targetsWithActionType.size(),
                     Collections.emptyList(), targetManagement);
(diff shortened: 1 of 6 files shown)


hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaDeploymentManagement.java
```diff
@@ -286,7 +284,9 @@ private DistributionSetAssignmentResult assignDistributionSetToTargets(@NotNull
     private void sendDistributionSetAssignmentEvent(final List<JpaTarget> targets, final Set<Long> targetIdsCancellList,
```
michahirsch: Due the method is now also sending `sendTargetUpdatedEvent` I would probably re-name the method from `sendDistributionSetAssignmentEvent` to `sendAssignmentEvents` or something because it's not only a `DistributionSetAssignmentEvent` anymore.

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaDeploymentManagement.java
```diff
@@ -286,7 +284,9 @@ private DistributionSetAssignmentResult assignDistributionSetToTargets(@NotNull
...
             final Map<String, JpaAction> targetIdsToActions) {
         targets.stream().filter(t -> !!!targetIdsCancellList.contains(t.getId()))
-                .forEach(t -> assignDistributionSetEvent(targetIdsToActions.get(t.getControllerId())));
+                .forEach(t -> sendTargetAssignDistributionSetEvent(targetIdsToActions.get(t.getControllerId())));
```
SirWayne: Do you think it's not better to iterate only once over the collection? Additional it's more readable, or?
```
        targets.forEach(target->{
            sendTargetUpdatedEvent(target);
            if(!targetIdsCancellList.contains(target.getId()){
                return;
            }
            
            sendTargetAssignDistributionSetEvent(targetIdsToActions.get(target.getControllerId());
        });
```

PR #477, opened 2017-04-10
Optimised locking on large scale tenant scenarios.

kaizimmerm:
The current setup for locking the tenant by the schedulers puts pretty high load on the server. Turned out during load testing. This is especially true for installations that use a cluster wide lock implementation (e.g. DB based).

This is refactored by means of lesser locks and less parallel processing for the rollouts.

### DIFF
--- a/hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaRolloutManagement.java
+++ b/hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaRolloutManagement.java
@@ -716,31 +716,31 @@ public class JpaRolloutManagement extends AbstractRolloutManagement {
     // No transaction, will be created per handled rollout
     @Transactional(propagation = Propagation.NEVER)
     public void handleRollouts() {
-        rolloutRepository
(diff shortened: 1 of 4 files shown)


kaizimmerm:
@michahirsch good point. I adapted this.

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaRolloutManagement.java
```diff
@@ -716,31 +716,31 @@ private void executeRolloutGroupSuccessAction(final Rollout rollout, final Rollo
     // No transaction, will be created per handled rollout
     @Transactional(propagation = Propagation.NEVER)
     public void handleRollouts() {
-        rolloutRepository
-                .findByStatusIn(Lists.newArrayList(RolloutStatus.CREATING, RolloutStatus.DELETING,
-                        RolloutStatus.STARTING, RolloutStatus.READY, RolloutStatus.RUNNING))
-                .forEach(this::handleRollout);
-    }
+        final List<Long> rollouts = rolloutRepository.findByStatusIn(Lists.newArrayList(RolloutStatus.CREATING,
+                RolloutStatus.DELETING, RolloutStatus.STARTING, RolloutStatus.READY, RolloutStatus.RUNNING));
```
kaizimmerm: Status list possible

Issue #467, opened 2017-03-31
Problems with the feedback chanel


EugenWiens:
During the update procedure our update agent reports the status to the hawkbit update server. We use following `URL POST /{tenant}/controller/v1/{targetid}/deploymentBase/{actionId}/feedback.` to report the status. After some feedback message we get an error back from hawkbit.
`errorcode:206 {\"exceptionClass\":\"org.eclipse.hawkbit.repository.exception.ConcurrentModificationException\",\"errorCode\":\"hawkbit.server.error.repo.concurrentModification\",\"message\":\"The given entity has been changed by another user/session\"}"`

![image](https://cloud.githubusercontent.com/assets/7880602/24545566/e7a041bc-1607-11e7-8bc0-bb251ee9c54e.png)

![image](https://cloud.githubusercontent.com/assets/7880602/24545651/41407318-1608-11e7-9963-ad3350c0979c.png)

kaizimmerm:
Well the exception said what the problem is. Two changes were pushed on the target at the same time. 

Is that possible in your scenario? One other question, your screenshot looks like that the feedback did arrive successfully? Is that correct?

EugenWiens:
> Well the exception said what the problem is. Two changes were pushed on the target at the same time.
What means at the same time. We report feedback with out queuing every event that occurs is directly pushed to the server. We don't wait for the HTTP OK of the last feedback message. Is this a problem ?

> Is that possible in your scenario? One other question, your screenshot looks like that the feedback did arrive successfully? Is that correct?
In the screenshot you can see the messages that arrived successfully. During the whole procedure we get 3 Errors during sending and the corresponding messages are missing on the server.

kaizimmerm:
This explains the behaviour. Multiple changes at the same time will result in this response. hawkBit has to ensure that one change does not override the other. For instance, one feedback reports successful, one reports error. You are not doing this obviously but hawkBit cannot tell and as a result rejects the update that came in later.

Imho you have the option to report in a synchronised fashion as you said or leave it that way. In this case your client has to consider this error and maybe react with a retry. I would go with this option as the risk that two changes happen on a target might also have other causes and its always good for a client to handle it.

EugenWiens:
Thank you for the information. We will check if we can add a que. We did not think that hawkBit has a problem with this scenario.

kaizimmerm:
You are welcome 😎

PR #478, opened 2017-04-10
Reduce entity changes as part of DDI/DMF update feedback

kaizimmerm:
DDI and DMF report if a targets has polled the API. The repository updates in this case the last query and address fields and sends a TargetPoll event.

However, it does this as feed for all the other calls (e.g. update feedback or artefact download). That puts unnecessary pressure of the repository and raised the probability of concurrent modification errors the device provides lots of feedback in parallel (seen over here as well by other users #467).

Use case wise its also wrong. The event is called TargetPoll for a reason. I reduced these updates accordingly so that they are only done in case of DDI poll and DMF THING_CREATE.

Reviewers: @schabdo

### DIFF
--- a/hawkbit-ddi-resource/src/main/java/org/eclipse/hawkbit/ddi/rest/resource/DdiArtifactStoreController.java
+++ b/hawkbit-ddi-resource/src/main/java/org/eclipse/hawkbit/ddi/rest/resource/DdiArtifactStoreController.java
@@ -22,6 +22,7 @@ import org.eclipse.hawkbit.repository.ControllerManagement;
 import org.eclipse.hawkbit.repository.EntityFactory;
 import org.eclipse.hawkbit.repository.RepositoryConstants;
 import org.eclipse.hawkbit.repository.exception.ArtifactBinaryNotFoundException;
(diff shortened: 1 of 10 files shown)


hawkbit-ddi-resource/src/main/java/org/eclipse/hawkbit/ddi/rest/resource/DdiArtifactStoreController.java
```diff
@@ -134,10 +134,10 @@
         return new ResponseEntity<>(HttpStatus.OK);
     }
 
-    private ActionStatus checkAndReportDownloadByTarget(final HttpServletRequest request, final String targetid,
+    private ActionStatus checkAndReportDownloadByTarget(final HttpServletRequest request, final String controllerId,
             final Artifact artifact) {
-        final Target target = controllerManagement.updateLastTargetQuery(targetid,
-                IpUtil.getClientIpFromRequest(request, securityProperties));
+        final Target target = controllerManagement.findByControllerId(controllerId)
+                .orElseThrow(() -> new EntityNotFoundException(Target.class, controllerId));
```
schabdo: The field `securityProperties` in line 66 is now not used anymore, so you can remove it

PR #483, opened 2017-04-11
ignore IOException on cleanup

michahirsch:
Ignore IOException on clean-up because e.g. on parallel test-execution the folder does not exists anymore and the test crashes e.g. because the folder does not exists to delete.


hawkbit-repository/hawkbit-repository-test/src/main/java/org/eclipse/hawkbit/repository/test/util/AbstractIntegrationTest.java
```diff
@@ -292,7 +292,11 @@ public void before() throws Exception {
 
     @After
     public void cleanCurrentCollection() throws IOException {
-        FileUtils.deleteDirectory(new File(artifactFilesystemProperties.getPath()));
+        try {
+            FileUtils.deleteDirectory(new File(artifactFilesystemProperties.getPath()));
+        } catch (final IOException e) {
```
schabdo: `cleanCurrentCollection()` don't throw an `IOException` anymore. So please remove it from the signature.

PR #484, opened 2017-04-12
Fix default isolation and auto commit

kaizimmerm:
Currently hawkBit is not very flexible concerning the transaction isolation level as these are defined hardcoded for all transactions and cannot be changed. However, depending on the underlying DB infrastructure it might be feasible to leverage a default isolation of the DB or set one for the application (i.e. on driver level).

The following topics where addressed to give hawkBit users the flexibility they (or their DBAs) need:

* Removed the static isolation default definition from the transaction annotations. From now on a transaction defines an isolation only if it has defined needs (e.g. needs READ_COMMITED). Otherwise DB or driver/connection default is used.
  * Fixed several bugs where the repository was actually relying on READ_UNCOMMITED which is risky as not every DB might behave the way we implicitly expect here (i.e. read uncommitted results from other transactions).
* Set EclipseLink flush-mode to COMMIT. Again to ensure that transactions are not relying on an implicit mechanism in this case by means of how the EntityManager flushes inside the transaction.
  * Fixed bugs where in fact transactions worked only due to EclipseLinks default flush behaviour. Execute manual flushes in these cases.

Two smaller issues where addressed as well:

* Disabled auto commit as hawkBit by definition does everything in a transaction (something else would be a bug). We do not need it and this way the transaction manager is not forced to disable/enable it during transaction runtime.
* Removed springs `@Mofifying` annotation from our own transactions. This is actually a leftover back from the day as where the repository was deeply integrated with Spring Data.

Reviewers: @schabdo @michahirsch

### DIFF
--- a/hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaArtifactManagement.java
+++ b/hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaArtifactManagement.java
@@ -32,8 +32,6 @@ import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.data.domain.Page;
 import org.springframework.data.domain.Pageable;
-import org.springframework.data.jpa.repository.Modifying;
-import org.springframework.transaction.annotation.Isolation;
 import org.springframework.transaction.annotation.Transactional;
 import org.springframework.validation.annotation.Validated;
 
@@ -72,7 +70,7 @@ public class JpaArtifactManagement implements ArtifactManagement {
     }
 
     @Override
-    @Modifying
+
     @Transactional
     public Artifact createArtifact(final InputStream stream, final Long moduleId, final String filename,
             final String providedMd5Sum, final String providedSha1Sum, final boolean overrideExisting,
@@ -103,7 +101,7 @@ public class JpaArtifactManagement implements ArtifactManagement {
     }
 
     @Override
-    @Modifying
+
     @Transactional
     public boolean clearArtifactBinary(final String sha1Hash, final Long moduleId) {
 
@@ -122,7 +120,7 @@ public class JpaArtifactManagement implements ArtifactManagement {
     }
 
     @Override
-    @Modifying
+
     @Transactional
     public void deleteArtifact(final Long id) {
         final JpaArtifact existing = (JpaArtifact) findArtifact(id)
@@ -190,7 +188,7 @@ public class JpaArtifactManagement implements ArtifactManagement {
     }
 
     @Override
-    @Modifying
+
     @Transactional
     public Artifact createArtifact(final InputStream inputStream, final Long moduleId, final String filename,
             final boolean overrideExisting) {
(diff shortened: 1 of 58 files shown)


hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaArtifactManagement.java
```diff
@@ -190,8 +185,8 @@ private Artifact storeArtifactMetadata(final SoftwareModule softwareModule, fina
     }
 
     @Override
-    @Modifying
-    @Transactional(isolation = Isolation.READ_UNCOMMITTED)
+
```
michahirsch: not necessary empty line between annoations, would remove it.

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaDeploymentManagement.java
```diff
@@ -141,8 +140,8 @@
     private PlatformTransactionManager txManager;
 
     @Override
-    @Modifying
-    @Transactional(isolation = Isolation.READ_UNCOMMITTED)
+
```
michahirsch: not necessary empty line between annoations, would remove it.

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaTargetManagement.java
```diff
@@ -184,8 +179,8 @@ public void deleteTargets(final Collection<Long> targetIDs) {
     }
 
     @Override
-    @Modifying
-    @Transactional(isolation = Isolation.READ_UNCOMMITTED)
+
```
michahirsch: not necessary empty line between annoations, would remove it.

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaTargetManagement.java
```diff
@@ -407,8 +399,8 @@ public TargetTagAssignmentResult toggleTagAssignment(final Collection<String> co
...
```
michahirsch: not necessary empty line between annoations, would remove it.

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaTargetManagement.java
```diff
@@ -553,23 +545,16 @@ public Long countTargetsByTargetFilterQueryAndNonDS(final Long distributionSetId
...
```
michahirsch: not necessary empty line between annoations, would remove it.

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaTargetManagement.java
```diff
@@ -553,23 +545,16 @@ public Long countTargetsByTargetFilterQueryAndNonDS(final Long distributionSetId
...
-        }
-
-        return targetRepository.save(target);
+        return targetRepository.save(create.build());
     }
 
     @Override
-    @Modifying
-    @Transactional(isolation = Isolation.READ_UNCOMMITTED)
+
```
michahirsch: not necessary empty line between annoations, would remove it.

PR #485, opened 2017-04-12
Fix Rollout UI performance.

kaizimmerm:
Optimised DB calls on Rollouts Management UI:

* Optimised DB call for rollout status count to reduce cost (still a very expensive dude, we have to think about caching here imho).
* Removed lazy query N+1 calls in rollout view (DS fetched, everything else removed). removed broken DS tooltip as part of this.
* Removed N+1 calls for group complete percentage computation.

As always a bit more stream usage as well :)


hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaRolloutManagement.java
```diff
@@ -870,7 +870,7 @@ private void deleteScheduledActions(final JpaRollout rollout, final Slice<JpaAct
 
     @Override
     public Long countRolloutsAll() {
-        return rolloutRepository.count(RolloutSpecification.isDeleted(false));
+        return rolloutRepository.count(RolloutSpecification.isDeletedWithDistributionSet(false));
```
schabdo: DS not needed in count

PR #492, opened 2017-04-20
Tenant delete by query and not select/delete on individual entities.

kaizimmerm:
Spring data by default deletes entities by selecting them and than individually deletes. That results in large transactions and very slow performance and sometimes even causes exceptions from the DB (to large queries).

Solution is defining JQL delete queries. (note: we are not using Springs deleteAllInBatch as we cannot add the tenant manually to the query. EclipseLink will add the tenant of course but I think for such a "delete all" case the risk is a bit high).


hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/TargetRepository.java
```diff
@@ -208,4 +209,15 @@ void setAssignedDistributionSetAndUpdateStatus(@Param("status") TargetUpdateStat
...
+
+    /**
+     * Deletes all {@link TenantAwareBaseEntity} of a given tenant.
+     *
+     * @param tenant
+     *            to delete data from
+     */
+    @Modifying
+    @Transactional
+    @Query("DELETE FROM JpaTenantConfiguration t WHERE UPPER(t.tenant) = UPPER(:tenant)")
```
SirWayne: Replace JpaTenantConfiguration with JpaTarget

PR #496, opened 2017-04-25
Remove unlimited collections from the repository API

kaizimmerm:
hawkBit in many areas of the repository returns unpaged and unlimited collections. This has a risk as the system will be asked to query huge collections of data in certain scenarios. Query 500k entities in one query for sure will have negative impact on the overall system. I can't even say what will explode first (persistence, Java Heap, APIs...). 

Examples on the entity APIs itself are:
* Action: List<ActionStatus> getActionStatus();
* DistributionSet: Set<DistributionSetTag> getTags();

Examples on repository APIs:
* DeploymentManagement: List<Action> findActiveActionsByTarget(@NotEmpty String controllerId);
* TargetManagement: List<Target> findTargetsByTag(@NotEmpty String tagName);

Well, the risk is different depending on the use case. While active actions on a target should in most cases be a two digit number max it is not actually guaranteed. All targets assigned to a tag however might easily go in the 6 digit direction.

I fixed most scenarios by means of providing or using repository APIs with limits. There are some left. Here, however I believe we should introduce configurable quotas (e.g. how many software modules can be in a distribution set). Until then the risk stays in the system.

However, unfortunately on the public APIs we are propagating unpaged access as well (targets or DS assigned to a tag):
* I created new resources with proper paging and restricted the other to 500 entities max and declared them deprecated.
* One API call (remove all targets/DSs from a tag) could not be fixed. I removed it. Technically an API break but I think the use case of that resource is highly questionable (better delete the tag and create a new one). 
* While looking for tests that have to be aligned I figured that tests for tag resources in mgmt API where completely missing. I added those as well.

On the way I fixed a few typos in events and provided a proper API for the quotas we already have in the system. Should be quite easy to extend those in the future.

Note: this is another task of https://github.com/eclipse/hawkbit/issues/197

Reviewers: @schabdo @SirWayne


hawkbit-mgmt-api/src/main/java/org/eclipse/hawkbit/mgmt/rest/api/MgmtRestConstants.java
```diff
@@ -110,9 +120,14 @@
     public static final String TARGET_FILTER_V1_REQUEST_MAPPING = BASE_V1_REQUEST_MAPPING + "/targetfilters";
 
     /**
+     * The deprecated tag URL mapping rest resource.
+     */
+    public static final String DEPRECATED_DISTRIBUTIONSET_REQUEST_MAPPING = "/{distributionsetTagId}/distributionsets";
```
SirWayne: Add `@Deprecated`

hawkbit-mgmt-resource/src/main/java/org/eclipse/hawkbit/mgmt/rest/resource/PagingUtility.java
```diff
@@ -60,6 +61,14 @@ static Sort sanitizeTargetSortParam(final String sortParam) {
         return new Sort(SortUtility.parse(TargetFields.class, sortParam));
     }
 
+    static Sort sanitizeTagSortParam(final String sortParam) {
+        if (sortParam == null) {
+            // default
+            return new Sort(Direction.ASC, TargetFields.NAME.getFieldName());
```
SirWayne: Use `TageFields.Name` instead of Target

hawkbit-repository/hawkbit-repository-core/src/main/java/org/eclipse/hawkbit/repository/StaticQuotaManagement.java
```diff
@@ -0,0 +1,33 @@
...
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package org.eclipse.hawkbit.repository;
+
+import org.eclipse.hawkbit.security.HawkbitSecurityProperties;
+
+public class StaticQuotaManagement implements QuotaManagement {
```
SirWayne: 1. Add javadoc
2. I would rename the class to `PropertiesQuotaManagement`, because you use properties for the quota

hawkbit-repository/hawkbit-repository-core/src/main/java/org/eclipse/hawkbit/repository/RolloutHelper.java
```diff
@@ -68,12 +69,14 @@ public static RolloutGroup verifyRolloutGroupHasConditions(final RolloutGroup gr
      * 
      * @param amountGroup
      *            amount of groups
+     * @param maxGroups
+     *            maximum number of groups allowed
      */
-    public static void verifyRolloutGroupParameter(final int amountGroup) {
+    public static void verifyRolloutGroupParameter(final int amountGroup, final int maxGroups) {
```
SirWayne: In my opinion it's better to make a real bean and inject the `QuotaManagement` here. Because now every developer can set another `maxGroups` constraint for every call.

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaTargetManagement.java
```diff
@@ -556,13 +533,26 @@ public Target createTarget(final TargetCreate c) {
     }
 
     @Override
-    public List<Target> findTargetsByTag(final String tagName) {
-        final Optional<TargetTag> tag = targetTagRepository.findByNameEquals(tagName);
-        if (!tag.isPresent()) {
-            return Collections.emptyList();
+    public Page<Target> findTargetsByTag(final Pageable pageable, final Long tagId) {
+        if (!targetTagRepository.exists(tagId)) {
```
SirWayne: create a private method to avoid duplicate code

kaizimmerm:
@SirWayne I like your idea. All implemented.

examples/hawkbit-example-mgmt-simulator/src/main/java/org/eclipse/hawkbit/mgmt/client/scenarios/ConfigurableScenario.java
```diff
@@ -316,31 +316,38 @@ private void waitUntilRolloutIsReady(final Long id) {
 
     private void createDistributionSets(final Scenario scenario) {
         LOGGER.info("Creating {} distribution sets", scenario.getDistributionSets());
+        final int pages = (scenario.getDistributionSets() / PAGE_SIZE) >= 1
```
schabdo: The page calculation is broken when you slightly exceed the page size. For example: 506 or 501 distribution sets will result in a page size of 1 so in the end this code creates only 500 entities and not the expected amount.

Same happens to the other entities e.g. targets so I won't comment it there as well

hawkbit-mgmt-resource/src/main/java/org/eclipse/hawkbit/mgmt/rest/resource/MgmtTagMapper.java
```diff
@@ -61,8 +62,10 @@ static MgmtTag toResponse(final TargetTag targetTag) {
 
         response.add(linkTo(methodOn(MgmtTargetTagRestApi.class).getTargetTag(targetTag.getId())).withSelfRel());
 
-        response.add(linkTo(methodOn(MgmtTargetTagRestApi.class).getAssignedTargets(targetTag.getId()))
-                .withRel("assignedTargets"));
+        response.add(linkTo(methodOn(MgmtTargetTagRestApi.class).getAssignedTargets(targetTag.getId(),
+                Integer.parseInt(MgmtRestConstants.REQUEST_PARAMETER_PAGING_DEFAULT_OFFSET),
+                Integer.parseInt(MgmtRestConstants.REQUEST_PARAMETER_PAGING_DEFAULT_LIMIT), null, null))
```
schabdo: There is an int value already available: Use `MgmtRestConstants.REQUEST_PARAMETER_PAGING_DEFAULT_LIMIT_VALUE` instead of parsing the string

hawkbit-mgmt-resource/src/main/java/org/eclipse/hawkbit/mgmt/rest/resource/MgmtTagMapper.java
```diff
@@ -93,8 +96,9 @@ static MgmtTag toResponse(final DistributionSetTag distributionSetTag) {
                 linkTo(methodOn(MgmtDistributionSetTagRestApi.class).getDistributionSetTag(distributionSetTag.getId()))
                         .withSelfRel());
 
-        response.add(linkTo(
-                methodOn(MgmtDistributionSetTagRestApi.class).getAssignedDistributionSets(distributionSetTag.getId()))
+        response.add(linkTo(methodOn(MgmtDistributionSetTagRestApi.class).getAssignedDistributionSets(
+                distributionSetTag.getId(), Integer.parseInt(MgmtRestConstants.REQUEST_PARAMETER_PAGING_DEFAULT_OFFSET),
+                Integer.parseInt(MgmtRestConstants.REQUEST_PARAMETER_PAGING_DEFAULT_LIMIT), null, null))
```
schabdo: Same here

hawkbit-repository/hawkbit-repository-jpa/src/test/java/org/eclipse/hawkbit/repository/jpa/RolloutManagementTest.java
```diff
@@ -1100,14 +1105,10 @@ public void createDuplicateRollout() {
...
 
-        try {
-            testdataFactory.createRolloutByVariables(rolloutName, "desc", amountGroups, "id==dup-ro-*", distributionSet,
-                    successCondition, errorCondition);
-            fail("Was able to create a duplicate Rollout.");
-        } catch (final EntityAlreadyExistsException e) {
-            // OK
-        }
-
+        assertThatExceptionOfType(EntityAlreadyExistsException.class)
```
schabdo: Cool 👍

hawkbit-repository/hawkbit-repository-test/src/main/java/org/eclipse/hawkbit/repository/test/util/JpaTestRepositoryManagement.java
```diff
@@ -15,10 +15,15 @@
...
 import org.slf4j.LoggerFactory;
+import org.springframework.data.domain.PageRequest;
+import org.springframework.data.domain.Pageable;
+import org.springframework.data.domain.Sort;
+import org.springframework.data.domain.Sort.Direction;
 
 public class JpaTestRepositoryManagement implements TestRepositoryManagement {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(JpaTestRepositoryManagement.class);
+    private static final Pageable pageReq = new PageRequest(0, 400, new Sort(Direction.ASC, "id"));
```
schabdo: Use CAPITAL LETTERS for variable `pageReq` => it is a constant

hawkbit-repository/hawkbit-repository-test/src/main/java/org/eclipse/hawkbit/repository/test/util/TestdataFactory.java
```diff
@@ -346,6 +346,26 @@ public DistributionSet createDistributionSet(final String prefix, final String v
...
+     * Create a list of {@link DistributionSet}s without modules, i.e.
+     * incomplete.
+     * 
+     * @param number
+     *            of {@link DistributionSet}s to create
+     * @return {@link List} of {@link DistributionSet} entities
+     */
+    public List<DistributionSet> createDistributionSetsWithoutModules(final int number) {
+
+        final List<DistributionSet> sets = new ArrayList<>();
```
schabdo: Set size of list => `new ArrayList<>(number)`

hawkbit-rest-core/src/test/java/org/eclipse/hawkbit/rest/util/JsonBuilder.java
```diff
@@ -33,6 +35,62 @@
...
+    public static String tags(final List<Tag> tags) throws JSONException {
+        final StringBuilder builder = new StringBuilder();
+
+        builder.append("[");
+        int i = 0;
+        for (final Tag tag : tags) {
+            try {
+                createTagLine(builder, tag);
+            } catch (final Exception e) {
+                e.printStackTrace();
```
schabdo: Why not throwing an exception here? It's obviously wrong JSON. It might be easier to find when you stop processing the tags here instead of waiting for the test to fail due to missing tags later

kaizimmerm: Well, the code is copied from the other builders but I actually don't know why the exception is watched. I will fix this in the entire class.

hawkbit-repository/hawkbit-repository-test/src/main/java/org/eclipse/hawkbit/repository/test/util/TestdataFactory.java
kaizimmerm: That does not help much. Lists.newArrayListWithExpectedSize(number) is the right one. But you are generally right of course.

hawkbit-repository/hawkbit-repository-test/src/main/java/org/eclipse/hawkbit/repository/test/util/JpaTestRepositoryManagement.java
kaizimmerm: Copied from abstract integration test :) I will fix this in both classes.

examples/hawkbit-example-mgmt-simulator/src/main/java/org/eclipse/hawkbit/mgmt/client/scenarios/ConfigurableScenario.java
kaizimmerm: Good point

hawkbit-mgmt-resource/src/main/java/org/eclipse/hawkbit/mgmt/rest/resource/MgmtTagMapper.java
kaizimmerm: Interesting. I did not know that we have this value constant. I fixed this on all the mappers and created one for the offset as well.

examples/hawkbit-example-mgmt-simulator/src/main/java/org/eclipse/hawkbit/mgmt/client/scenarios/ConfigurableScenario.java
schabdo: I'm afraid it is still broken. Use `BigDecimal` instead if int or float for the calculation. 

Example:
`new BigDecimal(1002).divide(new BigDecimal(500), BigDecimal.ROUND_UP)` => 3

PR #499, opened 2017-04-27
Renamed new DMF parent folder to be consistent with other parents.

kaizimmerm:


hawkbit-dmf-parent/hawkbit-dmf-api/.gitignore
```diff
@@ -0,0 +1 @@
+/bin/
```
SirWayne: Is this necessary?

PR #505, opened 2017-05-10
Reduce probability of ConcurrentModification errors

kaizimmerm:
Many users are facing ConcurrentModification errors when using hawkBit. These may have it's root cause either in optimistic locking (i.e. two requests are changing an entity at the same time) or pessimistic (e.g. deadlocks from the DB).

I tried to reduce the deadlock probability in one use case (i.e. tag assignment) and activated JDBC batch which should help by means of making bigger transactions faster.

However, we have been optimising in this area for quite some time and as every DBA know you cannot avoid these kind of problems no matter how much you optimise your transactions.

So, the primary change is that I introduce automatic retry in case of ConcurrentModification. All the scenarios we have for them do allow that in my opinion.

I used Spring retry for this. Every write operation is retried up to 10 times for one second. That should reduce probability that it happens by 99,99%.


kaizimmerm:
Done for maxAttempts` and `backoff`. It was not possible for `ConcurrencyFailureException.class`.

Issue #191, opened 2016-05-24
DDI API defines custom string time in body which is not specified


michahirsch:
The DDI-API uses a `time` field in the feedback channel which is intend for clients to write the time of message creation in it. [DDI-API](https://github.com/eclipse/hawkbit/wiki/Direct-Device-Integration-API)

Example

```
{
    "id": 1,
    "time": "20140511T121314",
    "status": {
        "execution": "closed",
        "result": {
            "finished": "success",
            "progress": {}
        }
    }
}
```

Unfortunately this `time` field is not really documented in which format the time is. Furthermore it might make sense to use a standard ISO time format like ISO 8601 standard for JSON dates.

singh-himanshu:
I guess from time to time contributors may want to submit documentation changes. Is the documentation source shared on GitHub already? If not what is the suggested mechanism to submit the documentation changes?

gdesmott:
This ticket has been closed but I did not find any documentation of the expected time format.
Did I miss anything?

schabdo:
You don't missed anything. Back in the days `time` field was specified in API but never implemented/evaluated by hawkBit (cf. [Gitter](https://gitter.im/eclipse/hawkbit?at=5c8b88fdf6999130a00bd751)). So we removed the field from documentation in order to avoid confusion. However it will still be accepted by hawkBit

gdesmott:
Thanks for the clarification. Shouldn't it be removed from https://www.eclipse.org/hawkbit/apis/ddi_api/ as well then?

schabdo:
True, good catch! Removed. Docs will be updated with next milestone.

PR #506, opened 2017-05-10
DDI API: Extend API response to report previous update execution status.

### LINKED ISSUE #381: DDI API: Ability to get informed about update execution status
Currently, a device is only able to send its update execution status but is not able to query it from hawkBit. 
Such a functionality may prove useful, for example, when the client software running on the device loses this information prior to reporting a final execution status such as `closed` to hawkBit. This may happen, e.g., due to a power cycle or simply a crash. Upon the client software restarting, it installs the same payload again as advertised by hawkBit (as the device has not sent a final update execution status). Instead, if the last feedback sent to hawkBit would be reported back to the device, the client may resume installation.

For the implementation, the DDI API response to a [`GET /{tenant}/controller/v1/{targetid}/deploymentBase/{actionId}`](https://docs.bosch-iot-rollouts.com/documentation/rest-api/rootcontroller-api-guide.html#_get_tenant_controller_v1_targetid_deploymentbase_actionid) could include the previously sent feedback via [`POST /{tenant}/controller/v1/{targetid}/deploymentBase/{actionId}/feedback`](https://docs.bosch-iot-rollouts.com/documentation/rest-api/rootcontroller-api-guide.html#_post_tenant_controller_v1_targetid_deploymentbase_actionid_feedback). 
In particular, `status.result.progress` and `status.details` for `status.execution`=`proceeding` is interesting here.

stormc:
Resolves #381

The mechanism is useful for example, when the client software running on
the device loses this information prior to reporting a final execution
status such as 'closed' to hawkBit. This may happen, e.g., due to a
power cycle or simply a crash. Upon the client software restarting, it
installs the same payload again as advertised by hawkBit (as the device
has not sent a final update execution status). Instead, if the last
feedback sent to hawkBit would be reported back to the device, the
client may resume installation.

Feedback messages sent as part of POST
    /{tenant}/controller/v1/{targetid}/deploymentBase/{actionId}/feedback,
are sent back to controller as part of response to GET
    /{tenant}/controller/v1/{targetid}/deploymentBase/{actionId}.

Following example illustrates the API changes:

1. After retrieving the action from server, controller sends:
curl
'http://127.0.0.1:8080/default/controller/v1/1/deploymentBase/1/feedback'
-i -X POST -H 'Content-Type: application/json;charset=UTF-8' -H 'Accept:
application/hal+json' -d '{
  "id" : "1",
  "time" : "20170406T121500",
  "status" : {
    "result" : {
      "progress" : {
        "of" : 1,
        "cnt" : 0
      },
      "finished" : "none"
    },
    "execution" : "proceeding",
    "details" : [ "proceeding with download" ]
  }
}'

2. Once finished downloading the artifact, controller sends
curl
'http://127.0.0.1:8080/default/controller/v1/1/deploymentBase/1/feedback'
-i -X POST -H 'Content-Type: application/json;charset=UTF-8' -H 'Accept:
application/hal+json' -d '{
{
  "id" : "1",
  "time" : "20170406T123000",
  "status" : {
    "result" : {
      "progress" : {
        "of" : 1,
        "cnt" : 0
      },
      "finished" : "none"
    },
    "execution" : "proceeding",
    "details" : [ "downloaded artifacts for update" ]
  }
}'

3. If there is a power outage now, the controller can retrieve the
action again when it restarts.
curl
'http://127.0.0.1:8080/default/controller/v1/1/deploymentBase/1?c=411599879'
-i -H 'Accept: application/hal+json'

Response will be like below
{
  "id": "1",
  "deployment": {
    "download": "forced",
    "update": "forced",
    "chunks": [
      {
      "part": "os",
      "version": "1",
      "name": "1",
      "artifacts": [....],
      }
    ],
  },
  "previousfeedback": {
    "status": "RETRIEVED",
    "type": "FORCED",
    "forcetime": 0,
    "messages": [
      "Update Server: Target retrieved update action and should start
now the download.",
      "Update Server: Target reported PROCEEDING at 20170406T121500",
      "proceeding with download",
      "Update Server: Target retrieved update action and should start
now the download.",
      "Update Server: Target reported PROCEEDING at 20170406T123000",
      "downloaded artifacts for update"
    ],
  }
}

4. Based on the feedback messages, controller may be able to skip the
download and resume with installation and send additional feedback
curl
'http://127.0.0.1:8080/default/controller/v1/1/deploymentBase/1/feedback'
-i -X POST -H 'Content-Type: application/json;charset=UTF-8' -H 'Accept:
application/hal+json' -d '{
  "id" : "1",
  "time" : "20170406T121314",
  "status" : {
    "result" : {
      "progress" : {
        "of" : 1,
        "cnt" : 0
      },
      "finished" : "none"
    },
    "execution" : "resumed",
    "details" : [ "resuming installation based on previous feedback,
download skipped" ]
  }
}'

### DIFF
--- a/hawkbit-ddi-api/src/main/java/org/eclipse/hawkbit/ddi/json/model/DdiStatus.java
+++ b/hawkbit-ddi-api/src/main/java/org/eclipse/hawkbit/ddi/json/model/DdiStatus.java
@@ -13,6 +13,7 @@ import java.util.List;
 
 import javax.validation.Valid;
 import javax.validation.constraints.NotNull;
+import javax.validation.constraints.Size;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
@@ -31,6 +32,7 @@ public class DdiStatus {
     @Valid
     private final DdiResult result;
 
+    @Size(min = 0, max = 50)
     private final List<String> details;
 
     /**
--- a/hawkbit-ddi-resource/src/main/java/org/eclipse/hawkbit/ddi/rest/resource/DataConversionHelper.java
+++ b/hawkbit-ddi-resource/src/main/java/org/eclipse/hawkbit/ddi/rest/resource/DataConversionHelper.java
@@ -118,7 +118,7 @@ public final class DataConversionHelper {
                 result.add(ControllerLinkBuilder
                         .linkTo(ControllerLinkBuilder.methodOn(DdiRootController.class, tenantAware.getCurrentTenant())
                                 .getControllerBasedeploymentAction(tenantAware.getCurrentTenant(),
-                                        target.getControllerId(), action.getId(), calculateEtag(action)))
+                                        target.getControllerId(), action.getId(), calculateEtag(action), null))
                         .withRel(DdiRestConstants.DEPLOYMENT_BASE_ACTION));
             }
         }
--- a/hawkbit-ddi-resource/src/main/java/org/eclipse/hawkbit/ddi/rest/resource/DdiRootController.java
+++ b/hawkbit-ddi-resource/src/main/java/org/eclipse/hawkbit/ddi/rest/resource/DdiRootController.java
@@ -19,6 +19,7 @@ import javax.validation.Valid;
 import org.eclipse.hawkbit.api.ArtifactUrlHandler;
 import org.eclipse.hawkbit.artifact.repository.model.DbArtifact;
 import org.eclipse.hawkbit.ddi.json.model.DdiActionFeedback;
+import org.eclipse.hawkbit.ddi.json.model.DdiActionHistory;
 import org.eclipse.hawkbit.ddi.json.model.DdiCancel;
 import org.eclipse.hawkbit.ddi.json.model.DdiCancelActionToStop;
 import org.eclipse.hawkbit.ddi.json.model.DdiChunk;
@@ -28,6 +29,7 @@ import org.eclipse.hawkbit.ddi.json.model.DdiDeployment;
 import org.eclipse.hawkbit.ddi.json.model.DdiDeployment.HandlingType;
 import org.eclipse.hawkbit.ddi.json.model.DdiDeploymentBase;
 import org.eclipse.hawkbit.ddi.json.model.DdiResult.FinalResult;
+import org.eclipse.hawkbit.ddi.rest.api.DdiRestConstants;
 import org.eclipse.hawkbit.ddi.rest.api.DdiRootControllerRestApi;
 import org.eclipse.hawkbit.repository.ArtifactManagement;
 import org.eclipse.hawkbit.repository.ControllerManagement;
@@ -231,7 +233,8 @@ public class DdiRootController implements DdiRootControllerRestApi {
     public ResponseEntity<DdiDeploymentBase> getControllerBasedeploymentAction(
             @PathVariable("tenant") final String tenant, @PathVariable("controllerId") final String controllerId,
             @PathVariable("actionId") final Long actionId,
-            @RequestParam(value = "c", required = false, defaultValue = "-1") final int resource) {
+            @RequestParam(value = "c", required = false, defaultValue = "-1") final int resource,
+            @RequestParam(value = "actionHistory", defaultValue = DdiRestConstants.NO_ACTION_HISTORY) final Integer actionHistoryMessageCount) {
         LOG.debug("getControllerBasedeploymentAction({},{})", controllerId, resource);
 
         final Target target = controllerManagement.findByControllerId(controllerId)
@@ -251,8 +254,15 @@ public class DdiRootController implements DdiRootControllerRestApi {
 
             final HandlingType handlingType = action.isForce() ? HandlingType.FORCED : HandlingType.ATTEMPT;
 
+            List<String> actionHistoryMsgs = controllerManagement.getActionHistoryMessages(action.getId(),
+                    actionHistoryMessageCount == null ? Integer.parseInt(DdiRestConstants.NO_ACTION_HISTORY)
+                            : actionHistoryMessageCount);
+
+            DdiActionHistory actionHistory = actionHistoryMsgs.isEmpty() ? null
+                    : new DdiActionHistory(action.getStatus().name(), actionHistoryMsgs);
+
             final DdiDeploymentBase base = new DdiDeploymentBase(Long.toString(action.getId()),
-                    new DdiDeployment(handlingType, handlingType, chunks));
+                    new DdiDeployment(handlingType, handlingType, chunks), actionHistory);
 
             LOG.debug("Found an active UpdateAction for target {}. returning deyploment: {}", controllerId, base);
 
--- a/docs/src/main/resources/documentation/interfaces/ddi-api.md
+++ b/docs/src/main/resources/documentation/interfaces/ddi-api.md
@@ -97,7 +97,14 @@ _Example Response_
             }
         ]
     },
-    "id": "1"
+    "id": "1",
+    "actionHistory": {
+        "status": "RETRIEVED",
+        "messages": [
+            "Installing update",
+            "Downloading artifacts"
+            ]
+    }
 }
 ```
 
--- a/hawkbit-ddi-api/src/main/java/org/eclipse/hawkbit/ddi/json/model/DdiActionHistory.java
+++ b/hawkbit-ddi-api/src/main/java/org/eclipse/hawkbit/ddi/json/model/DdiActionHistory.java
@@ -1,0 +1,55 @@
+/**
+ * Copyright (c) Siemens AG, 2017
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package org.eclipse.hawkbit.ddi.json.model;
+
+import java.util.List;
+
+import org.eclipse.hawkbit.ddi.rest.api.DdiRootControllerRestApi;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonPropertyOrder;
+
+/**
+ * Provide action history information to the controller as part of response to
+ * {@link DdiRootControllerRestApi#getControllerBasedeploymentAction}: 1.
+ * Current action status at the server; 2. List of messages from action history
+ * that were sent to server earlier by the controller using
+ * {@link DdiActionFeedback}.
+ */
+
+@JsonPropertyOrder({ "status", "messages" })
+public class DdiActionHistory {
+
+    @JsonProperty("status")
+    private final String actionStatus;
+
+    @JsonProperty("messages")
+    private final List<String> messages;
+
+    /**
+     * Parameterized constructor for creating {@link DdiActionHistory}.
+     *
+     * @param actionStatus
+     *            is the current action status at the server
+     * @param messages
+     *            is a list of messages retrieved from action history.
+     */
+    @JsonCreator
+    public DdiActionHistory(@JsonProperty("status") final String actionStatus,
+            @JsonProperty("messages") List<String> messages) {
+        this.actionStatus = actionStatus;
+        this.messages = messages;
+    }
+
+    @Override
+    public String toString() {
+        return "Action history [" + "status=" + actionStatus + ", messages={" + messages.toString() + "}]";
+    }
+}
--- a/hawkbit-ddi-api/src/main/java/org/eclipse/hawkbit/ddi/json/model/DdiDeploymentBase.java
+++ b/hawkbit-ddi-api/src/main/java/org/eclipse/hawkbit/ddi/json/model/DdiDeploymentBase.java
@@ -13,19 +13,31 @@ import javax.validation.constraints.NotNull;
 import org.springframework.hateoas.ResourceSupport;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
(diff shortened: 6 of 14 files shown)


stormc:
Hm, the circleci is complaining about a missing header in [DdiPreviousActionFeedback.java](https://github.com/eclipse/hawkbit/pull/506/commits/b3e01f7711d4feb4c4f16a34e9e85b54803d4a23#diff-961f0d882ec6104d2c4c840ac261b4e2R2) which however is there. Is there some constraint on what to put in which order there that I'm not aware of?

michahirsch:
Thanks a lot for the contribution.

There is an issue with the [LICENCE_HEADER_TEMPLATE.txt](https://github.com/eclipse/hawkbit/blob/master/LICENCE_HEADER_TEMPLATE.txt) which needs to be fixed! Sorry for that.

RajHS:
@michahirsch , instead of a verbose we can make the number of feedback messages configurable using some property `http://127.0.0.1:8080/default/controller/v1/1/deploymentBase/1?c=411599879&msgcnt=3`. If the property is not set or set to Zero then no feedback message is sent.

michahirsch:
@RajHS sounds good 👍 I also would introduce an own flag for the history, is better than just a general `verbose` flag

michahirsch:
The license-header issue is probably not easy to fix, I guess we need to disable the automatic license check/generation, other projects like che also having trouble with it. But I'll check first if we can find a workaround otherwise we disable the automatic check for the builds.

singh-himanshu:
@michahirsch, @RajHS one way to address licenses is to add multiple license files:

move licenses to a folder
	licenses/LICENCE_HEADER_TEMPLATE_BOSCH.txt
	licenses/LICENCE_HEADER_TEMPLATE_SIEMENS.txt

add to com.mycil plugin configuration in parent pom.xml
	<header>licenses/LICENCE_HEADER_TEMPLATE_BOSCH.txt</header>
	<validHeaders>
	    <validHeader>licenses/LICENCE_HEADER_TEMPLATE_SIEMENS.txt</validHeader>
	</validHeaders>

singh-himanshu:
@michahirsch, @RajHS, once can manage licenses in following way:

move licenses to a folder
	licenses/LICENCE_HEADER_TEMPLATE_BOSCH.txt
	licenses/LICENCE_HEADER_TEMPLATE_SIEMENS.txt

add to com.mycil plugin configuration in parent pom.xml
	\<header>licenses/LICENCE_HEADER_TEMPLATE_BOSCH.txt\</header>
	\<validHeaders>
	    \<validHeader>licenses/LICENCE_HEADER_TEMPLATE_SIEMENS.txt\</validHeader>
	\</validHeaders>

singh-himanshu:
Recent changes due to #496 removes an API from Action.java `(getActionStatus())` that was used within this patch, hence will have to re-do this patch. I believe circleCI should have complained for DDI Resources but did not.

kaizimmerm:
#512 is on master. Thanks for the fix btw. :)

kaizimmerm:
@singh-himanshu yeah, I am sorry for that removal. But this unpaged DB access was pretty problematic. 

I will create a proper method for you in ControllerManagement. Give me an hour.

RajHS:
@kaizimmerm Currently, we thought of introducing another method which provides a wrapper in ControllerManagement over `actionStatusRepository.findByActionId` which collects the list of action statuses. Let us know if you would suggest an alternate approach.

kaizimmerm:
Jupp, that is exactly what I was going for. I created a pull request here: #514

You can wait until that is on master or do it on your own.

kaizimmerm:
Its on master. I hope that helps.

hawkbit-ddi-resource/src/main/java/org/eclipse/hawkbit/ddi/rest/resource/DdiRootController.java
```diff
@@ -251,14 +253,28 @@ private static boolean checkModule(final String fileName, final SoftwareModule m
 
             final HandlingType handlingType = action.isForce() ? HandlingType.FORCED : HandlingType.ATTEMPT;
 
-            final DdiDeploymentBase base = new DdiDeploymentBase(Long.toString(action.getId()),
-                    new DdiDeployment(handlingType, handlingType, chunks));
+            DdiPreviousActionFeedback previousActionFeedback = null;
+            Optional<Action> actionWithDetails = controllerManagement.findActionWithDetails(actionId);
+
+            if (actionWithDetails.isPresent()) {
```
kaizimmerm: if you use ifPresent you can avoid all these get() later on

hawkbit-ddi-resource/src/main/java/org/eclipse/hawkbit/ddi/rest/resource/DdiRootController.java
```diff
@@ -251,14 +253,28 @@ private static boolean checkModule(final String fileName, final SoftwareModule m
...
+                List<String> feedbackMessages = new ArrayList<>();
 
-            LOG.debug("Found an active UpdateAction for target {}. returning deyploment: {}", controllerId, base);
+                actionWithDetails.get().getActionStatus().stream()
+                        .forEach(actionStatus -> feedbackMessages.addAll(actionStatus.getMessages()));
```
kaizimmerm: I think using `flatMap` instead for forach->addAll would make things easier readable.

kaizimmerm:
While thinking again about it I believe instead of pulling everting together in single queries from the repository and than flattening it in Java as hinted by me above it will be better to query the flattened message list of the action directly from the repository, i.e. create in `ControllerManagement` a method findMessagesByActionIdNotLike(page request, actionid, string).

I good starting point for that is `findMessagesByActionStatusId` in `DeploymentManagement` but based on action id and including the filter to sort out the server generated messages.

This way we would save lots of queries.

Let me know what you think.

hawkbit-ddi-resource/src/main/java/org/eclipse/hawkbit/ddi/rest/resource/DdiRootController.java
singh-himanshu: It does make sense to prune out the server generated messages. Also I think the name "previousfeedback" is a bit confusing, rather "targethistory" seems more apt since we are only interested in what the target/controller reported.

singh-himanshu:
Since the last API changes, we added additional API  `controllerManagement.getActionStatusHistory` that in turn uses findActionStatusByAction to get the list of maximum 100 action status and then flatten them into a list.
List<ActionStatus> actionStatusList = findActionStatusByAction(pageable, actionId).getContent();

I think this way we still have one query to get recent action statuses, but subsequent extraction of messages and filtering the server generated messages is being done in Java rather than in DB. Since the default behavior is not to extract and send back any message, the existing controller should not experience any significant performance impact.

Later we may want to do all this in DB as @kaizimmerm suggested.

kaizimmerm:
I don't think this is going to work. Lets assume you have an action with 1000 action status entries, each 100 messages. You are going to query all of that from the DB and loading it into heap just to skip than 99% of it. This is highly inefficient and especially in DDI will practically kill the system.

hawkbit-ddi-resource/src/main/java/org/eclipse/hawkbit/ddi/rest/resource/DdiRootController.java
```diff
@@ -326,7 +344,10 @@ private ActionStatusCreate generateUpdateStatus(final DdiActionFeedback feedback
         }
 
         if (feedback.getStatus().getDetails() != null) {
-            messages.addAll(feedback.getStatus().getDetails());
+            messages.addAll(feedback.getStatus()
```
kaizimmerm: I don't think that you can do it that way. First of all people may rely on the structure of these messages (as you intend to do). So they might rely on the fact that a DDI provided message "A" is represented in management API as "A". The second problem is that we haven't used the time field yet as we have an API problem there (see #191). Last but not least I don't think that we should store this field in every message in the DB, kind of inefficient.

Long story short if you need this time field in your use-case I kindly ask you to introduce a new DB column for ActionStatus, store the time in there and fix the time field in the API (see #191). I am not sure however how to properly return it in your DDI extension. I personally would go for:

1) First guaranteeing that the time field is filled in the DB (i.e. by setting it with server TIME if not provided by controller).
2) Structure the messages in your extension by the time (this way the controller even sees which messages belonged together).

singh-himanshu: Yes, I think we will remove the time stamp prefix. But will still just keep the message list flat without additional structure. We can leave it to controller to add meaning to the messages and reverse interpret them.

Other issue is that of getting the flattened list of messages directly from database using the filter. Having messages in `JpaActionStatus` as `ElementCollection` with a `CollectionTable` doesn't seem to work. What we are doing is actually a query similar to `.... detail_message NOT LIKE 'Update Server%'`. Criteria API cannot filter out the elements from the element collection, either all or none gets returned. The `CriteriaBuilder.notLike` expects `Element<String>` and not `Element<List<String>>`. This may be a JPA restriction so I'm not sure even if JPQL or Spring Query Methods is going to do that.

I'm considering following options:
1. Instead of restricting by count of messages, we restrict by count of action status, and then filter out the server generated messages in Java. We can use a method similar to `DeploymentManagement.findMessagesByActionStatusId` for this.
2. Change the mapping to OneToMany, but this required JPA model change which is something that I want to avoid.

kaizimmerm:
```
@Query("SELECT message FROM JpaActionStatus acs JOIN acs.messages message WHERE acs.action.id = :actionId AND message NOT LIKE :filter")
    Page<String> findMessagesByActionIdAndMessageNotLike(Pageable pageReq,
            @Param("actionId") Long actionId, @Param("filter") String filter);
```

should do the trick.

kaizimmerm: Note: that query will page/limit and order based on the action status but I think that is ok. Important is that the DB takes care of the filtering.

singh-himanshu: Thanks @kaizimmerm. This works and looks much cleaner than the Criteria API. Will soon update the merge request soon.

hawkbit-ddi-api/src/main/java/org/eclipse/hawkbit/ddi/json/model/DdiStatus.java
```diff
@@ -31,6 +32,7 @@
     @Valid
     private final DdiResult result;
 
+    @Size(min = 0, max = 50)
```
kaizimmerm: Can you explain what this is for?

hawkbit-ddi-resource/src/main/java/org/eclipse/hawkbit/ddi/rest/resource/DataConversionHelper.java
```diff
@@ -118,7 +118,8 @@ static DdiControllerBase fromTarget(final Target target, final Action action,
                 result.add(ControllerLinkBuilder
                         .linkTo(ControllerLinkBuilder.methodOn(DdiRootController.class, tenantAware.getCurrentTenant())
                                 .getControllerBasedeploymentAction(tenantAware.getCurrentTenant(),
-                                        target.getControllerId(), action.getId(), calculateEtag(action)))
+                                        target.getControllerId(), action.getId(), calculateEtag(action),
+                                        Integer.parseInt(DdiRestConstants.NO_ACTION_HISTORY)))
```
kaizimmerm: I think it would be better to put null in here (and change signature to Integer). This is the default value anyhow so no reason to put it into the link.

hawkbit-ddi-api/src/main/java/org/eclipse/hawkbit/ddi/json/model/DdiStatus.java
singh-himanshu: This is explained to some extent in the first commit that is part of this pull request: 

" Even though the number of ActionStatus entries is limited to 1000 per action using QuotaManagement, there is no limit on the number of messages that can be sent as part of a single ActionStatus. This allows a controller to potentially send large number of messages for a single action status. Hence, limit the number of messages allowed to 50 using the javax.validation.constraints.Size within DdiStatus."

Since there is no limitation of messages per action status, currently it is possible that the last action status reports for example 200 or even larger number of messages. Besides the potential of flooding the database, we may only get be able to retrieve part of the last feedback.

This is because we just get a flat list of messages based on the message count and limit the maximum possible count to 100 messages for retrieval. With limit of 50 above, we are can get back at least 2 complete set of action status messages.

kaizimmerm: Ok, understood. However, if you do it that way its basically not configurable. But I think we can keep for now. 

If its ok for you I would later move this into the QuotaManagement service so that hawkBit operators can customise this with a property.

hawkbit-ddi-resource/src/main/java/org/eclipse/hawkbit/ddi/rest/resource/DataConversionHelper.java
singh-himanshu: Yes.. I think it makes sense not to add default values of parameters to the link.

hawkbit-ddi-api/src/main/java/org/eclipse/hawkbit/ddi/json/model/DdiStatus.java
stormc: Yes, this is fine for us. Thanks!

kaizimmerm:
Thanks for the contribution 👍

PR #508, opened 2017-05-11
Device simulator DMF target delete aware

kaizimmerm:


examples/hawkbit-device-simulator/src/main/java/org/eclipse/hawkbit/simulator/amqp/SpReceiverService.java
```diff
@@ -71,20 +78,26 @@ public SpReceiverService(final RabbitTemplate rabbitTemplate, final AmqpProperti
...
     }
 
-    private void delegateMessage(final Message message, final String type, final String thingId) {
+    private void delegateMessage(final Message message, final String type, final String thingId, final String tenant) {
         final MessageType messageType = MessageType.valueOf(type);
 
         if (MessageType.EVENT.equals(messageType)) {
             handleEventMessage(message, thingId);
             return;
+        } else if (MessageType.THING_DELETED.equals(messageType)) {
```
SirWayne: You can remove the `else`, because of the early return 👍

PR #509, opened 2017-05-11
Introduce cache for rollouts status computation.

kaizimmerm:
Status overview for rollouts is actually pretty expensive on the database and may take very long depending on the number of devices. 

I improved the queries a bit and introduced a cache to improve performance both for rollouts as well as rollout groups. Cache is invalidated through change events.

Reviewers: @melanieretter @Jonkno 


kaizimmerm:
#512 is on master. Thanks for the fix btw. :)

kaizimmerm:
...and would only be a problem when that happens during that last action of that rollout. Winning the jackpot has a higher probability ... :)

PR #510, opened 2017-05-12
remove license:check from sonar-build due multiple copyrights problem

michahirsch:
Removing the `license:check` goal from the sonar-build due problems of multiple copyrights e.g. from pull-requests from other organizations.

Related #506.

There is no easy way to support a template with the mycila license-plugin to allow different copyrights.
See https://github.com/mycila/license-maven-plugin/issues/119

The eclipse-che project are also looking for a solution
https://github.com/eclipse/che/issues/3281


michahirsch:
Close in favor of  #512

PR #512, opened 2017-05-12
Introduce license header templates

stormc:
To support multiple license headers simultaneously, a license folder
for holding licenses is introduced. The original "Bosch Software
Innovations GmbH" license header is set as default and the Siemens
license header is added as an example for a further valid license.


michahirsch:
I like this idea and it's better than remove the `license:check` goal from the build https://github.com/eclipse/hawkbit/pull/510 

@kaizimmerm what do you think?

kaizimmerm:
Jupp, good workaround :)

PR #526, opened 2017-05-29
Fix to many request filter usage in DDI

kaizimmerm:
I did not got it fully right in https://github.com/eclipse/hawkbit/pull/522

- I fixed that problem (tenant part in DDI urls cannot be handled by regular servlet filters).
- Cleanup a little bit as well.

Reviewers: @SirWayne @jonkno


hawkbit-autoconfigure/src/main/java/org/eclipse/hawkbit/autoconfigure/security/SecurityManagedConfiguration.java
```diff
@@ -295,15 +299,52 @@ public FilterRegistrationBean eTagFilter() {
         // in the ShallowEtagHeaderFilter, just using the SH1 hash of the
         // artifact itself as 'ETag', because otherwise the file will be copied
         // in memory!
-        filterRegBean.setFilter(
-                new ExcludePathAwareShallowETagFilter("/rest/v1/softwaremodules/{smId}/artifacts/{artId}/download",
-                        "/{tenant}/controller/v1/{controllerId}/softwaremodules/{softwareModuleId}/artifacts/**",
-                        "/api/v1/downloadserver/**"));
+        filterRegBean.setFilter(new ExcludePathAwareShallowETagFilter("/UI/**",
```
Jonkno: Double check the @Order() annotations in this class to have a clear structure. I saw that @Order(100) eTagFilter comes after @Order(300) ControllerSecurityConfigurationAdapter.

hawkbit-autoconfigure/src/main/java/org/eclipse/hawkbit/autoconfigure/security/SecurityManagedConfiguration.java
```diff
@@ -411,7 +455,7 @@ public void destroy() {
         @Bean
         public FilterRegistrationBean dosMgmtUiFilter(final HawkbitSecurityProperties securityProperties) {
 
-            final FilterRegistrationBean filterRegBean = dosFilter(securityProperties.getDos().getUiFilter(),
+            final FilterRegistrationBean filterRegBean = dosFilter(null, securityProperties.getDos().getUiFilter(),
                     securityProperties.getClients());
             // All URLs that can be called anonymous
             filterRegBean.setUrlPatterns(Arrays.asList("/UI/login", "/UI/login/*", "/UI/logout", "/UI/logout/*"));
```
Jonkno: What is with "/", "/UI" ... DOS filter will not react in these cases

kaizimmerm: Those are redirects.

PR #528, opened 2017-05-31
Restructuring properties

melanieretter:
Restructuring properties for better organization and no need to hold the same information multiple times in different files.

Reviewers: @schabdo @SirWayne 


hawkbit-core/src/main/java/org/eclipse/hawkbit/tenancy/configuration/validator/TenantConfigurationValidatorException.java
```diff
@@ -1,3 +1,4 @@
+
```
SirWayne: Remove line

SirWayne: This line is wrong, or?

PR #531, opened 2017-06-06
New repository quota on messages per ActionStatus

kaizimmerm:
Makes the quota configurable in repository compared to hard coded constraint check in API.

Reviewers: @SirWayne @melanieretter


hawkbit-ddi-resource/src/test/java/org/eclipse/hawkbit/ddi/rest/resource/DdiDeploymentBaseTest.java
```diff
@@ -563,6 +563,31 @@ public void toMuchDeplomentActionFeedback() throws Exception {
     }
 
     @Test
+    @Description("The server protects itself against to large feedback bodies. The test verfies that "
+            + "it is not possible to exceed the configured maximum number of feedback details.")
+    public void tooMuchDeplomentActionMessagesInFeedback() throws Exception {
```
kaizimmerm: y

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaControllerManagement.java
```diff
@@ -278,6 +276,7 @@ public Action addCancelActionStatus(final ActionStatusCreate c) {
         default:
             // information status entry - check for a potential DOS attack
             checkForToManyStatusEntries(action);
+            checkForToManyStatusMessages(actionStatus);
```
kaizimmerm: oo

singh-himanshu:
This is similar (almost) to how we initially wanted to restrict the number of status messages per action status. The drawback in my opinion is that this way the object is already deserialised to heap, so essentially we will only be able to protect the server against large number of database inserts but the heap utilisation can still be high.
Idea behind using Bean validation was that it can be used to check for constraints while deserialising and save unnecessary heap utilisation. I'm not sure whether Spring/Jackson already does it this way.

hawkbit-ddi-resource/src/test/java/org/eclipse/hawkbit/ddi/rest/resource/DdiDeploymentBaseTest.java
```diff
@@ -563,6 +563,31 @@ public void toMuchDeplomentActionFeedback() throws Exception {
     }
 
     @Test
+    @Description("The server protects itself against to large feedback bodies. The test verfies that "
```
melanieretter: too, verifies

hawkbit-repository/hawkbit-repository-api/src/main/java/org/eclipse/hawkbit/repository/exception/QuotaExceededException.java
```diff
@@ -0,0 +1,63 @@
...
+     * 
+     * @param type
+     *            that hit quota
+     * @param inserted
+     *            cause for the hit
+     * @param quota
+     *            that is defined by the repository
+     */
+    public QuotaExceededException(final String type, final long inserted, final int quota) {
+        super("Too many entries of " + type + " in the request. {" + inserted + "} is bejond the permitted {" + quota
```
melanieretter: beyond.
How do you like this?
Request contains too many entries of [type]:  .....

kaizimmerm:
@singh-himanshu You are absolutely right the quota is not meant to protect the server against heap overloads. To avoid that you can set on application server level a maximum request size (e.g. `server.max-http-post-size`)

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaControllerManagement.java
```diff
@@ -366,10 +374,8 @@ private void checkForToManyStatusEntries(final JpaAction action) {
             final Long statusCount = actionStatusRepository.countByAction(action);
 
             if (statusCount >= quotaManagement.getMaxStatusEntriesPerAction()) {
-                LOG_DOS.error(
-                        "Potential denial of service (DOS) attack identfied. More status entries in the system than permitted ({})!",
+                throw new QuotaExceededException(ActionStatus.class, statusCount,
```
SirWayne: I think `>` instead of `>=` is better

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaControllerManagement.java
SirWayne: oo

PR #534, opened 2017-06-06
Remove rabbit mq custom heartbeat property

SirWayne:
Remove rabbit mq custom heartbeat property, because rabbit has it by itself


schabdo:
In the meanwhile `RabbitProperties` are providing even more properties to set. I think you can replace `maxConcurrentConsumers`, `prefetchCount`, `initialConcurrentConsumers` and `declarationRetries` as well

PR #535, opened 2017-06-07
245 maintenance window

### LINKED ISSUE #245: Download and deferred installation in a maintenance window
Currently, hawkBit's mode is download+install/activate, i.e., as soon as a distributionset is assigned to a target, the target device gets to know about the available update upon its next poll operation, subsequently downloading, installing, and activating the update.
Although one can orchestrate the start time of announcing updates via, e.g., rollout management, it would be nice to have, e.g., connectivity-constrained devices to download the update artifact beforehand and then to install the downloaded artifact once the defined maintenance window is reached.

This may be achieved by splitting download and installation/activation into two steps, thereby putting the burden of keeping track of the maintenance window upon hawkBit. As a result -- although not strictly required as the artifact should have been downloaded by then -- devices need to be online in order to receive the install/activate order from hawkBit. Furthermore, this would be a (major) API change.
As an alternative, the maintenance window may be communicated to the device which downloads the artifact but defers installation/activation until the maintenance window is reached. This would require to extend the feedback options in hawkBit to reflect the "downloaded but waiting for maintenance window" state. If no maintenance window is specified, the current behavior is maintained.

stormc:
Resolves #245

This feature allows the splitting of download and installation/activation
into two steps. The device gets the update status as 'skip' and can
download and keep the artifacts ready for installation. Then, the device
needs to be online in order to receive the install/activate from hawkBit
through an ETAG update and the change in update state to 'force'.
Additionally, since the device will not start the update unless 'commanded'
by its server, it needs to be able to fetch the command from hawkBit near
to the start of the maintenance window. For this purpose, the polling interval
near to the start of the maintenance window is shortened so that the window is
not missed if the default polling interval is longer.

1. Specify maintenance window while assigning a distribution set to a target.
```
[2017-04-11T02:25:00+0200] $ curl 'http://127.0.0.1:8080/rest/v1/targets/2/assignedDS' -i -X POST -H 'Content-Type: application/json;charset=UTF-8' -d '{
    "forcetime" : "0",
    "id" : "1",
    "type" : "forced",
    "maintenanceWindow" : {
        "maintenanceWindowStartTime" : "2017-04-11T02:45:00+0200",
        "maintenanceWindowEndTime" : "2017-04-11T03:00:00+0200"
    }
}'
```

2. The device gets to know the scheduled action when it polls next.
```
[2017-04-11T02:30:00+0200] $ curl 'http://127.0.0.1:8080/default/controller/v1/2' -i -H 'Accept: application/hal+json'
HTTP/1.1 200 OK
Content-Type: application/hal+json;charset=UTF-8
Content-Length: 356
{
    "config": {
        "polling": {
        "sleep": "00:05:00"
        }
    },
    "_links": {
        "deploymentBase": {
            "href":
"http://127.0.0.1:8080/default/controller/v1/2/deploymentBase/1?c=411629670"
        },
        "configData": {
            "href":
"http://127.0.0.1:8080/default/controller/v1/2/configData"
        }
    }
}

[2017-04-11T02:30:30+0200] $ curl 'http://127.0.0.1:8080/default/controller/v1/2/deploymentBase/1?c=411629670' -i -H 'Accept: application/hal+json'
HTTP/1.1 200 OK
Content-Type: application/hal+json;charset=UTF-8
Content-Length: 1665
{
    "id": "1",
    "deployment": {
        "download": "forced",
        "update": "skip",
        "chunks": [ {
            "part": "os",
            "version": "1",
            "name": "1",
            "artifacts" : [ {
                "filename" : "binary.tgz",
                "hashes" : {
                    "sha1" : "8b71973e835cd0718827238b1cf89f0079e44dae",
                    "md5" : "a90218782025c48e09ba0436cf30d029"
                },
                "size" : 5,
               "_links" : {
                    "download" : {
                        "href" : "http://127.0.0.1:8080/api/v1/default/download/controller/2/softwaremodules/8/filename/binary.tgz"
                    },
                    "md5sum" : {
                        "href" : "http://127.0.0.1:8080/api/v1/default/download/controller/2/softwaremodules/8/filename/binary.tgz.MD5SUM"
                    },
                    "download-http" : {
                        "href" : "http://127.0.0.1:8080/api/v1/default/download/controller/2/softwaremodules/8/filename/binary.tgz"
                    },
                    "md5sum-http" : {
                        "href" : "http://127.0.0.1:8080/api/v1/default/download/controller/2/softwaremodules/8/filename/binary.tgz.MD5SUM"
                    }
                }
            } ]
        } ],
        "maintenanceWindow": {
            "maintenanceWindowStartTime": "2017-04-11T00:45:00+0000",
            "maintenanceWindowEndTime": "2017-04-11T01:00:00+0000"
        }
    }
}
```

3. The device downloads the artifacts and then waits for the install command from
hawkBit. The device polls near to the maintenance window start time and gets the
updated polling interval.
```
[2017-04-11T02:35:00+0200] $ curl 'http://127.0.0.1:8080/default/controller/v1/2' -i -H 'Accept: application/hal+json'
HTTP/1.1 200 OK
Content-Type: application/hal+json;charset=UTF-8
Content-Length: 356
{
    "config": {
        "polling": {
        "sleep": "00:03:49"
        }
    },
    "_links": {
        "deploymentBase": {
            "href": "http://127.0.0.1:8080/default/controller/v1/2/deploymentBase/1?c=411629670"
        },
        "configData": {
            "href": "http://127.0.0.1:8080/default/controller/v1/2/configData"
       }
    }
}

[2017-04-11T02:39:00+0200] $ curl 'http://127.0.0.1:8080/default/controller/v1/2' -i -H 'Accept: application/hal+json'
HTTP/1.1 200 OK
Content-Type: application/hal+json;charset=UTF-8
Content-Length: 356
{
    "config": {
        "polling": {
        "sleep": "00:02:00"
        }
    },
    "_links": {
        "deploymentBase": {
            "href": "http://127.0.0.1:8080/default/controller/v1/2/deploymentBase/1?c=411629670"
        },
        "configData": {
            "href": "http://127.0.0.1:8080/default/controller/v1/2/configData"
        }
    }
}

[2017-04-11T02:41:00+0200] $ curl 'http://127.0.0.1:8080/default/controller/v1/2' -i -H 'Accept: application/hal+json'
HTTP/1.1 200 OK
Content-Type: application/hal+json;charset=UTF-8
Content-Length: 356
{
    "config": {
        "polling": {
        "sleep": "00:01:39"
        }
    },
    "_links": {
        "deploymentBase": {
            "href": "http://127.0.0.1:8080/default/controller/v1/2/deploymentBase/1?c=411629670"
        },
        "configData": {
            "href": "http://127.0.0.1:8080/default/controller/v1/2/configData"
        }
    }
}
```

4. The polling interval does not go below the minimum polling time.
```
[2017-04-11T02:43:00+0200] $ curl 'http://127.0.0.1:8080/default/controller/v1/2' -i -H 'Accept: application/hal+json'
HTTP/1.1 200 OK
Content-Type: application/hal+json;charset=UTF-8
Content-Length: 356
{
    "config": {
        "polling": {
        "sleep": "00:01:00"
        }
    },
    "_links": {
        "deploymentBase": {
            "href": "http://127.0.0.1:8080/default/controller/v1/2/deploymentBase/1?c=411629670"
        },
        "configData": {
            "href": "http://127.0.0.1:8080/default/controller/v1/2/configData"
        }
    }
}

[2017-04-11T02:44:00+0200] $ curl 'http://127.0.0.1:8080/default/controller/v1/2' -i -H 'Accept: application/hal+json'
HTTP/1.1 200 OK
Content-Type: application/hal+json;charset=UTF-8
Content-Length: 356
{
    "config": {
        "polling": {
        "sleep": "00:01:00"
        }
    },
    "_links": {
        "deploymentBase": {
            "href": "http://127.0.0.1:8080/default/controller/v1/2/deploymentBase/1?c=411629670"
        },
        "configData": {
            "href": "http://127.0.0.1:8080/default/controller/v1/2/configData"
        }
    }
}
```

5. The device polls just after the maintenance window start time and gets the
new ETAG for deploymentBase and the polling time is reset.
```
[2017-04-11T02:45:00+0200] $ curl 'http://127.0.0.1:8080/default/controller/v1/2' -i -H 'Accept: application/hal+json'
HTTP/1.1 200 OK
Content-Type: application/hal+json;charset=UTF-8
Content-Length: 356
{
    "config": {
        "polling": {
        "sleep": "00:05:00"
        }
    },
    "_links": {
        "deploymentBase": {
            "href": "http://127.0.0.1:8080/default/controller/v1/2/deploymentBase/1?c=411659461"
        },
        "configData": {
            "href": "http://127.0.0.1:8080/default/controller/v1/2/configData"
        }
    }
}
```

6. The 'update' status has been changed to 'forced' to signal the start
of installation.
```
[2017-04-11T02:45:30+0200] $ curl 'http://127.0.0.1:8080/default/controller/v1/2/deploymentBase/1?c=411659461' -i -H 'Accept: application/hal+json'
HTTP/1.1 200 OK
Content-Type: application/hal+json;charset=UTF-8
Content-Length: 1665
{
    "id": "1",
    "deployment": {
        "download": "forced",
        "update": "forced",
        "chunks": [ {
            "part": "os",
            "version": "1",
            "name": "1",
            "artifacts" : [ {
                "filename" : "binary.tgz",
                "hashes" : {
                    "sha1" : "8b71973e835cd0718827238b1cf89f0079e44dae",
                    "md5" : "a90218782025c48e09ba0436cf30d029"
                },
                "size" : 5,
                "_links" : {
                    "download" : {
                        "href" : "http://127.0.0.1:8080/api/v1/default/download/controller/2/softwaremodules/8/filename/binary.tgz"
                    },
                    "md5sum" : {
                        "href" : "http://127.0.0.1:8080/api/v1/default/download/controller/2/softwaremodules/8/filename/binary.tgz.MD5SUM"
                    },
                    "download-http" : {
                        "href" : "http://127.0.0.1:8080/api/v1/default/download/controller/2/softwaremodules/8/filename/binary.tgz"
                    },
                    "md5sum-http" : {
                        "href" : "http://127.0.0.1:8080/api/v1/default/download/controller/2/softwaremodules/8/filename/binary.tgz.MD5SUM"
                    }
                }
            } ]
        } ],
        "maintenanceWindow": {
            "maintenanceWindowStartTime": "2017-04-11T00:45:00+0000",
            "maintenanceWindowEndTime": "2017-04-11T01:00:00+0000"
        }
    }
}
```

7. In case the maintenance window has elapsed before the device was able to send the
completed status, the action is cancelled.
```
[2017-04-11T03:05:00+0200] $ curl 'http://127.0.0.1:8080/default/controller/v1/2' -i -H 'Accept: application/hal+json'
HTTP/1.1 200 OK
Content-Type: application/hal+json;charset=UTF-8
Content-Length: 340
{
    "config": {
        "polling": {
        "sleep": "00:05:00"
        }
    },
    "_links": {
        "cancelAction": {
            "href": "http://127.0.0.1:8080/default/controller/v1/2/cancelAction/1"
        },
        "configData": {
            "href": "http://127.0.0.1:8080/default/controller/v1/2/configData"
        }
    }
}
```

### DIFF
--- a/hawkbit-ddi-api/src/main/java/org/eclipse/hawkbit/ddi/json/model/DdiDeployment.java
+++ b/hawkbit-ddi-api/src/main/java/org/eclipse/hawkbit/ddi/json/model/DdiDeployment.java
@@ -12,13 +12,14 @@ import java.util.Collections;
 import java.util.List;
 
 import javax.validation.constraints.NotNull;
-
+import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.annotation.JsonValue;
 
 /**
  * Detailed update action information.
  */
+@JsonInclude(JsonInclude.Include.NON_NULL)
 public class DdiDeployment {
 
     private HandlingType download;
@@ -28,6 +29,8 @@ public class DdiDeployment {
     @JsonProperty("chunks")
     @NotNull
     private List<DdiChunk> chunks;
+
+    private MaintenanceWindowStatus maintenanceWindow = null;
 
     /**
      * Constructor.
@@ -45,11 +48,21 @@ public class DdiDeployment {
      *            handling type
      * @param chunks
      *            to handle.
+     * @param maintenanceWindow
+     *            specifying whether there is a maintenance schedule associated.
+     *            If it is, the value is either 'available' (i.e. the
+     *            maintenance window is now available as per defined schedule
+     *            and the update can progress) or 'unavailable' (implying that
+     *            maintenance window is not available now and update should not
+     *            be attempted). If there is no maintenance schedule defined,
+     *            the parameter is null.
      */
-    public DdiDeployment(final HandlingType download, final HandlingType update, final List<DdiChunk> chunks) {
+    public DdiDeployment(final HandlingType download, final HandlingType update, final List<DdiChunk> chunks,
+            final MaintenanceWindowStatus maintenanceWindow) {
         this.download = download;
         this.update = update;
         this.chunks = chunks;
+        this.maintenanceWindow = maintenanceWindow;
     }
 
     public HandlingType getDownload() {
@@ -66,6 +79,10 @@ public class DdiDeployment {
         }
 
         return Collections.unmodifiableList(chunks);
+    }
+
+    public MaintenanceWindowStatus getMaintenanceWindow() {
+        return this.maintenanceWindow;
     }
 
     /**
@@ -100,9 +117,41 @@ public class DdiDeployment {
         }
     }
 
+    /**
+     * Status of the maintenance window for action.
+     */
+    public enum MaintenanceWindowStatus {
+        /**
+         * A window is currently available, target can go ahead with
+         * installation.
+         */
+        AVAILABLE("available"),
+
+        /**
+         * A window is not available, target should wait and skip the
+         * installation.
+         */
+        UNAVAILABLE("unavailable");
+
+        private String status;
+
+        MaintenanceWindowStatus(final String status) {
+            this.status = status;
+        }
+
+        /**
+         * @return status of maintenance window.
+         */
+        @JsonValue
+        public String getStatus() {
+            return this.status;
+        }
+    }
+
     @Override
     public String toString() {
-        return "Deployment [download=" + download + ", update=" + update + ", chunks=" + chunks + "]";
+        return "Deployment [download=" + download + ", update=" + update + ", chunks=" + chunks
+                + (maintenanceWindow == null ? "]" : (", maintenanceWindow=" + maintenanceWindow + "]"));
     }
 
 }
--- a/hawkbit-ddi-resource/src/main/java/org/eclipse/hawkbit/ddi/rest/resource/DdiRootController.java
+++ b/hawkbit-ddi-resource/src/main/java/org/eclipse/hawkbit/ddi/rest/resource/DdiRootController.java
@@ -29,6 +29,7 @@ import org.eclipse.hawkbit.ddi.json.model.DdiConfigData;
 import org.eclipse.hawkbit.ddi.json.model.DdiControllerBase;
 import org.eclipse.hawkbit.ddi.json.model.DdiDeployment;
 import org.eclipse.hawkbit.ddi.json.model.DdiDeployment.HandlingType;
+import org.eclipse.hawkbit.ddi.json.model.DdiDeployment.MaintenanceWindowStatus;
 import org.eclipse.hawkbit.ddi.json.model.DdiDeploymentBase;
 import org.eclipse.hawkbit.ddi.json.model.DdiResult.FinalResult;
 import org.eclipse.hawkbit.ddi.rest.api.DdiRestConstants;
@@ -41,6 +42,7 @@ import org.eclipse.hawkbit.repository.SystemManagement;
 import org.eclipse.hawkbit.repository.builder.ActionStatusCreate;
 import org.eclipse.hawkbit.repository.event.remote.DownloadProgressEvent;
 import org.eclipse.hawkbit.repository.exception.ArtifactBinaryNotFoundException;
+import org.eclipse.hawkbit.repository.exception.CancelActionNotAllowedException;
 import org.eclipse.hawkbit.repository.exception.EntityNotFoundException;
 import org.eclipse.hawkbit.repository.exception.SoftwareModuleNotAssignedToTargetException;
 import org.eclipse.hawkbit.repository.model.Action;
@@ -140,9 +142,13 @@ public class DdiRootController implements DdiRootControllerRestApi {
 
         final Target target = controllerManagement.findOrRegisterTargetIfItDoesNotexist(controllerId, IpUtil
                 .getClientIpFromRequest(requestResponseContextHolder.getHttpServletRequest(), securityProperties));
-        return new ResponseEntity<>(DataConversionHelper.fromTarget(target,
-                controllerManagement.findOldestActiveActionByTarget(controllerId).orElse(null),
-                controllerManagement.getPollingTime(), tenantAware), HttpStatus.OK);
+        final Action action = controllerManagement.findOldestActiveActionByTarget(controllerId).orElse(null);
+
+        checkAndCancelExpiredAction(action);
+
+        return new ResponseEntity<>(
+                DataConversionHelper.fromTarget(target, action, controllerManagement.getPollingTime(), tenantAware),
+                HttpStatus.OK);
     }
 
     @Override
@@ -266,14 +272,14 @@ public class DdiRootController implements DdiRootControllerRestApi {
             return ResponseEntity.notFound().build();
         }
 
+        checkAndCancelExpiredAction(action);
+
         if (!action.isCancelingOrCanceled()) {
 
             final List<DdiChunk> chunks = DataConversionHelper.createChunks(target, action, artifactUrlHandler,
                     systemManagement,
                     new ServletServerHttpRequest(requestResponseContextHolder.getHttpServletRequest()),
                     controllerManagement);
-
-            final HandlingType handlingType = action.isForce() ? HandlingType.FORCED : HandlingType.ATTEMPT;
 
             final List<String> actionHistoryMsgs = controllerManagement.getActionHistoryMessages(action.getId(),
                     actionHistoryMessageCount == null ? Integer.parseInt(DdiRestConstants.NO_ACTION_HISTORY)
@@ -282,8 +288,17 @@ public class DdiRootController implements DdiRootControllerRestApi {
             final DdiActionHistory actionHistory = actionHistoryMsgs.isEmpty() ? null
                     : new DdiActionHistory(action.getStatus().name(), actionHistoryMsgs);
 
+            final HandlingType downloadType = action.isForce() ? HandlingType.FORCED : HandlingType.ATTEMPT;
+            final HandlingType updateType = action.hasMaintenanceSchedule()
+                    ? (action.isMaintenanceWindowAvailable() ? downloadType : HandlingType.SKIP) : downloadType;
+
+            MaintenanceWindowStatus maintenanceWindow = action.hasMaintenanceSchedule()
+                    ? (action.isMaintenanceWindowAvailable() ? MaintenanceWindowStatus.AVAILABLE
+                            : MaintenanceWindowStatus.UNAVAILABLE)
+                    : null;
+
             final DdiDeploymentBase base = new DdiDeploymentBase(Long.toString(action.getId()),
-                    new DdiDeployment(handlingType, handlingType, chunks, null), actionHistory);
+                    new DdiDeployment(downloadType, updateType, chunks, maintenanceWindow), actionHistory);
 
             LOG.debug("Found an active UpdateAction for target {}. returning deyploment: {}", controllerId, base);
 
@@ -350,6 +365,13 @@ public class DdiRootController implements DdiRootControllerRestApi {
             break;
         case CLOSED:
             status = handleClosedCase(feedback, controllerId, actionid, messages);
+            break;
+        case DOWNLOADED:
+            LOG.debug(
+                    "Controller confirmed download of distribution set (actionId: {}, controllerId: {}) as we got {} report.",
+                    actionid, controllerId, feedback.getStatus().getExecution());
+            status = Status.DOWNLOADED;
+            messages.add(RepositoryConstants.SERVER_MESSAGE_PREFIX + "Target confirmed download of distribution set.");
             break;
         default:
             status = handleDefaultCase(feedback, controllerId, actionid, messages);
@@ -513,5 +535,22 @@ public class DdiRootController implements DdiRootControllerRestApi {
     private Action findActionWithExceptionIfNotFound(final Long actionId) {
         return controllerManagement.findActionWithDetails(actionId)
                 .orElseThrow(() -> new EntityNotFoundException(Action.class, actionId));
+    }
+
+    /**
+     * If the action has a maintenance schedule defined but is no longer valid,
+     * cancel the action.
+     *
+     * @param action
+     *            is the {@link Action} to check.
+     */
+    private void checkAndCancelExpiredAction(final Action action) {
+        if (action != null && action.hasMaintenanceSchedule() && action.isMaintenanceScheduleLapsed()) {
+            try {
+                controllerManagement.cancelAction(action.getId());
+            } catch (final CancelActionNotAllowedException e) {
+                LOG.info("Cancel action not allowed exception :{}", e);
+            }
+        }
     }
 }
--- a/hawkbit-ddi-resource/src/main/java/org/eclipse/hawkbit/ddi/rest/resource/DdiRootController.java
+++ b/hawkbit-ddi-resource/src/main/java/org/eclipse/hawkbit/ddi/rest/resource/DdiRootController.java
@@ -146,9 +146,8 @@ public class DdiRootController implements DdiRootControllerRestApi {
 
         checkAndCancelExpiredAction(action);
 
-        return new ResponseEntity<>(
-                DataConversionHelper.fromTarget(target, action, controllerManagement.getPollingTime(), tenantAware),
-                HttpStatus.OK);
+        return new ResponseEntity<>(DataConversionHelper.fromTarget(target, action,
+                controllerManagement.getPollingTimeForAction(action), tenantAware), HttpStatus.OK);
     }
 
     @Override
--- a/hawkbit-ddi-resource/src/main/java/org/eclipse/hawkbit/ddi/rest/resource/DdiRootController.java
+++ b/hawkbit-ddi-resource/src/main/java/org/eclipse/hawkbit/ddi/rest/resource/DdiRootController.java
@@ -283,7 +283,7 @@ public class DdiRootController implements DdiRootControllerRestApi {
                     : new DdiActionHistory(action.getStatus().name(), actionHistoryMsgs);
 
             final DdiDeploymentBase base = new DdiDeploymentBase(Long.toString(action.getId()),
-                    new DdiDeployment(handlingType, handlingType, chunks), actionHistory);
+                    new DdiDeployment(handlingType, handlingType, chunks, null), actionHistory);
 
             LOG.debug("Found an active UpdateAction for target {}. returning deyploment: {}", controllerId, base);
 
--- a/hawkbit-repository/hawkbit-repository-api/src/main/java/org/eclipse/hawkbit/repository/model/Action.java
+++ b/hawkbit-repository/hawkbit-repository-api/src/main/java/org/eclipse/hawkbit/repository/model/Action.java
@@ -16,6 +16,21 @@ import java.util.concurrent.TimeUnit;
 public interface Action extends TenantAwareBaseEntity {
 
     /**
+     * Maximum length of controllerId.
+     */
+    int MAINTENANCE_SCHEDULE_CRON_LENGTH = 128;
+
+    /**
+     * Maximum length of controllerId.
+     */
+    int MAINTENANCE_WINDOW_DURATION_LENGTH = 16;
+
+    /**
+     * Maximum length of controllerId.
+     */
+    int MAINTENANCE_WINDOW_TIMEZONE_LENGTH = 8;
+
+    /**
      * @return the distributionSet
      */
     DistributionSet getDistributionSet();
@@ -168,7 +183,13 @@ public interface Action extends TenantAwareBaseEntity {
         /**
          * Cancellation has been rejected by the controller.
          */
-        CANCEL_REJECTED;
+        CANCEL_REJECTED,
+
+        /**
+         * Action has been downloaded by the target and waiting for update to
+         * start.
+         */
+        DOWNLOADED;
     }
 
     /**
@@ -193,4 +214,36 @@ public interface Action extends TenantAwareBaseEntity {
          */
         TIMEFORCED;
     }
+
+    /**
+     * The method checks whether the action has a maintenance schedule defined
+     * for it. A maintenance schedule defines a set of maintenance windows
+     * during which actual update can be performed. A valid schedule defines at
+     * least one maintenance window.
+     *
+     * @return true if action has a maintenance schedule, else false.
+     */
+    boolean hasMaintenanceSchedule();
+
+    /**
+     * The method checks whether the maintenance schedule has already lapsed for
+     * the action, i.e. there are no more windows available for maintenance.
+     * Controller manager uses the method to check if the maintenance schedule
+     * has lapsed, and automatically cancels the action if it is lapsed.
+     *
+     * @return true if maintenance schedule has lapsed, else false.
+     */
+    boolean isMaintenanceScheduleLapsed();
+
+    /**
+     * The method checks whether a maintenance window is available for the
+     * action to proceed. If it is available, a 'true' value is returned. The
+     * maintenance window is considered available: 1) If there is no maintenance
+     * schedule at all, in which case device can start update any time after
+     * download is finished; or 2) the current time is within a scheduled
+     * maintenance window start and end time.
+     *
+     * @return true if maintenance window is available, else false.
+     */
+    boolean isMaintenanceWindowAvailable();
 }
--- a/hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaControllerManagement.java
+++ b/hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaControllerManagement.java
@@ -9,7 +9,15 @@
 package org.eclipse.hawkbit.repository.jpa;
 
 import java.net.URI;
+import java.time.Duration;
+import java.time.ZonedDateTime;
+import java.time.temporal.ChronoUnit;
+import java.time.temporal.TemporalUnit;
 import java.util.Collection;
+import java.time.Duration;
+import java.time.ZonedDateTime;
+import java.time.temporal.ChronoUnit;
+import java.time.temporal.TemporalUnit;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
@@ -29,6 +37,7 @@ import javax.persistence.criteria.Root;
 
 import org.eclipse.hawkbit.repository.ControllerManagement;
 import org.eclipse.hawkbit.repository.EntityFactory;
+import org.eclipse.hawkbit.repository.MaintenanceScheduleHelper;
 import org.eclipse.hawkbit.repository.QuotaManagement;
 import org.eclipse.hawkbit.repository.RepositoryConstants;
 import org.eclipse.hawkbit.repository.RepositoryProperties;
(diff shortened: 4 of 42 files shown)


kaizimmerm:
First of all thank you for this pull request 👍 

There are however a few issues :)
1) I will definitely need some time to look through all of this. I think I understand what you are going for but I don't see all the consequences yet of introducing that maintenance window function (e.g. on Rollout Management and DMF).
2) I definitely need @michahirsch 's opinion on this. he will be back in office next week.
3) This goes beyond 1000k lines of code. As you guys are not committers this has to go through the Eclipse IP process.

singh-himanshu:
Just re-based and pushed the changes again as there were some conflicts with recent commits on master.
All the tests pass locally after doing a clean build, but on CI there is a test failure. All tests for hawbit-repositoty-jpa pass except one:
ControllerManagementTest.tryToFinishWithErrorUpdateProcessMoreThanOnce Did not receive the expected amount of events form class org.eclipse.hawkbit.repository.event.remote.entity.TargetCreatedEvent Expected: 1 but was: 0

Any thoughts on what could be the reason?

michahirsch:
Thanks a lot for the PR 👍 . First look I was thinking `maintaince-window` what? :) But looking through the PR it makes more and more sense about the approach. I still have some comments on the concept of the maintenance-window, here are my points/thoughts on it. Really would like also to hear your background and thoughts on this.

* Changing the polling-interval dynamically into the start of the maintenance window (as I understand will the `EventTimer` caluclate) can be problematic because you are forcing all clients to poll at the exact same time (start of maintenance-window), so all clients will poll at the exact same time at once. Actually in case the polltime-interval is less than the maintenance-window you don't have to caluclate another poll-time because the clients will poll somwhere in the maintenance-window then. 

  Nevertheless I'm still not sure if hawkBit should change the poll-interval dynamically because we would need to spread the load of the devices between the maintenance-window to avoid polling at the same time. The only other approach I can come up with currently, is to just tell the device the maintenance-window so the device needs to remember the maintenance-window and needs to care automatically the update and on further polls hawkBit just tells the next-maintenance-window (see below repeatable maintenance-window). I'm not sure if this is feasible for a client to remember the next-maintenance-window and install it automatically.

  Also for the DMF-interface hawkBit would need a scheduler which needs to caluclate if devices are currently in a maintenance-window and then send a push-event notification. DMF-interface does not need to be included in this PR (can also be done later) but just need to fit into the concept :)

  It is a nicer aproach if hawkBit takes care to tell the devices the maintenance-window and to tell the device that it is allowed to update now, just difficult with the polling approach :). So keeping changing the polling-time dynamically is not a bad idea but maybe we need some spreading of the poll-time so not all device poll at the same time and in case the poll-interval is shorter than the maintenance-window we don't need to adjust the polling-time at all? 

* An fixed `one-time` maintenance-window with a hard end could be problematic e.g. in case of an large rollout. An repeatable maintenance-window e.g. every night between 2am - 3am would be great. So a kind of cron-pattern as a maintenance-window would be good. Maybe hawkBit could persist only the cron-pattern (e.g. `0 0 23 1/1 * ? *` - Every day at 23:00 o'clock) and the length of the maintenance-window (e.g. 2hours) and then hawkBit could caluclate the next-maintenance-window by it's own. Spring has utility classes to support cron (see `org.springframework.scheduling.support.CronSequenceGenerator`) so no new library is needed I guess.

* Action status `Downloaded` is just used for information. An `update: forced` is sent anyway in the maintaince window, even software is not downloaded yet, in purposes? Should an update software request only sent to the device in case it reported the `downloaded` state? What happens in case of message loss? That would be ugly.

* Client might retrieve a cancel request from hawkBit if the client polls during updating software and maintenance-window is elapsed, hawkBit sends automatically a cancel request back. Not sure if clients keep polling during a software update (maybe to a certain point it's possible) they would retrieve then an cancel-request because the maintenance-window is elapsed. Does hawkBit needs to take care of it? Then we would need an "Installing" status type so hawkBit is aware that the client is currently installing.
   

There are also some technical things which might be adapted but I guess it would be better to clearify the conceptional things first.

singh-himanshu:
Thanks @michahirsch for your comments, here are some additional clarifications:

**Dynamic polling interval change**: Our initial thought were exactly that, just tell the client what is the start time and put the onus on client to start the update based on maintenance window start time. But this approach is dependent on client side implementation which has couple of limitations: first, all client software need to change to make use of this feature, second client running on constrained hardware may not have additional timer available to schedule the update. Our thought is to go with your suggestion for this PR, and have an optional dynamic polling time later. Currently we implemented the dynamic change in polling interval solely based on whether the latest action for the controller has a maintenance window or not. We may make this behaviour configurable either per action or per tenant. So tenants that have an updated client software or more capable hardware may not want to use dynamic polling time.

**DMF**: For DMF, currently the behaviour still relies on a poll like behaviour from client implementations. In which case server will check and send a message when it receives a message from client on UPDATE_ACTION_STATUS topic (I think I missed to modify the check before call to lookIfUpdateAvailabe in AmqpMessageHandlerService). We may keep this behaviour for now and later introduce a scheduler based handling.

**Rollouts and cron like specification**: We viewed roll-out as a higher level abstraction over the core per device update aspects. Similar feature may also be provided by a third-party fleet manager. Hence cron-like specification of maintenance window was left out. This may also be done as part of the core update feature, but we propose to have this done as a separate PR and within Rollout Management. For now the hawkBit rollout management behaves as if not supporting maintenance window.

**Action Status**: You are right that the DOWNLOADED status is just informational, set based on status sent by client, to keep track on server side; _update=forced_ will be sent regardless. Whereas change of status to START_UPDATE by hawkBit is used to force the ETAG change prompting client to retrieve deployment base information again, where it will find _update=forced_.

**Cancellation**: Our expectation is not different than regular cancellation, that is even if hawkBit cancels the action based on maintenance window, client needs to acknowledge.

My understanding based on the comments so far is that we _drop the dynamic polling time change_ from the current PR. We create _additional PRs later_ to **(a)** introduce a scheduler based mechanism for DMF **(b)** support cron like specification for maintenance window to support roll outs **(c)** have an optional (per action or per tenant) dynamic polling time. What are you thoughts?

sbabic:
Hi Michael,

On 16/06/2017 09:31, Michael Hirsch wrote:
> Thanks a lot for the PR 👍 . First look I was thinking
> |maintaince-window| what? :) But looking through the PR it makes more
> and more sense about the approach. I still have some comments on the
> concept of the maintenance-window, here are my points/thoughts on it.
> Really would like also to hear your background and thoughts on this.
> 
>   *
> 
>     Changing the polling-interval dynamically into the start of the
>     maintenance window (as I understand will the |EventTimer| caluclate)
>     can be problematic because you are forcing all clients to poll at
>     the exact same time (start of maintenance-window), so all clients
>     will poll at the exact same time at once. Actually in case the
>     polltime-interval is less than the maintenance-window you don't have
>     to caluclate another poll-time because the clients will poll
>     somwhere in the maintenance-window then.

Right - anyway, how the absolute time is propagated to the devices /
clients ? Up now there is no need to synchronize the time on the Hawkbit
server with the time on the devices. Some devices are not aware of the
absolute time, that means they have neither a RTC nor they are connected
to a NTP server. With a relative time as now (polling time), there is no
need at all. But in the maintenance window, an exact time is set as
"absolute" time, and this requires a synchronization between server and
clients. If a device has no time at all (but still Linux), I assume the
time is 1/1/1970 when device starts.....

Adding a dynamical polling-interval means also that devices *must*
follow it. But the current "polling" interval is currently a
"nice-to-follow" rule, and it is not a must. For example, a device can
be stored for months before to be sold, and it is high desired that the
device updates itself when it is firstly started to the current release.
I do this changing the polling interval on myself, independently what
Hawkbit has sent back - at the start up, device polls more often as
rulled by "polling", and when there is no updates available, the polling
interval provided by Hawkbit is set. The server does not comply if a
device polls more often as specified and this "hidden" (unwanted ?)
feature can be profitable used.

Anyway, I guess that an overload can always happen, for example after a
power-cut with a large number of installed devices. More or less, all
devices announce themselves quite at the same time - do we need to add
some randomness on the devices ?

> 
>     Nevertheless I'm still not sure if hawkBit should change the
>     poll-interval dynamically because we would need to spread the load
>     of the devices between the maintenance-window to avoid polling at
>     the same time.

Well, strictly speaking, I do not know how you can forbid this. You can
just limit the traffic to the server to avoid overloads.

> The only other approach I can come up with currently,
>     is to just tell the device the maintenance-window so the device
>     needs to remember the maintenance-window and needs to care
>     automatically the update and on further polls hawkBit just tells the
>     next-maintenance-window

But again, this imply that device is synchronized (RTC, NTP, ..) with
hawkbit. Or hawkbit sends a relative time (in seconds) from now to the
maintenance window, but this time must be updated after each polling.

> (see below repeatable maintenance-window).
>     I'm not sure if this is feasible for a client to remember the
>     next-maintenance-window and install it automatically.

The question is if the client should introduce some states. In most
cases, the device is stateless, that means after a power cut the whole
procedure is started again. And what happens if the device was offline
in the whole maintenance-window ?

> 
>     Also for the DMF-interface hawkBit would need a scheduler which
>     needs to caluclate if devices are currently in a maintenance-window
>     and then send a push-event notification. DMF-interface does not need
>     to be included in this PR (can also be done later) but just need to
>     fit into the concept :)
> 
>     It is a nicer aproach if hawkBit takes care to tell the devices the
>     maintenance-window and to tell the device that it is allowed to
>     update now, just difficult with the polling approach :). So keeping
>     changing the polling-time dynamically is not a bad idea but maybe we
>     need some spreading of the poll-time so not all device poll at the
>     same time and in case the poll-interval is shorter than the
>     maintenance-window we don't need to adjust the polling-time at all?
> 
>   *
> 
>     An fixed |one-time| maintenance-window with a hard end could be
>     problematic e.g. in case of an large rollout. An repeatable
>     maintenance-window e.g. every night between 2am - 3am would be
>     great. So a kind of cron-pattern as a maintenance-window would be
>     good. Maybe hawkBit could persist only the cron-pattern (e.g. |0 0
>     23 1/1 * ? *| - Every day at 23:00 o'clock) and the length of the
>     maintenance-window (e.g. 2hours) and then hawkBit could caluclate
>     the next-maintenance-window by it's own. Spring has utility classes
>     to support cron (see
>     |org.springframework.scheduling.support.CronSequenceGenerator|) so
>     no new library is needed I guess.
> 
>   *
> 
>     Action status |Downloaded| is just used for information. An |update:
>     forced| is sent anyway in the maintaince window, even software is
>     not downloaded yet, in purposes? Should an update software request
>     only sent to the device in case it reported the |downloaded| state?
>     What happens in case of message loss? That would be ugly.

Right - and what happens if the update on Hawkbit is changed from
"attempted" to "force" ? As response to this transition, a device should
install the new software, without regarding the window (else what is the
meaning of "force" ?)

> 
>   *
> 
>     Client might retrieve a cancel request from hawkBit if the client
>     polls during updating software and maintenance-window is elapsed,
>     hawkBit sends automatically a cancel request back.

ok

> Not sure if
>     clients keep polling during a software update (maybe to a certain
>     point it's possible)

I confirm that this is done, but it has effect only if the time to
update is greater as the polling time. In most cases, it is shorter. But
it is used in case of low-bandwidth connection (GSM for example).

> they would retrieve then an cancel-request
>     because the maintenance-window is elapsed. Does hawkBit needs to
>     take care of it? Then we would need an "Installing" status type so
>     hawkBit is aware that the client is currently installing.

I do not see the reason why Hawkbit should track this - the device must
check if a cancel-request is present and confirm it with a
cancel-feedback. The update is not available and not valid anymore
(Hawkbit is the "master" in the transaction), and the device should not
activate it.

> 
> There are also some technical things which might be adapted but I guess
> it would be better to clearify the conceptional things first.
> 

Best regards,
Stefano


-- 
=====================================================================
DENX Software Engineering GmbH,      Managing Director: Wolfgang Denk
HRB 165235 Munich, Office: Kirchenstr.5, D-82194 Groebenzell, Germany
Phone: +49-8142-66989-53 Fax: +49-8142-66989-80 Email: sbabic@denx.de
=====================================================================

michahirsch:
Thanks a lot for the comments

*dynamic polling interval* 
I agree with @sbabic about the absolute timestamps. For clients it would be much easier it the client retrieves a relative timestamp based on the its poll so timestamps must no be synchronized. I guess the device actually does not care about the `maintenance-window` because using the dynamic-polling-interval the device just needs to follow the poll and hawkBit tells him to download or install the software. So the `maintenanceWindow` does not need to be included in the response of a device poll at all, right?.

Devices poll all at the same time, yes @sbabic hawkBit needs to handle load of many devices polling at the same time, even a power-cut would probabbly not causing all device forcing to poll at the same time :) So introducing the dynamic-polling for all devices in your system having the same maintaince-window you are doing extra high peak load on your system. 

*repeatable maintenance-window* 
I think this featutre is crucial due what happens if devices are power-lost during the maintaince window, or not connected on the internet. Then the action gets automatically canceled and the device missed the update.

*So what I'm thinking is*

* keeping the dynamic-polling feature because it allows to be very flexible due it does not break or enhance the API. It's ment to tell the device when it should ask the hawkBit server again for updates and why not forcing the device to poll in the server-side defined maintaince-window, actually not bad. And if a device is not following the server-side specific polling-interval it maybe misses the maintaince-window or if it ask more frequent it will maybe poll in the maintance-window. Currently hawkBit does not enforce the minimum-polling interval as @sbabic mentioned.

* Can we remove the maintenance-window from the DDI response? The client does not need to know the maintenance-window and we get rid of absolute timestamps.

* Introducing a repeatble maintanence-window right away? My thoughts on this is that it will be more difficult to introduce later on, because then hawkBit needs to support both and maybe migrate some data. But this a just thoughts, maybe it's an easier way to start and later another PR enables a repeatable maintanence-window

* DMF event scheduler can be added later in a PR, totally agree.

How sounds this? @singh-himanshu @sbabic 

I will play around with the branch on this and play with some simulated devices etc.

singh-himanshu:
We can go with the repeatable window with dynamic polling, even though the repeatability could also be achieved using a higher level fleet manager that uses hawkBit.

But we do suggest splitting the PR to keep them small(er), also we think they are fairly decoupled in the sense that regular polling is sufficient if the maintenance window is reasonably longer than polling time. 
- set of commits for creating repeatable maintenance windows
- set of commits to add dynamic polling

Going with repeatable window, there is a need to "pause" the action. Currently we would just cancel the action once the window elapses and leave it for the client to acknowledge. There are I think a couple of ways to inform the client to "pause" the action: 
- One is to use download and update statuses within currently implemented scheme
-- switch the action status back to `RETRIEVED` or `RUNNING` from `START_UPDATE`;
-- client receives "`download=force, update=skip`" as current window has lapsed. one can optimize  more depending on if client has already sent the `DOWNLOADED` status, then client will receive "`download=skip, update=skip`" until the start of next window;
-- client receive either "`download=force, update=force`" or "`download=skip, update=force`" within the next window;
-- if client had already started install before getting a "pause", it may continue to complete the action without acknowledging the "pause".

- Another mechanism could be to introduce an entirely new API (or modify the cancellation behaviour) to pause the action, but we need to figure out a consistent semantic of such API.

We initially did not want a status `START_UPDATE` just to affect `ETAG` update, and thought of using the `calculateEtag` in `DataConversionHelper`, but did not quite get the semantics of:
`result = 31 * action.hashCode() + (action.isHitAutoForceTime(System.currentTimeMillis()) ? 1231 : 1237);`
to me choice of `31`, `1231` and `1237` were not clear.

Another point is that instead of dropping the `maintanenaceWindow` information for client altogether, does it make sense to make it a boolean `true/false` or tri-state `none/fixed/repeat`; this way client can know to expect changes in `"download=* and update=*"` and handle "pause" events.

singh-himanshu:
@michahirsch, @sbabic, @stormc thinking about how to specify the maintenance schedule using the management API, we may need time zone information for the management client as well. So for the management client to specify maintenance window using a cron pattern, ISO 8601 compatible duration, and time zone we will have something like:

```
"maintenanceWindow": {
    "schedule": "0 0 12 ? * WED",
    "duration": "P3Y6M4DT12H30M5S",
    "timezone": "+02:00"
}
```
Since install and activate should ideally take order of few hours at max, rather than order of days, we can specify duration in hh:mm:ss format as well. But then should we be restricting it to 23:59:59 or still allow specification like for example 120:00:00 to specify 5 days.

singh-himanshu:
>On 16/06/2017 09:31, @michahirsch wrote:
>Spring has utility classes to support cron (see org.springframework.scheduling.support.CronSequenceGenerator) so no new library is needed I guess.

Since `CronSequenceGenerator` is based on `java.util.Date` (which is rather cubersome to work with when you have date and duration manipulation) I looked at few other options:
* Quartz: http://www.quartz-scheduler.org: This also tries to support Java 6 and above, hence based on java.util rather than java.time.
* Cron-Parser: http://cron-parser.com/: is a simply a parser and not a complete task scheduler. Version 5.0.5 uses Java 8, but in 6.0.x they introduced a back port to also support Java 7 using `threetenbp` rather than java.time. But this may go away in favour of Java 8. 

@michahirsch, what are your thoughts here. I think choice currently is to either use version 5.0.5 of cron-parser since it will allow direct use of `java.time` or work with org.springframework.scheduling.support.CronSequenceGenerator with slightly more wrapper code (an additional advantage perhaps will be easier support for DMF scheduler).

michahirsch:
@singh-himanshu thanks a lot for all your thoughts.

**pausing action** I don't see the need of pausing an action? I think hawkBit just needs to calculate like the *forceTime* if the current poll device is in a maintaince window or not. hawkBit should be easily calculate this also with a cron-pattern. So you can just adapt the hashCode like *force-time* is doing it `(action.isHitAutoForceTime(System.currentTimeMillis())`. This will then cause the change of the eTag.

So my thoughts are little bit trivial on this. In case a device is not running in the maintaince-window it retrieves the response like `download=force, update=skip` and in case of a poll in the maintaince window you just return `download=force, update=force` so the device actually does not need to know the maintenance-window because hawkBit drives him with the *dynamic-polling* into the maintenance-window and then hawkBit calculates the the maintenance-window and return the device now to install.

**management-api** yes for the management API we would need to the maintenance-window payload when assigning *distribution-set* to a device.  
Not sure if we need a `duration-pattern` or is sufficient to just using milliseconds to define the length of the maintenance-window? The start is defined by the cron-pattern and you just need the information how long the maintenance-window is open. I do think that we would probably need the `timezone` information though.

**cron-library** @kaizimmerm and myself would prefer avoiding introducing new libraries in hawkBit when ever possible. So in case it is possible with not too much overhead to use the `CronSequenceGenerator` it would be nicer otherwise I also stumpled over the http://cron-parser.com/ utility and this looked very promising, so this would be the plan B in my opinion :)

Thanks a lot 👍

singh-himanshu:
I am done with incorporating comments regarding repeatable window and need to do some across time zone testing.

>On 21/06/2017 08:39, @michahirsch wrote:
>..
>cron-library @kaizimmerm and myself would prefer avoiding introducing new libraries in hawkBit when ever possible. So in case it is possible with not too much overhead to use the CronSequenceGenerator it would be nicer otherwise I also stumpled over the http://cron-parser.com/ utility and this looked very promising, so this would be the plan B in my opinion..
>

I tried both the libraries, and there is not much difference in using either of them. With `CronSequenceGenerator` there is a constraint that we cannot specify the year, hence not possible to define a one time or closed schedule with just one expression. Hence the action will not auto expire. Another thing is that the exception message is not very informative in case invalid schedule is provided.
`CronParser` provides very informative message regarding which field of the expression may be a problem. It also accepts an optional field to specify the year or year range. This makes it possible to provide a closed schedule.

Let me know of your opinion based on the library choice based on above.

michahirsch:
Hi @singh-himanshu,

sorry for the late response! Thanks a lot for your time. I guess the library should be ok but I'll double check it and need to check the CQ on Eclipse side.

I'll go through the code and will test it locally. Just give me a little bit time on this.

👍  Cheers
Michael

singh-himanshu:
Hello @michahirsch, it will be some time before we can update the PR with the changes based on cron parser, even though implementation is almost done. The delay is mainly due to some process hiccups and unavailability of effort on our side over next 3 to 4 weeks.

michahirsch:
hi @singh-himanshu, thanks for the hint. Looking forward to seeing your changes regarding the cron-maintenance-window

kaizimmerm:
@stormc @singh-himanshu is there a chance that you guys will continue to work on this? We are still highly interested.

stormc:
Hi @kaizimmerm, yes, we certainly do! Unfortunately, we were distracted by other things but now we got some time for it. A final (internal) review is pending and then we'll submit a new version.

stormc:
@kaizimmerm: Just pushed the refreshed and reviewed version of 2018, which is why the license check fails as it checks for `Copyright (c) Siemens AG, 2017`.
Should I put a commit to fix this on top or how do you want (me) to address this?

kaizimmerm:
@stormc this header check is really annoying. But I will fix this soon #634 .

kaizimmerm:
@stormc should be fixed on all masters now (core, extensions,examples).

hawkbit-ddi-resource/src/main/java/org/eclipse/hawkbit/ddi/rest/resource/DdiRootController.java
```diff
@@ -266,24 +271,33 @@ private static boolean checkModule(final String fileName, final SoftwareModule m
...
 
             final DdiActionHistory actionHistory = actionHistoryMsgs.isEmpty() ? null
                     : new DdiActionHistory(action.getStatus().name(), actionHistoryMsgs);
 
+            final HandlingType downloadType = action.isForce() ? HandlingType.FORCED : HandlingType.ATTEMPT;
+            final HandlingType updateType = action.hasMaintenanceSchedule()
+                    ? (action.isMaintenanceWindowAvailable() ? downloadType : HandlingType.SKIP) : downloadType;
+
+            MaintenanceWindowStatus maintenanceWindow = action.hasMaintenanceSchedule()
+                    ? (action.isMaintenanceWindowAvailable() ? MaintenanceWindowStatus.AVAILABLE
```
michahirsch: I'd use `MaintenanceWindowStatus.AVAILABLE.getStatus` to retrieve the lower case name of `available` and `unavailable` in the json-representation. Currently it's upper-case
```
{
    "id": "4",
    "deployment": {
        "download": "forced",
        "update": "forced",
        "maintenanceWindow": "AVAILABLE",
        "chunks": []
}
```

hawkbit-ddi-api/src/main/java/org/eclipse/hawkbit/ddi/json/model/DdiDeployment.java
```diff
@@ -100,9 +117,40 @@ public String getName() {
         }
     }
 
+    /**
+     * Status of the maintenance window for action.
+     */
+    public enum MaintenanceWindowStatus {
```
michahirsch: Is the maintenance status necessary in the deployment base result? 
```
"deployment": {
        "download": "forced",
        "update": "forced",
        "maintenanceWindow": "AVAILABLE"
.....
......
```
Wouldn't be sufficient the `"update" : "skip"` and `"update" : "forced"` indicate the update should be installed? is there an advanatge or use-case to cover the extra `maintenanceWindow` indicator?

hawkbit-repository/hawkbit-repository-api/src/main/java/org/eclipse/hawkbit/repository/model/Action.java
```diff
@@ -168,7 +168,13 @@ default boolean isForced() {
...
          * Cancellation has been rejected by the controller.
          */
-        CANCEL_REJECTED;
+        CANCEL_REJECTED,
+
+        /**
+         * Action has been downloaded by the target and waiting for update to
+         * start.
+         */
+        DOWNLOADED;
```
michahirsch: You need to map the enum in the `JpaActionStatus` and `JpaAction` in the `@ObjectTypeConverter`. Otherwise you retrieve an exception when using it to give a `downloaded` type feedback. 
`No conversion value provided for the attribute [DOWNLOADED]`
```
@ObjectTypeConverter(name = "status", objectType = Action.Status.class, dataType = Integer.class, conversionValues = {
            @ConversionValue(objectValue = "FINISHED", dataValue = "0"),
            @ConversionValue(objectValue = "ERROR", dataValue = "1"),
            @ConversionValue(objectValue = "WARNING", dataValue = "2"),
            @ConversionValue(objectValue = "RUNNING", dataValue = "3"),
            @ConversionValue(objectValue = "CANCELED", dataValue = "4"),
            @ConversionValue(objectValue = "CANCELING", dataValue = "5"),
            @ConversionValue(objectValue = "RETRIEVED", dataValue = "6"),
            @ConversionValue(objectValue = "DOWNLOAD", dataValue = "7"),
            @ConversionValue(objectValue = "SCHEDULED", dataValue = "8"),
            @ConversionValue(objectValue = "CANCEL_REJECTED", dataValue = "9"),
            @ConversionValue(objectValue = "DOWNLOADED", dataValue = "10") })
```

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaControllerManagement.java
```diff
@@ -175,6 +186,142 @@ public String getPollingTime() {
...
+    /**
+     * EventTimer to handle reduction of polling interval based on maintenance
+     * window start time. Class models the next polling time as an event to be
+     * raised and time to next polling as a timer. The event, in this case the
+     * polling, should happen when timer expires. Class makes use of java.time
+     * package to manipulate and calculate timer duration.
+     */
+    private class EventTimer {
+
+        String defaultEventInterval;
```
michahirsch: fields can be private and final because only accessed and set within constructor, rigth?

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/model/JpaAction.java
```diff
@@ -112,6 +117,20 @@
     @JoinColumn(name = "rollout", updatable = false, foreignKey = @ForeignKey(value = ConstraintMode.CONSTRAINT, name = "fk_action_rollout"))
     private JpaRollout rollout;
 
+    @Column(name = "maintenance_cron_schedule")
+    private String maintenanceSchedule;
```
michahirsch: Can you add `javax.validation` to the fields like min or max length? You can checkout the constraints and their definition like on the JpaTarget how it's used and for documentation purposes the constraints are public available in the API of the entities like `Target#ADDRESS_MAX_SIZE`

michahirsch:
@stormc  

following steps:

* minor improvements on pull-request
* committers needs to open a CQ for the additional added library - there is already am approval for the `cron-utils` in the specific version ( so piggyback CQ is sufficient like https://dev.eclipse.org/ipzilla/show_bug.cgi?id=12858 ). 
   ```
<groupId>com.cronutils</groupId>
<artifactId>cron-utils</artifactId>
<version>5.0.5</version>
   ```
* Due this contribution contains more than 1,000 lines we need to apply the `Everything Else` workflow on the Eclipse-Legal-Process https://www.eclipse.org/legal/EclipseLegalProcessPoster.pdf - was thinking to split the code semantically by core and UI extension etc but the pull-request won't be less than 1,000 lines anyway, I guess.

hawkbit-ddi-api/src/main/java/org/eclipse/hawkbit/ddi/json/model/DdiDeployment.java
singh-himanshu: hmm... i guess this will depend on the client side agent... since we do not pass the information about the exact time of the window and duration, not sure what client can do with this extra piece of information.... as of now it is serving more of an indicator that this particular update was created with a maintenance schedule..

singh-himanshu:
here is some stats on volume of changes:
core changes: ~900 line changes
+ test cases: ~1200 line changes
+ UI ~1550: line changes
+ DMF ~1650: line changes
+ dynamic update to polling: ~1950 line changes

So I guess, we are looking at  minimum 1200 lines including the test cases.

singh-himanshu:
I have incorporated the code comments. Will revise the merge request soon.

michahirsch:
I've filled out the Contribution Questionnaire and opened the CQ for the contribution
see: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=15759

kaizimmerm:
And here is the CQ for the cron-utils https://dev.eclipse.org/ipzilla/show_bug.cgi?id=15762

stormc:
Thanks guys for stepping forward here! As I'm not that familiar with the process we have to go though, is there something I should do or can help you with? Just drop me a note...

michahirsch:
@stormc  no worries. Unfortunately you cannot help here. @kaizimmerm opened the CQ for piggybacking the `corn-utils` dependency which is actually already approved.

We already got a +1 from PMC for the contribution but now we need the approval and the checks from the Eclipse Foundation,. So currently we need to be a little bit patient.

Cheers,
Michael

kaizimmerm:
CQs where approved. Thanks for the contribution.

stormc:
Thanks @kaizimmerm, @michahirsch for considering this contribution and your patience! :)

PR #540, opened 2017-06-13
Refactor error message if software module assignment to distribution fails

melanieretter:
When assigning a software module to a distribution, the type of the software module is validated. Only software modules of the allowed types can be assigned to a distribution, The allowed types are set in the distribution type. If an assignment fails due to the wrong type of software module, an error message is displayed.
For better usability the error message was refactored. Now the error message tells the user what types of software modules are allowed, and what type is the software module he has tried to assign.

refactor_error_message

Furthermore there were two small changes in error message wording, to improve the understanding. 


hawkbit-ui/src/main/resources/messages.properties
```diff
@@ -402,9 +402,9 @@ label.drop.dist.delete.area = Drop here<br>to delete
...
 caption.assign.software.dist.accordion.tab = Assign Software Modules
 message.software.assignment = {0} Software Module(s) Assignment(s) done
-message.dist.inuse = {0} Distribution is already assigned to target
-message.software.dist.already.assigned = {0} Distribution already has Software Module {1}
-message.software.dist.type.notallowed = {0} Software Module type can not assign to Distribution {1}
+message.dist.inuse = Distribution {0} is already assigned to target
+message.software.dist.already.assigned = Software Module {0} is already assigned to Distribution {1}
+message.software.dist.type.notallowed = Software Module {0} cannot be assigned to Distribution {1}, because Software Module has to be of type {2}, but is of type {3}
 message.target.assigned =  {0} is assigned to {1}
 message.dist.type.delete = {0} DistributionType(s) Deleted successfully.
```
Jonkno: deleted

PR #543, opened 2017-06-16
Fix NPE push Management UI on events

kaizimmerm:
For a reason that i do not fully understand yet it might happen that getEntity on UI vents does not work. NP checks have been in place however not in every location.

I fixed as well a performance problems. DS table was calling getEntity in N+1 fashion.

Reviewers: @schabdo

### DIFF
--- a/hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/distributions/dstable/DistributionSetTable.java
+++ b/hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/distributions/dstable/DistributionSetTable.java
@@ -13,6 +13,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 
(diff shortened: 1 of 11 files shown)


hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/distributions/dstable/DistributionSetTable.java
```diff
@@ -129,7 +131,7 @@ private void handleSelectedAndUpdatedDs(final List<DistributionSetUpdatedEvent>
     private void updateVisableTableEntries(final List<DistributionSetUpdatedEvent> events,
             final List<Long> visibleItemIds) {
         events.stream().filter(event -> visibleItemIds.contains(event.getEntityId()))
-                .filter(event -> event.getEntity().isComplete())
+                .filter(event -> !Objects.isNull(event.getEntity())).filter(DistributionSetUpdatedEvent::isComplete)
```
kaizimmerm: change order

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/distributions/dstable/DistributionSetTable.java
```diff
@@ -120,7 +122,7 @@ void onDistributionSetUpdateEvents(final DistributionSetUpdatedEventContainer ev
 
     private void handleSelectedAndUpdatedDs(final List<DistributionSetUpdatedEvent> events) {
         manageDistUIState.getLastSelectedDistribution()
-                .ifPresent(lastSelectedDsIdName -> events.stream()
+                .ifPresent(lastSelectedDsIdName -> events.stream().filter(event -> !Objects.isNull(event.getEntity()))
```
kaizimmerm: nonNull

PR #549, opened 2017-06-21
Alternative less expensive stats service.

kaizimmerm:
That does not run extra for every tenant.


michahirsch:
Maybe introducing an own object for the `SystemUsageWithTenantInformation`

PR #560, opened 2017-07-12
New API: Extension for storing offline-updates on hawkBit server

### LINKED ISSUE #242: Provide means to record manual off-line device updates into hawkBit's history
When having to manually update a device, e.g., for failure-recovery from a bad update, it would be nice to have a means to record this manual update into hawkBit's history so that it can serve as the single source of truth regarding the updates done to a device.

Currently, one can simulate the effect by creating an according softwaremodule/distributionset and "installing" that update on a device simulator resembling the real device in terms of configuration. However, this is a bit tedious and may mess up the history, e.g., in case the real device has performed an update in the meantime.

stormc:
Resolves #242

Introduction and Usage:
There are scenarios where the regular device update may fail or there is
a need for a manual intervention for device update. This feature is
provided by the Offline Update API which is built as a separate hawkBit
extension and is implemented based on Spring plug-in mechanism. The
hawkBit can be extended to provide this feature by adding
"offline.update.enabled=true" in the hawkbitdefaults.properties and
corresponding dependency in the pom.xml of hawkBit-update-server module.

Functionality:
This feature provides a new REST API which can be invoked to send the
details of the operating system or the firmware updates performed
offline on a single or a set of devices. The response is almost similar
to that of creation of a new DistributionSet.

Note : The new distribution set name is generated in the format
DS_<timestamp>_<uniqueId> and the version is always 1.0.0.

Mechanism:
1. Retrieves the list of controller Ids from the input.
2. Checks if the targets exist for corresponding controllers. If not
then throw appropriate exception.
3. Create the software modules. Upload artifacts if present, else update
only metadata with filename and corresponding href.
4. Create a DistributionSet for the given module types.
5. For each controller, create an action by assigning the distribution
set.
6. Mark all the actions created above as complete.

API : /rest/v1/distributionsets/offlineInstall

Request:
```
curl -X POST
http://127.0.0.1:8080/rest/v1/distributionsets/offlineInstall -H
'accept: application/json' -H 'content-type: multipart/form-data; -F
'offlineUpdateInfo={
  "softwareModules":[
    {
      "name":"offlineModule",
      "description":"operating system updated offline",
      "artifactData":[
        {
          "filename":"service.bin",
          "md5Hash":"a10f3b3bbdf5536550f6ac76e5a26db3",
          "sha1Hash":"dd31d32801d571c833295a568373b92d81ba2e15",
          "href":"https://FILESERVER/software/os/updates/service.bin",
          "version":"1.0"
        } ],
        "type":"os",
        "version":"1.0.0"
        } ],
  "migrationStepRequired":true,
  "controllerIds":["controller1"]
}'
```
Response:
```
{
  "createdBy": "admin",
  "createdAt": 1495121741558,
  "lastModifiedBy": "admin",
  "lastModifiedAt": 1495121741558,
  "name":
"DS_DEFAULT_1495121741474_06d0fbf3-b575-48b7-b5c7-8776bfbb9b2b",
  "version": "1.0.0",
  "modules": [
    {
      "createdBy": "admin",
      "createdAt": 1495121741438,
      "lastModifiedBy": "admin",
      "lastModifiedAt": 1495121741466,
      "name": "offlineModule",
      "description": "operating system updated offline",
      "version": "1.0.0",
      "type": "os",
      "_links": {
        "self": {
          "href": "http://127.0.0.1:8080/rest/v1/softwaremodules/1"
        },
        "type": {
          "href": "http://127.0.0.1:8080/rest/v1/softwaremoduletypes/1"
        }
      },
      "id": 1
    } ],
  "migrationStepRequired": true,
  "type": "os",
  "complete": true,
  "_links": {
    "self": {
      "href": "http://127.0.0.1:8080/rest/v1/distributionsets/1"
    },
    "type": {
      "href": "http://127.0.0.1:8080/rest/v1/distributionsettypes/2"
    },
    "metadata": {
      "href":
"http://127.0.0.1:8080/rest/v1/distributionsets/1/metadata?offset=0&limit=50{&sort,q}",
      "templated": true
    }
  },
  "id": 1
}
```
After a successful offline update request sent to the hawkBit server,
the below APIs can be used to verify the offline update action:

1. GetActionHistory

Request:
```
curl http://127.0.0.1:8080/rest/v1/targets/CONTROLLER_ID/actions
```

Response:
```
{
  "content": [
    {
      "createdBy": "admin",
      "createdAt": 1491987888972,
      "lastModifiedBy": "admin",
      "lastModifiedAt": 1491987889271,
      "type": "update",
      "status": "finished",
      "_links": {
        "self": {
          "href":
"http://127.0.0.1:8080/rest/v1/targets/controller1/actions/1"
        }
      },
      "id": 1
    } ],
  "total": 1,
  "size": 1
}
```

2. GetInstalledDS

Request:
```
curl http://127.0.0.1:8080/rest/v1/targets/TARGET_ID/installedDS
```

Response:
```
{
  "createdBy": "admin",
  "createdAt": 1491987888518,
  "lastModifiedBy": "admin",
  "lastModifiedAt": 1491987888518,
  "name":"DS_DEFAULT_1491987888198_d6ddff59-e793-47d6-9d4c-04b5bbde67df",
  "version": "1.0.0",
  "modules": [
    {
      "createdBy": "admin",
      "createdAt": 1491987888134,
      "lastModifiedBy": "admin",
      "lastModifiedAt": 1491987888134,
      "name": "offlineModule",
      "description": "operating system updated offline",
      "version": "1.1.0",
      "type": "os",
      "vendor": "Test Vendor",
      "_links": {
        "artifacts": {
          "href":
"http://127.0.0.1:8080/rest/v1/softwaremodules/1/artifacts"
        },
        "self": {
          "href": "http://127.0.0.1:8080/rest/v1/softwaremodules/1"
        },
        "type": {
          "href": "http://127.0.0.1:8080/rest/v1/softwaremoduletypes/1"
        },
        "metadata": {
          "href":
"http://127.0.0.1:8080/rest/v1/softwaremodules/1/metadata?offset=0&limit=50"
        }
      },
      "id": 1
    } ],
  "migrationStepRequired": true,
  "type": "os",
  "complete": true,
  "_links": {
    "self": {
      "href": "http://127.0.0.1:8080/rest/v1/distributionsets/1"
    },
    "type": {
      "href": "http://127.0.0.1:8080/rest/v1/distributionsettypes/1"
    },
    "metadata": {
      "href":
"http://127.0.0.1:8080/rest/v1/distributionsets/1/metadata?offset=0&limit=50{&sort,q}",
      "templated": true
    }
  },
  "id": 1
}
```


kaizimmerm:
Hi,

first of all thanks for the contribution!

However, I am not sure if I understand correctly why you have created this new resource. I mean 90% of what this resource is doing can already be done through Management API (e.g. create modules, upload artefacts, create a DistributionSet and assign the DS to targets).

Concerning the "completed action" feature: 

As mentioned before I definitely see this in hawkBit but I don't think that you can do this repository wise in two steps. Works partly for DDI but not DMF, i.e. as soon as you call  assignDistributionSet DMF clients get informed about a new update (TargetAssignDistributionSetEvent event) and start updating. Event in DDI case in there is a small window where DDI clients see the (intermediate) step as you do this in two transactions. Next to that I think there are some corner cases to think about (e.g. Target has already running actions).

I think the only way to support this is to make the completed action part of one transaction which does not send TargetAssignDistributionSetEvent but simply inserts the `Action`as finished into the history and updates the targets installed DS field.

I would think that we need in `DeploymentManagement` a new method:
```
public DistributionSetAssignmentResult assignedDistributionSet(final Long dsID, final Collection<String> controllerIDs)
```

and in Management API `MgmtTargetResource.postAssignedDistributionSet` and `MgmtDistributionSetResource.createAssignedTarget` a new optional GET parameter `boolean completed` which if true calls the new `DeploymentManagement. assignedDistributionSet`.

I guess that needs quite some refactoring in `JpaDeploymentManagement` and cannot be easily done in an extension without duplicating to much code so I believe it needs to go into core.

Please let me know if your are interested implementing it in such a direction. If not I could do that part as part of core and you can integrate the feature it into your module.

I see you put a lot of work into this but to be honest I am not convinced why the convenience part should be maintained by the project even as an extension.

Kai

singh-himanshu:
Hi @kaizimmerm, thanks for taking the time to review the PR.
> I mean 90% of what this resource is doing can already be done through Management API (e.g. create modules, upload artefacts, create a DistributionSet and assign the DS to targets).

You are right here. I think we had initially started with the use case that the device will be posting the offline update information or the initial state, and having a single resource would have required lesser change on the device side agent. Somewhere down the line, the implementation was shaped to support an offline administrator using a single REST resource.

> I don't think that you can do this repository wise in two steps. Works partly for DDI but not DMF, i.e. as soon as you call assignDistributionSet DMF clients get informed about a new update (TargetAssignDistributionSetEvent) and start updating.

Right, this exactly was the concern why we moved to a management API, under assumption that the device would still be offline when the updates are posted to server.

> Please let me know if your are interested implementing it in such a direction. If not I could do that as part of core and you can integrate the feature into your module.

Having a single API that takes only the list of files that have been installed and the list of devices is still our main concern.

But since we will not be put time on this for next 3 to 4 weeks at least, I suggest you can go ahead with changes you recommend, and we can then incorporate the feature later.

Regards,
Himanshu

kaizimmerm:
Ok, sounds great.

kaizimmerm:
#563 is on master now. Let me know if this works for you.

kaizimmerm:
I have not head about this for a  while. I will close this for now. Let me know if you think about another contribution.

stormc:
Considering the feature has been partly implemented and this version being somewhat outdated by now, it's probably best to close it as of now. We'll take a look and see if we have the need to contribute on top of the changes being done. Thanks, @kaizimmerm.

PR #561, opened 2017-07-12
DosFilter can be disabled.

kaizimmerm:
While at it refactored as well:
* Moved DosFilter in rest-security
* Moved etagFilter into rest-core

Reviewers: @michahirsch @schabdo 


hawkbit-autoconfigure/src/main/java/org/eclipse/hawkbit/autoconfigure/security/SecurityManagedConfiguration.java
```diff
@@ -355,6 +357,7 @@ protected void configure(final AuthenticationManagerBuilder auth) throws Excepti
          *         of service protection filter in the filter chain
          */
         @Bean
+        @ConditionalOnProperty(prefix = "hawkbit.server.security.dos.filter", name = "enabled", matchIfMissing = true)
```
michahirsch: For the `dosMgmtUiFilter` there is an extra property to enable it `hawkbit.server.security.dos.ui-filter` for `dos-ddif`, `dos-system` and `dosMgmt` it's all the same property, purpose?

kaizimmerm: There was no reason back in the day to configure them separately. UI simply has less requirements concerning amount of requests.

PR #571, opened 2017-08-29
Unused enum constant "skip" removed

schabdo:
The HandlingType "skip" is a leftover from the early days of HawkBit. Actually it was never used neither returned at API level (see [here](https://github.com/eclipse/hawkbit/blob/master/hawkbit-ddi-resource/src/main/java/org/eclipse/hawkbit/ddi/rest/resource/DdiRootController.java#L277])). So it can be removed to prevent confusion.

### DIFF
--- a/hawkbit-ddi-api/src/main/java/org/eclipse/hawkbit/ddi/json/model/DdiDeployment.java
+++ b/hawkbit-ddi-api/src/main/java/org/eclipse/hawkbit/ddi/json/model/DdiDeployment.java
@@ -64,14 +64,9 @@ public class DdiDeployment {
     }
 
     /**
-     * The handling type for the update action.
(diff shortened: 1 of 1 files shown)


singh-himanshu:
Sorry for a late comment...

We had planned to use this `enum` for providing maintenance window option (#245). Idea is that API returns `SKIP` when a maintenance window is not available. See last pull request [here](https://github.com/eclipse/hawkbit/pull/535) and relevant code line [here](https://github.com/eclipse/hawkbit/pull/535/commits/e35b0630554ed5743eb5faa430804ade213be87a#diff-f5f55359116301d299f5f10c0c47d8a7R260).

There are some changes since last pull request. Once tested we will update the pull request. We can add back the `enum` as part of the updated merge request.

PR #572, opened 2017-08-30
Flag parameter for management simulator to wait till scenario ends

Jonkno:
- Extended the Scenario for management api simulations with a flag parameter to decide if the scenario should wait till a running rollout has finished. 
- Fixed rollout deleted check 
- Added a setter for scenarios

Reviewers: @schabdo @kaizimmerm

### DIFF
--- a/examples/hawkbit-example-mgmt-simulator/src/main/java/org/eclipse/hawkbit/mgmt/client/ClientConfigurationProperties.java
+++ b/examples/hawkbit-example-mgmt-simulator/src/main/java/org/eclipse/hawkbit/mgmt/client/ClientConfigurationProperties.java
@@ -46,6 +46,7 @@ public class ClientConfigurationProperties {
      */
     public static class Scenario {
         private boolean cleanRepository;
+        private boolean waitTillRolloutComplete = true;
         private int targets = 100;
         private int distributionSets = 10;
         private int appModulesPerDistributionSet = 2;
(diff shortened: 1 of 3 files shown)


examples/hawkbit-example-mgmt-simulator/src/main/java/org/eclipse/hawkbit/mgmt/client/ClientConfigurationProperties.java
```diff
@@ -227,6 +236,10 @@ public void setRolloutSuccessThreshold(final short rolloutSuccessThreshold) {
         return scenarios;
     }
 
+    public void addScenarios(final Scenario scenario) {
```
Jonkno: look if add list is some were

examples/hawkbit-example-mgmt-simulator/src/main/java/org/eclipse/hawkbit/mgmt/client/scenarios/ConfigurableScenario.java
```diff
@@ -278,19 +281,30 @@ private void runRollout(final MgmtDistributionSet set, final Scenario scenario)
...
             } catch (final InterruptedException e) {
                 LOGGER.warn("Interrupted!");
                 Thread.currentThread().interrupt();
             }
-        } while (rolloutResource.getRollout(id).getStatusCode() != HttpStatus.NOT_FOUND);
+            try {
+                final ResponseEntity<MgmtRolloutResponseBody> rollout = rolloutResource.getRollout(id);
+                httpStatus = rollout.getStatusCode();
+            } catch (final FeignException e) {
+                LOGGER.info("FeignException " + e);
```
kaizimmerm: I don't get the reasoning behind this change. Looks like a quick debug hack that has nothing to do with this pull request. Am I right?

Jonkno: No, I ran into this issue when running the Scenario a second time and trying to delete an existing rollout. When we call `rolloutResource.getRollout(id)` and the rollout does not exist anymore , we will not get `ResponseEntity` with status code 404, but a feign exception is thrown, since we are using the feign client for our http request (exception are wrapped in feign exceptions). That's why we have to catch the feign exception and look if the reason was a 404.

PR #574, opened 2017-09-01
Readd skip type for upcoming maintenance window feature

schabdo:
Readd the previously removed `skip` status, as it will be used soon for the maintenance window feature (see  [here](https://github.com/eclipse/hawkbit/pull/571#issuecomment-326530905])).

This basically reverts commit: https://github.com/eclipse/hawkbit/pull/571/commits/54c4c8d1f0d7c8d354277d33660b834ca90cab96


schabdo:
@singh-himanshu: I'm sorry, I wasn't aware of that you'll going to make use of the unused `skip` type. Thanks for your hint!

PR #577, opened 2017-09-12
DMF health check

kaizimmerm:
Support for DMF health check, i.e. client can send PING message that will be responded with identical correlationId.

Reviewers: @michahirsch @Jonkno

### DIFF
--- a/examples/hawkbit-device-simulator/src/main/java/org/eclipse/hawkbit/simulator/DeviceSimulatorRepository.java
+++ b/examples/hawkbit-device-simulator/src/main/java/org/eclipse/hawkbit/simulator/DeviceSimulatorRepository.java
@@ -9,10 +9,11 @@
 package org.eclipse.hawkbit.simulator;
 
 import java.util.Collection;
+import java.util.HashSet;
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 
-import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
 
 /**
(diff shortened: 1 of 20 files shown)


examples/hawkbit-device-simulator/src/main/java/org/eclipse/hawkbit/simulator/DeviceSimulatorRepository.java
```diff
@@ -78,12 +79,17 @@ public AbstractSimulatedDevice remove(final String tenant, final String id) {
         return devices.remove(new DeviceKey(tenant.toLowerCase(), id));
     }
 
+    public Set<String> getTenants() {
+        return tenants;
```
michahirsch: instead storing the tenant to an extra set we could retrieve the tenants on the fly from the devices map.
```
public Set<String> getTenants() {
        devices.keySet().stream().map(key -> key.tenant).collect(Collectors.toSet());
    }
```

kaizimmerm: Sounds pretty expensive. I for instance simulate 500k devices most of the time.

examples/hawkbit-device-simulator/src/main/java/org/eclipse/hawkbit/simulator/amqp/SpReceiverService.java
```diff
@@ -82,28 +89,58 @@ public SpReceiverService(final RabbitTemplate rabbitTemplate, final AmqpProperti
...
         if (MessageType.THING_DELETED.equals(messageType)) {
+            checkContentTypeJson(message);
             repository.remove(tenant, thingId);
             return;
         }
 
+        if (MessageType.PING_RESPONSE.equals(messageType)) {
+            final String correlationId = new String(message.getMessageProperties().getCorrelationId(),
+                    StandardCharsets.UTF_8);
+            openPings.remove(correlationId);
```
michahirsch: I would check the return value of `remove` and in case we got a response with a correlationId which didn't exists I would probably log and error?
```
if(!openPings.remove(correlationId)){
  //TODO: Log an error 
}
```

PR #581, opened 2017-09-22
Consistent listing of hypermedia links

kaizimmerm:
The standard is that we list only the self link in collection GET requests and all links only on single entity GET requests. 

That was actually not implemented constantly for all resources.

Reviewers: @Jonkno

### DIFF
--- a/hawkbit-mgmt-resource/src/main/java/org/eclipse/hawkbit/mgmt/rest/resource/MgmtDistributionSetMapper.java
+++ b/hawkbit-mgmt-resource/src/main/java/org/eclipse/hawkbit/mgmt/rest/resource/MgmtDistributionSetMapper.java
@@ -117,6 +117,10 @@ public final class MgmtDistributionSetMapper {
         response.add(linkTo(methodOn(MgmtDistributionSetRestApi.class).getDistributionSet(response.getDsId()))
                 .withSelfRel());
 
+        return response;
+    }
+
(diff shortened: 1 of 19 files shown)


hawkbit-mgmt-resource/src/main/java/org/eclipse/hawkbit/mgmt/rest/resource/MgmtDistributionSetMapper.java
```diff
@@ -117,6 +117,10 @@ static MgmtDistributionSet toResponse(final DistributionSet distributionSet) {
         response.add(linkTo(methodOn(MgmtDistributionSetRestApi.class).getDistributionSet(response.getDsId()))
                 .withSelfRel());
 
+        return response;
+    }
+
+    static void addLinks(final DistributionSet distributionSet, final MgmtDistributionSet response) {
```
michahirsch: the method `addLinks`  could return the `response` then you could chain the calls in the actual resources implemenation instead of calling 
```
MgmtDistributionSet response = toResponse(set);
addLinks(set, response);
Response.ok(response);
```

```
MgmtDistributionSet response = addLinks(set, toResponse(set));
Response.ok(response);
```

kaizimmerm: Good point

PR #600, opened 2017-11-10
Artefact last modified is returned my DMF auth call

kaizimmerm:
* Use artifact createdAt field as binary is immutable.
* Add to DMF API


michahirsch:
Codacs is right and the private field `private AbstractDbArtifact file;` is not read in the `MgmtDownloadResourceTest` you can remove it as a field

PR #621, opened 2018-01-17
RedirectController into auto config as default and null pointer fix.

kaizimmerm:
RedirectController defines a default behaviour and as a result should not be in the runtime but autoconfiguration.

Fixed a small nullpointer as well.

Fixed in addition a few errors where where accessing data the user is not permitted to use.


hawkbit-autoconfigure/src/main/java/org/eclipse/hawkbit/autoconfigure/mgmt/ui/MgmtUiAutoConfiguration.java
```diff
@@ -41,13 +41,18 @@
 @Import(MgmtUiConfiguration.class)
 public class MgmtUiAutoConfiguration {
 
+    @Bean
+    RedirectController uiRedirectController() {
```
michahirsch: I would annotate it with `@ConditionalOnMissingBean` so it's possible to overwrite the redirect-controller in case you don't want to redirect `/` to `/UI/` like the default.

PR #622, opened 2018-01-17
Some small UI glitches 

melanieretter:
Several small UI glitches are fixed with that pull request:

* Create Rollout: When selecting a target filter which does not match at least one target, a form validation error is presented and the rollout cannot be saved.
rolloutfilternotmatchtarget

* The size of the table software modules inside the distribution set details is now exactly as the content to show. No more unnecessary scroll bars.
softwaremodulesfitstable

* The action id in the table Action History is now presented as a String value.
actionhistory

* If an unexpected exception occurs the error notification displays the exception message and it disappears after 1 second. No manual closing is necessary anymore.

Reviewer: @Jonkno 


hawkbit-ui/src/main/resources/messages.properties
```diff
@@ -542,6 +542,7 @@ message.rollout.duplicate.check = Rollout [ {0} ] must be unique, entered value
 message.correct.invalid.value = Please correct invalid values
 message.enter.number = Please enter number
 message.rollout.field.value.range = Value should be in range {0} to {1}
+message.rollout.filter.target.exists = The selected target filter does not match any existing target.
```
blomark: delete .

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/rollout/rollout/AddUpdateRolloutWindowLayout.java
```diff
@@ -868,6 +873,18 @@ private int getGroupSize() {
         return (int) Math.ceil((double) totalTargetsCount / Double.parseDouble(noOfGroups.getValue()));
     }
 
+    class TargetExistsValidator implements Validator {
```
Jonkno: I would call the class different, e.g. TotalTargetsCountExistsValidator

melanieretter: I prefer TargetExistsValidator

PR #624, opened 2018-01-29
Introduce new permission structure for rollout management.

kaizimmerm:
As with target and repository we now have now (C/R/U/D)_ROLLOUT and in addition HANDLE_ROLLOUT for start/pause/resume.

### DIFF
--- a/hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/SpPermissionChecker.java
+++ b/hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/SpPermissionChecker.java
@@ -14,11 +14,11 @@ import org.eclipse.hawkbit.im.authentication.PermissionService;
 import org.eclipse.hawkbit.im.authentication.SpPermission;
 
 /**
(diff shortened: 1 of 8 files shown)


hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/SpPermissionChecker.java
```diff
@@ -126,41 +126,46 @@ public boolean hasDeleteRepositoryPermission() {
...
     /**
-     * Gets the SP rollout create permission.
-     * 
-     * @return permission for rollout create
+     * @return <code>true</code> if rollout create permission
      */
     public boolean hasRolloutCreatePermission() {
-        return hasUpdateTargetPermission() && hasReadRepositoryPermission()
-                && permissionService.hasPermission(SpPermission.ROLLOUT_MANAGEMENT);
+        return hasReadRepositoryPermission() && permissionService.hasPermission(SpPermission.CREATE_ROLLOUT);
```
blomark: add readTarget permission

PR #626, opened 2018-02-05
Improvements repository validation constraints

kaizimmerm:
* Rep API -> switch from `@NotNull Long` to `long`
* Create/update builder: Validate incoming strings against HTML tags (i.e. XSS attempts)


kaizimmerm:
[CQ15540](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=15540)

hawkbit-repository/hawkbit-repository-api/src/main/java/org/eclipse/hawkbit/repository/ArtifactManagement.java
```diff
@@ -63,7 +63,7 @@
      *             if given software module does not exist
      */
     @PreAuthorize(SpringEvalExpressions.HAS_AUTH_UPDATE_REPOSITORY)
```
stefbehl: Would have expected SpringEvalExpressions.HAS_AUTH_CREATE_REPOSITORY for this method...

kaizimmerm: Good catch

PR #641, opened 2018-02-22
Fix gpg validation

oflebbe:
Signed-off-by: Olaf Flebbe <of@oflebbe.de>

Getting gpg keys is insecure. First anyone can send keys to the PGP Keyserver (and create collision attacks), secondly the connection to pgp.mit.edu is not secure, it is plain http. We should put other measures like checking fingerprints into the loop. Third even the download server was specified with http only, so an attacker had to possibility to forge the download and the gpg key verification, rendering the complete verification ineffective.

See for instance "there is no point in ..." answer in this article for an explanation: https://security.stackexchange.com/questions/4161/shouldnt-gpg-key-fetching-use-a-secure-connection

I would recommend to put the actual GPG certificates into the code, since this establishes trust between the download and the user which is supposed to have signed the download.

Cheers 
   Olaf


michahirsch:
Hey @oflebbe,

thanks for the contribution. I see your point and totally agree to add the public keys into the code and import them in the dockerfile directly instead of getting the gpg keys during build time from gpg server.

Using https to download the artifact from maven-central is also a good point and was wrong in the dockerfiles. 

Thanks for that
Cheers,
Michael

PR #650, opened 2018-03-05
Add missing option searching for empty field (e.g. description=='')

kaizimmerm:
Fixed in addition a work builder constraint that metadata value cannot be empty which is in fact not how the entity is defined.


hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaEntityFactory.java
```diff
@@ -57,7 +58,7 @@
 
     @Override
     public MetaData generateMetadata(final String key, final String value) {
-        return new JpaDistributionSetMetadata(key, value);
+        return new JpaDistributionSetMetadata(key, StringUtils.trimWhitespace(value));
```
blomark: trim key

PR #656, opened 2018-03-08
Support for MS SQL Server and RabbitMQ 3.7

kaizimmerm:
* Added support for MS SQL Server 2017 (DDLs and test run support).
* Added DDLs for DB2 as well (however, as no JDBC driver as OSS exist for that we will keep the support pretty low by means of maintaining DDLs only for the time being).
* Fixed in general a bunch auf SQL/JPA problems where DB2 and/or SQL Server where less forgiving than MySQL or H2, e.g.
  * Teached RSQL utils to handle DBs differently, e.g. use different escaping mechanism  in SQL Server.
  * Always define escape char in RSQL utils "llike search escaping".
  * Ensure that we as well in controller case write update audit fields.
  * Added missing "non null"s as not all DBs support foreign keys if otherwise
* Fixed problem (by upgrading HTTP admin client) with RabbitMQ 3.7 in tests.
* I took the opportunity to migrate more beans to constructor injection as I had to work on Bean dependencies in the repository anyhow.

Reviewers: @schabdo


kaizimmerm:
CQ for Rabbit http-client upgrade: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=15867

PR #683, opened 2018-05-15
Ignore cookies with script content in login UI

kaizimmerm:
Reduce risk that an attacker can inject a script into the cookie values and execute scripts
locally on a victim’s computer.


schabdo:
@hawkbit-bot verify please

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/login/AbstractHawkbitLoginUI.java
```diff
@@ -81,6 +82,7 @@
 
     private static final String SP_LOGIN_USER = "sp-login-user";
     private static final String SP_LOGIN_TENANT = "sp-login-tenant";
+    private static final Pattern FORBIDDEN_COOKIE_CONTENT = Pattern.compile(".*(<|>).*");
```
blomark: unit test?

PR #684, opened 2018-05-16
MS SQL Server support in hawkBit runtime

kaizimmerm:


schabdo:
@hawkbit-bot verify please

PR #685, opened 2018-05-16
Improve user experience for uploading an artifact

blomark:
- removed confirmation window when uploading a file
- moved upload button to the right side of the window to save space in the bottom of the page


schabdo:
@hawkbit-bot verify please

PR #698, opened 2018-06-15
Add only eventbus subscription where a method listens to it.

melanieretter:
If an eventbus is subscribed to a view where no listener method exists, Vaadin framework is logging a warning. To avoid blowing up log files with this warning, we subscribe only views to the eventbus where exists at least one method that listens to it.


schabdo:
@hawkbit-bot verify please

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/common/table/AbstractTable.java
```diff
@@ -83,6 +83,15 @@ protected AbstractTable(final UIEventBus eventBus, final VaadinMessageSource i18
         setDefault();
         addValueChangeListener(event -> onValueChange());
         setPageLength(SPUIDefinitions.PAGE_SIZE);
+        subscribeToEventBus();
```
blomark: insert method with boolean return value

schabdo:
@hawkbit-bot verify please

PR #715, opened 2018-07-30
Management API supports sorting distributionset by "createdAt" #687

### LINKED ISSUE #687: Management API supports sorting distributionset by "createdAt"
When executing the REST call to the Hawkbit Management API
`rest/v1/distributionsets?sort=createdAt%3ADESC`
I get the following response: 400 Bad Request
```
{
    "exceptionClass": "org.eclipse.hawkbit.rest.exception.SortParameterUnsupportedFieldException",
    "errorCode": "hawkbit.server.error.rest.param.invalidField",
    "message": "The given sort parameter field does not exist"
}
```
However, a similar call succeeds:
`rest/v1/distributionsets?sort=id%3ADESC`
Result is: 200 ok.

P.S: I have also tried `created_at`, `createdat`. Without success.

laverman:
Fixes #687 

Signed-off-by: Jeroen Laverman <jeroen.laverman@bosch-si.com>

### DIFF
--- a/hawkbit-core/src/main/java/org/eclipse/hawkbit/repository/DistributionSetFields.java
+++ b/hawkbit-core/src/main/java/org/eclipse/hawkbit/repository/DistributionSetFields.java
@@ -26,6 +26,14 @@ public enum DistributionSetFields implements FieldNameProvider {
      */
     DESCRIPTION("description"),
     /**
+     * The createdAt field.
+     */
+    CREATEDAT("createdAt"),
+    /**
+     * The lastModifiedAt field.
+     */
+    LASTMODIFIEDAT("lastModifiedAt"),
+    /**
      * The version field.
      */
     VERSION("version"),


hawkbit-core/src/main/java/org/eclipse/hawkbit/repository/DistributionSetFields.java
```diff
@@ -25,6 +25,14 @@
      * The description field.
      */
     DESCRIPTION("description"),
+    /**
+     * The createdAt field.
+     */
+    CREATEDAT("createdAt"),
```
stefbehl: I'd prefer "CREATED_AT" and "LAST_MODIFIED_AT"

laverman: While I agree, that it would improve readability, it would be inconsistent with the other field enums (cf. TargetFields).

stefbehl: Ok, makes sense... thanks!

PR #722, opened 2018-08-16
RFC: Make HttpControllerPreAuthenticatedSecurityHeaderFilter a bean

stschake:
This makes it possible to overwrite the security header filter in an
extensions, which is desirable since the current default hawkBit
implementation relies on custom HTTP headers.

Marking as RFC since it's messy to overwrite the AuthenticationManager later on.
I'm also new to Spring Boot in general so it's possible I'm missing a more idiomatic way of doing this.


laverman:
Hi @stschake thanks for contributing to hawkBit! 👍

schabdo:
@stschake Glad to see that you took the time to contribute. Please give me a few more days to have a look at your pull request

schabdo:
Indeed that's a tricky one. Didn't find a better way of extracting the filter to a conditional bean so far. 
@michahirsch do you have any thoughts on this?

schabdo:
@stschake @michahirsch and I didn't find a better approach so I think we go for it for the time being

PR #734, opened 2018-08-30
Fix rollout approval caption

laverman:


bogdan-bondar:
The messages are save to delete

PR #736, opened 2018-09-04
Include attribute values in target search text filter

stschake:
Most of our devices are registered through plug&play so the majority of useful information is in the controller attributes. This PR also applies the search text filter to the attribute values.

### DIFF
--- a/hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/specifications/TargetSpecifications.java
+++ b/hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/specifications/TargetSpecifications.java
@@ -38,6 +38,7 @@ import org.eclipse.hawkbit.repository.model.Target;
 import org.eclipse.hawkbit.repository.model.TargetTag;
 import org.eclipse.hawkbit.repository.model.TargetUpdateStatus;
 import org.springframework.data.jpa.domain.Specification;
+import org.springframework.data.jpa.domain.Specifications;
 
 /**
  * Specifications class for {@link Target}s. The class provides Spring Data JPQL
@@ -165,6 +166,18 @@ public final class TargetSpecifications {
             query.distinct(true);
             return cb.like(cb.lower(attributeMap.value()), searchTextToLower);
         };
+    }
+
+    /**
+     * {@link Specification} for retrieving {@link Target}s by "like
+     * controllerId or like name or like description or like attribute value".
+     *
+     * @param searchText
+     *            to be filtered on
+     * @return the {@link Target} {@link Specification}
+     */
+    public static Specification<JpaTarget> likeIdOrNameOrDescriptionOrAttributeValue(final String searchText) {
+        return Specifications.where(likeIdOrNameOrDescription(searchText)).or(likeAttributeValue(searchText));
     }
 
     /**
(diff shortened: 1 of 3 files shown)


hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/specifications/TargetSpecifications.java
```diff
@@ -148,6 +150,35 @@ private TargetSpecifications() {
...
+     * @param searchText
+     *            to be filtered on
+     * @return the {@link Target} {@link Specification}
+     */
+    public static Specification<JpaTarget> likeAttributeValue(final String searchText) {
+        return (targetRoot, query, cb) -> {
+            final String searchTextToLower = searchText.toLowerCase();
+            final MapJoin<JpaTarget, String, String> attributeMap = targetRoot.join(JpaTarget_.controllerAttributes,
+                    JoinType.LEFT);
+            return cb.or(cb.like(cb.lower(attributeMap.value()), searchTextToLower));
```
stefbehl: In order to eliminate duplicates in the search result, you need to set the distinct flag to true:
`query.distinct(true);`

You can easily recreate this by creating a target with a controller ID that is equal to one of the attribute values. In that case the target would appear twice in the list.

schabdo:
@hawkbit-bot verify please

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/specifications/TargetSpecifications.java
stschake: I've updated the test to check for that, but can't seem to reproduce the issue with it.

stefbehl: Thanks for updating the test! I could easily recreate this in the UI a few weeks ago but I will double-check...

stefbehl: Hi Stefan,
The new test fails for me! However, if I add query.distinct(true) as proposed above, everything works fine :-) Would be great if you could add that.
Thanks and regards.

stschake: It seems to pass on CircleCI? Happy to add it, just wondering what's going on there.

stefbehl: That's indeed weird. I ran the tests in my Eclipse and there it failed. After adding the distinct flag that tests passed... Anyway, thanks for adding this line!

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/specifications/TargetSpecifications.java
```diff
@@ -148,6 +150,36 @@ private TargetSpecifications() {
...
+     *            to be filtered on
+     * @return the {@link Target} {@link Specification}
+     */
+    public static Specification<JpaTarget> likeAttributeValue(final String searchText) {
+        return (targetRoot, query, cb) -> {
+            final String searchTextToLower = searchText.toLowerCase();
+            final MapJoin<JpaTarget, String, String> attributeMap = targetRoot.join(JpaTarget_.controllerAttributes,
+                    JoinType.LEFT);
+            query.distinct(true);
+            return cb.or(cb.like(cb.lower(attributeMap.value()), searchTextToLower));
```
stefbehl: The "cb.or" at the beginning of the return statement is not needed. Please remove it. Thanks

stschake: Done. I've also removed the whitespace changes.

PR #743, opened 2018-09-21
Added translation resource file for Chinese

ten2net:


laverman:
I'm afraid, I have to reject this PR, since (besides the missing IP validation) we can't maintain nor verify this contribution. Sorry.

PR #772, opened 2018-12-10
Fix missing localization

laverman:


hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/AbstractHawkbitUI.java
```diff
@@ -164,7 +168,7 @@ public void afterViewChange(final ViewChangeEvent event) {
         navigator.addView(EMPTY_VIEW, new Navigator.EmptyView());
         // set locale is required for I18N class also, to get the locale from
         // cookie
-        final String locale = getLocaleId(SPUIDefinitions.getAvailableLocales());
+        final String locale = getLocaleId(new HashSet<>(uiProperties.getLocalization().getAvailableLocals()));
```
schabdo: If you'll change signature of `getLocaleId(Set<String)` to `getLocaleId(List<String)` there is no need for conversion  to HashSet

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/UiProperties.java
```diff
@@ -39,6 +42,41 @@ public void setGravatar(final boolean gravatar) {
...
+    /**
+     * Localization information
+     */
+    public static class Localization implements Serializable {
+        private static final long serialVersionUID = 1L;
+
+        /**
+         * Default localization
+         */
+        private String defaultLocal;
```
schabdo: What about `default` only? Looks better `uiProperties.getLocalization().getDefault()`. You could also set the default value here so you don't have to specify the default within `application properties`

laverman: `default` is a keyword and can't be used as variable, or can it?

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/UiProperties.java
```diff
@@ -39,6 +42,41 @@ public void setGravatar(final boolean gravatar) {
...
+
+        /**
+         *  List of available localizations
+         */
+        private List<String> availableLocals;
```
schabdo: You could also add the available local "en" here as well, since it is the default

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/rollout/rollout/AddUpdateRolloutWindowLayout.java
```diff
@@ -1106,18 +1106,11 @@ private void setAutoStartType(final Rollout rollout) {
...
-        private final String value;
-
-        ERROR_THRESHOLD_OPTIONS(final String val) {
-            value = val;
-        }
-
-        private String getValue() {
-            return value;
+        private String getValue(final VaadinMessageSource i18n) {
+            return  i18n.getMessage("label.errorthreshold.option." + name().toLowerCase());
```
schabdo: Be careful here: Dynamic labels are hard to maintain. If you search e.g. for `label.errorthreshold.option.percent` your IDE won't find it hence you might conclude it's not used anywhere and remove it.

What about `PERCENT(label.errorthreshold.option.percent), COUNT(label.errorthreshold.option.count);` ?

laverman: good point. like it.

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/UiProperties.java
schabdo: Yes indeed, sometimes Java is doing things I don't expect. Method `getDefault()` is ok, but `private String default` not

laverman: Would it be acceptable to name the getter different than the variable?

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/rollout/rollout/AddUpdateRolloutWindowLayout.java
```diff
@@ -1106,18 +1106,17 @@ private void setAutoStartType(final Rollout rollout) {
...
+        COUNT("label.errorthreshold.option.count");
 
-        COUNT("Count");
+        private String value;
 
-        private final String value;
-
-        ERROR_THRESHOLD_OPTIONS(final String val) {
-            value = val;
+        ERROR_THRESHOLD_OPTIONS(final String value) {
```
schabdo: Constructor can be private as well

laverman: IMHO this is redundant, since it will be private anyway, right?

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/UiProperties.java
schabdo: I don't think you can convince Spring to deal with different names. But would be fine for me if we go for `defaultLocal` (as it is)

PR #775, opened 2018-12-18
Enable/disable buttons in action history based on user permission

blomark:
The buttons  in action history table
- Cancel Action
- Force Action
- Force Quit
should only be enabled if the user has target update permission


schabdo:
@hawkbit-bot verify please

PR #777, opened 2018-12-20
Distribution set assignment: Fix NPE in AbstractDsAssignmentStrategy

stefbehl:
Assume a target with controller ID "Target1". If you assign a distribution set via the Mgmt REST API by sending a HTTP POST to 
`[hawkbit-update-server]/rest/v1/distributionsets/[distribution-set-id]/assignedTargets`
you need to specify this controller ID in the POST body of the request.

If the controller ID is specified using a different spelling (lower case vs. upper case vs. mixed case), e.g. "TARGET1" instead of "Target1", you will encounter the following NullPointerException:

java.lang.NullPointerException: null
at o.e.h.r.j.AbstractDsAssignmentStrategy.createTargetAction(AbstractDsAssignmentStrategy.java:227)
at o.e.h.r.j.OnlineDsAssignmentStrategy.createTargetAction(OnlineDsAssignmentStrategy.java:95)
at o.e.h.r.j.JpaDeploymentManagement.lambda$assignDistributionSetToTargets$4(JpaDeploymentManagement.java:314)

This change delivers a fix for this test case (incl. an appropriate unit test).

Signed-off-by: Stefan Behl <stefan.behl@bosch-si.com>

### DIFF
--- a/hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaDeploymentManagement.java
+++ b/hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaDeploymentManagement.java
@@ -88,9 +88,6 @@ import org.springframework.transaction.annotation.Transactional;
 import org.springframework.util.CollectionUtils;
 import org.springframework.validation.annotation.Validated;
 
-import com.fasterxml.jackson.core.JsonProcessingException;
-import com.fasterxml.jackson.databind.ObjectMapper;
-import com.fasterxml.jackson.databind.ObjectWriter;
 import com.google.common.base.Joiner;
 import com.google.common.collect.Lists;
 
@@ -288,7 +285,7 @@ public class JpaDeploymentManagement implements DeploymentManagement {
                 targetEntitiesIdsChunks);
 
         final Map<String, JpaAction> controllerIdsToActions = createActions(targetsWithActionType, targetEntities,
-                assignmentStrategy, controllerIDs, distributionSetEntity);
+                assignmentStrategy, distributionSetEntity);
         // create initial action status when action is created so we remember
         // the initial running status because we will change the status
         // of the action itself and with this action status we have a nicer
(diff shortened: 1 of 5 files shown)


stefbehl:
@hawkbit-bot verify please

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaDeploymentManagement.java
```diff
@@ -376,13 +373,11 @@ private void setAssignedDistributionSetAndTargetUpdateStatus(final AbstractDsAss
...
             final List<JpaTarget> targets, final AbstractDsAssignmentStrategy assignmentStrategy,
-            final List<String> controllerIDs, final JpaDistributionSet set) {
+            final JpaDistributionSet set) {
         final Map<String, TargetWithActionType> targetsWithActionMap = targetsWithActionType.stream()
                 .collect(Collectors.toMap(TargetWithActionType::getControllerId, Function.identity()));
 
-        return targets.stream()
-                .map(trg -> createTargetAction(assignmentStrategy, targetsWithActionType, controllerIDs, targets,
-                        targetsWithActionMap, trg, set))
+        return targets.stream().map(trg -> assignmentStrategy.createTargetAction(targetsWithActionMap, trg, set))
```
bogdan-bondar: You could get the nullpointer exception in `actionRepository::save` method, if `createTargetAction` returns null. So it is better to filter the null objects out.

hawkbit-rest/hawkbit-mgmt-resource/src/test/java/org/eclipse/hawkbit/mgmt/rest/resource/MgmtDistributionSetResourceTest.java
```diff
@@ -273,6 +273,32 @@ public void assignMultipleTargetsToDistributionSet() throws Exception {
...
+    public void assignTargetsToDistributionSetIgnoreCase() throws Exception {
+        final DistributionSet createdDs = testdataFactory.createDistributionSet();
+
+        // prepare targets
+        final String[] knownTargetIds = new String[] { "64-da-a0-02-43-8b", "Trg1", "target2", "target4" };
+        final String[] knownTargetIdDifferentCase = new String[] { "64-DA-A0-02-43-8b", "TRG1", "TarGET2", "target4" };
+        for (final String targetId : knownTargetIds) {
+            testdataFactory.createTarget(targetId);
+        }
+        final JSONArray list = new JSONArray();
```
bogdan-bondar: You could also move it to `JsonBuilder` class, so it can be reused later

stefbehl: For consistency reasons, I would prefer to stick to the current "manual" approach. That would also require JsonBuilder to offer a method like "targetsWithActionType" which returns the serialized JSON representation).

bogdan-bondar:
Looks good!

PR #780, opened 2019-01-10
Fix NullpointerException in UploadProgressInfoWindow

blomark:
Fixed Nullpointer Exception in case upload was interrupted and UploadProgressInfoWindow gets updated.


schabdo:
@hawkbit-bot verify please

PR #791, opened 2019-01-31
Fix JavaDoc syntax errors. Add check to circle build.

kaizimmerm:
It seems that the maven Javadoc Plugin update has now lint enabled by default. 

In order to have our release build run again I did:

* Reduced lint to syntax check only.
* Fixed many syntax errors in JavaDoc.
* Moved one class that is JPA specific out of the repository API package (i.e. it had JPA JavaDoc dependencies).
* Add JavaDoc check to Circle so that we can see errors in PRs.

While at it I updated EclipseLInk to 2.7.4

### DIFF
--- a/hawkbit-core/src/main/java/org/eclipse/hawkbit/repository/FieldNameProvider.java
+++ b/hawkbit-core/src/main/java/org/eclipse/hawkbit/repository/FieldNameProvider.java
@@ -28,17 +28,16 @@ public interface FieldNameProvider {
     String SUB_ATTRIBUTE_SEPERATOR = ".";
 
     /**
-     * @return the string representation of the underlying persistence field
-     *         name e.g. in case of sorting. Never {@code null}.
+     * @return the string representation of the underlying persistence field name
+     *         e.g. in case of sorting. Never {@code null}.
      */
     String getFieldName();
 
     /**
      * Contains the sub entity the given field.
      *
-     * @param propertyField
-     *            the given field
-     * @return <true> contains <false> contains not
+     * @param propertyField the given field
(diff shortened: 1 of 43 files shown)


schabdo:
@hawkbit-bot verify please

hawkbit-core/src/main/java/org/eclipse/hawkbit/repository/FieldNameProvider.java
```diff
@@ -76,7 +76,7 @@ default boolean containsSubEntityAttribute(final String propertyField) {
     /**
      * Is the entity field a {@link Map}.
      *
-     * @return <true> is a map <false> is not a map
+     * @return <code>true</code> is a map false is not a map
```
laverman: surround 'false' also with <code> tag

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/common/CommonDialogWindow.java
```diff
@@ -554,22 +554,23 @@ public void setCancelButtonEnabled(final boolean enabled) {
...
+         * @return <code>true</code> = window will close <code>false</code> =
+         *         will not closed.
          */
         default boolean canWindowClose() {
             return true;
         }
 
         /**
-         * Saves/Updates action. Is called if canWindowSaveOrUpdate is <true>.
+         * Saves/Updates action. Is called if canWindowSaveOrUpdate is true.
```
laverman: Please add <code> tag to 'true'

kaizimmerm:
@laverman good catch, done.

PR #799, opened 2019-02-13
Feature extend target attributes constraints

bogdan-bondar:
The limit for Target controller attributes key is extended to 128 characters. This allows devices to send attributes, having composite keys.


hawkbit-repository/hawkbit-repository-api/src/main/java/org/eclipse/hawkbit/repository/model/Target.java
```diff
@@ -38,7 +38,7 @@
     /**
      * Maximum length of key of controller attribute
      */
-    int CONTROLLER_ATTRIBUTE_KEY_SIZE = 32;
+    int CONTROLLER_ATTRIBUTE_KEY_SIZE = 128;
```
blomark: check if Mgmt API and UI validate this

schabdo:
@hawkbit-bot verify please

hawkbit-repository/hawkbit-repository-api/src/main/java/org/eclipse/hawkbit/repository/model/Target.java
bogdan-bondar: There is no possibility to add/change controller attributes within Mgmt API/UI as they are coming from devices per DDI/DMF, and are additionally validated in `JpaControllerManagement`

PR #802, opened 2019-02-27
Added `/target/generated-sources/apt` to resources for maven and m2e

vigier:
Additionally it was needed to use the `build-helper-maven-plugin`, as
m2e would otherwise exclude by default all sources from the additional
source folder (s. https://stackoverflow.com/a/7759911 )

Signed-off-by: Peter Vigier <Peter.Vigier@bosch-si.com>

### DIFF
--- a/hawkbit-repository/hawkbit-repository-jpa/pom.xml
+++ b/hawkbit-repository/hawkbit-repository-jpa/pom.xml
@@ -24,7 +24,7 @@
       <jaxb.api.version>2.3.0</jaxb.api.version>
       <javax.annotation.version>1.3.1</javax.annotation.version>
    </properties>
-   
+
    <dependencies>
       <!-- Hawkbit -->
       <dependency>
@@ -61,7 +61,7 @@
       <dependency>
          <groupId>cz.jirutka.rsql</groupId>
          <artifactId>rsql-parser</artifactId>
-      </dependency>      
+      </dependency>
       <dependency>
          <groupId>com.google.guava</groupId>
          <artifactId>guava</artifactId>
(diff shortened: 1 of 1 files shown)


schabdo:
@hawkbit-bot verify please

StefanKlt:
👍 looks good, thanks

hawkbit-repository/hawkbit-repository-jpa/pom.xml
```diff
@@ -91,6 +91,24 @@
    </dependencies>
 
    <build>
+      <resources>
+         <!-- We need to add also the default-resources, as this configuration overwrites the defaults -->
+         <resource>
+            <directory>src/main/java</directory>
```
StefanKlt: This default-resources seems to be included even if it is removed from this list. Please check if it is needed here.

hawkbit-repository/hawkbit-repository-jpa/pom.xml
```diff
@@ -91,6 +91,24 @@
...
+         </resource>
+         <resource>
+            <directory>src/main/resources</directory>
+         </resource>
+         <resource>
+            <directory>target/generated-sources/apt</directory>
+         </resource>
+         <resource>
+            <directory>src/test/java</directory>
```
StefanKlt: This default-resources seems to be included even if it is removed from this list. Please check if it is needed here.

vigier: I removed the resource section completely as for adding an additional "source" folder the build-helper-plugin alone is enough.

PR #807, opened 2019-03-13
Fix localization vaadin system messages

StefanKlt:
Enable localization of Vaadin system messages


schabdo:
@hawkbit-bot verify please

PR #815, opened 2019-03-29
Improve request rejected exception handling

a-sayyed:
The default behavior of the StrictHttpFirewall results in a 500 internal server error when a request url contains potentially malicious strings. This Exception shoulld instead be caught and translated into a 400 BAD REQUEST response.

### DIFF
--- a/hawkbit-core/src/main/java/org/eclipse/hawkbit/exception/SpServerError.java
+++ b/hawkbit-core/src/main/java/org/eclipse/hawkbit/exception/SpServerError.java
@@ -232,7 +232,12 @@ public enum SpServerError {
      * invalid.
      */
     SP_AUTO_ASSIGN_DISTRIBUTION_SET_INVALID("hawkbit.server.error.repo.invalidAutoAssignDistributionSet",
-            "The given distribution set for auto-assignment is invalid: it is either incomplete (i.e. mandatory modules are missing) or soft deleted");
+            "The given distribution set for auto-assignment is invalid: it is either incomplete (i.e. mandatory modules are missing) or soft deleted"),
+
+    /**
+     * Error message informing that the current request url contains malicious characters
+     */
+    SP_MALICIOUS_URL_STRING("hawkbit.server.error.rest.maliciousUrl", "The given url contains malicious characters");
 
     private final String key;
     private final String message;
--- a/hawkbit-http-security/src/main/java/org/eclipse/hawkbit/security/aop/FilterChainProxyAdvice.java
+++ b/hawkbit-http-security/src/main/java/org/eclipse/hawkbit/security/aop/FilterChainProxyAdvice.java
@@ -1,0 +1,78 @@
+/**
+ * Copyright (c) 2019 Bosch Software Innovations GmbH and others.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package org.eclipse.hawkbit.security.aop;
+
+import java.io.IOException;
+
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+import org.aspectj.lang.ProceedingJoinPoint;
+import org.aspectj.lang.annotation.Around;
+import org.aspectj.lang.annotation.Aspect;
+import org.eclipse.hawkbit.exception.SpServerError;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.http.MediaType;
+import org.springframework.security.web.FilterChainProxy;
+import org.springframework.security.web.firewall.RequestRejectedException;
+import org.springframework.security.web.firewall.StrictHttpFirewall;
+
+/**
+ * Workaround for issue
+ * https://github.com/spring-projects/spring-security/issues/5007
+ *
+ * Advice class for the {@link FilterChainProxy} to handle the
+ * {@link RequestRejectedException} thrown by the {@link StrictHttpFirewall} and
+ * logged by Jetty/Tomcat.
+ *
+ * Using AspectJ advice around the {@link FilterChainProxy} makes it possible to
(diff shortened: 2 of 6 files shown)


schabdo:
@hawkbit-bot verify please

hawkbit-core/src/main/java/org/eclipse/hawkbit/exception/SpServerError.java
```diff
@@ -232,7 +232,12 @@
      * invalid.
      */
     SP_AUTO_ASSIGN_DISTRIBUTION_SET_INVALID("hawkbit.server.error.repo.invalidAutoAssignDistributionSet",
-            "The given distribution set for auto-assignment is invalid: it is either incomplete (i.e. mandatory modules are missing) or soft deleted");
+            "The given distribution set for auto-assignment is invalid: it is either incomplete (i.e. mandatory modules are missing) or soft deleted"),
+
+    /**
+     * Error message informing that the current request url contains malicious characters
+     */
+    SP_MALICIOUS_URL_STRING("hawkbit.server.error.rest.maliciousUrl", "The given url contains malicious characters or strings");
```
laverman: The distinction between characters and strings is irrelevant. Therefore, I would suggest to remove "or strings"

hawkbit-runtime/hawkbit-update-server/src/main/java/org/eclipse/hawkbit/app/aop/StrictHttpFirewallAdvice.java
```diff
@@ -0,0 +1,72 @@
...
+import org.springframework.stereotype.Component;
+
+/**
+ * Advice class for the {@link FilterChainProxy} to handle the
+ * {@link RequestRejectedException} thrown by the {@link StrictHttpFirewall}.
+ * Results in an error code 400 BAD_REQUEST instead of the (default) 500
+ */
+@Aspect
+@Component
+public class StrictHttpFirewallAdvice {
```
schabdo: Wouldn't the `hawkbit-http-security` maven module be a better place for this class?

hawkbit-runtime/hawkbit-update-server/src/main/java/org/eclipse/hawkbit/app/aop/StrictHttpFirewallAdvice.java
```diff
@@ -0,0 +1,72 @@
...
+import org.springframework.security.web.FilterChainProxy;
+import org.springframework.stereotype.Component;
+
+/**
+ * Advice class for the {@link FilterChainProxy} to handle the
+ * {@link RequestRejectedException} thrown by the {@link StrictHttpFirewall}.
+ * Results in an error code 400 BAD_REQUEST instead of the (default) 500
+ */
+@Aspect
+@Component
```
schabdo: Please remove the `@Component` annotation and register it within `SecurityAutoConfiguration` so it can be better customised later

hawkbit-runtime/hawkbit-update-server/src/main/java/org/eclipse/hawkbit/app/aop/StrictHttpFirewallAdvice.java
```diff
@@ -0,0 +1,72 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.http.MediaType;
+import org.springframework.security.web.firewall.RequestRejectedException;
+import org.springframework.security.web.firewall.StrictHttpFirewall;
+import org.springframework.security.web.FilterChainProxy;
+import org.springframework.stereotype.Component;
+
+/**
+ * Advice class for the {@link FilterChainProxy} to handle the
```
schabdo: May I ask you to extend the comment and explain why you're using an advice instead of configuring the bean via:

```
@Bean
public HttpFirewall myFirewall() {
    // ...
}
```

hawkbit-runtime/hawkbit-update-server/pom.xml
```diff
@@ -59,6 +59,16 @@
          <groupId>com.microsoft.sqlserver</groupId>
          <artifactId>mssql-jdbc</artifactId>
       </dependency>
+      <!-- TEST -->
+      <dependency>
+         <groupId>org.springframework.boot</groupId>
+         <artifactId>spring-boot-starter-test</artifactId>
```
schabdo: I think there is a scope `test` missing but I'd rather prefer to change `hawkbit-parent`'s pom.xml

a-sayyed: Yes, thank you

schabdo: Welcome. May I ask you to add the `test` scope [to the managed location](https://github.com/eclipse/hawkbit/blob/master/pom.xml#L756) as well?
I think it would be good if `test` is the default scope since it is a "test-only" dependency

hawkbit-http-security/src/main/java/org/eclipse/hawkbit/security/aop/FilterChainProxyAdvice.java
```diff
@@ -0,0 +1,79 @@
...
+     * {@link StrictHttpFirewall} and sends a 400 BAD REQUEST response
+     *
+     * @param pjp
+     *            the {@link ProceedingJoinPoint} around the {@link FilterChainProxy}.doFilter() method
+     */
+    @Around("execution(public void org.springframework.security.web.FilterChainProxy.doFilter("
+            + "javax.servlet.ServletRequest, javax.servlet.ServletResponse, javax.servlet.FilterChain))")
+    @SuppressWarnings("squid:S1166") // Exception should not be (fully) logged nor rethrown
+    public void handleRequestRejectedExceptionInFilterChain(final ProceedingJoinPoint pjp) throws Throwable {
+        LOG.debug("Applying advice FilterChainProxyAdvice around request");
```
bogdan-bondar: remove, because `doFilter` is called for every request, so it will pollute our logs in DEBUG level

a-sayyed: Good catch, will remove it, thanks!

PR #816, opened 2019-04-01
Fix cast exception in UI

StefanKlt:
Table.getValue() returns either an ID or set of IDs. This must be checked before casting the object.

Signed-off-by: Stefan Klotz <stefan.klotz@bosch-si.com>


schabdo:
@hawkbit-bot verify please

PR #819, opened 2019-04-05
Bugfix npe target percentage

Nkyn:
Set the value range of target percentage to "1 to 100" in the Configure Rollouts view. The value '0' is no longer valid for the target percentage in the Advanced Group definition.


schabdo:
@hawkbit-bot verify please

bogdan-bondar:
@hawkbit-bot verify please

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/rollout/rollout/DefineGroupsLayout.java
```diff
@@ -331,22 +334,29 @@ private void setGroupsValidation(final RolloutGroupsValidation validation) {
...
+
+    private void singleGroupsValidation(final RolloutGroupsValidation validation, final GroupRow lastRow,
+            final boolean remainingError) {
+        final int maxTargets = quotaManagement.getMaxTargetsPerRolloutGroup();
+
         for (int i = 0; i < groupRows.size(); ++i) {
             final GroupRow row = groupRows.get(i);
             // do not mask the 'remaining targets' error
-            if (hasRemainingTargetsError && row.equals(lastRow)) {
+            if (remainingError && row.equals(lastRow)) {
```
bogdan-bondar: You don't need the _remainingError_ method parameter, and _ hasRemainingTargetsError_ field, just use the `validationStatus == ValidationStatus.INVALID` here directly:
```suggestion
            if (validationStatus == ValidationStatus.INVALID && row.equals(lastRow)) {
```

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/rollout/rollout/DefineGroupsLayout.java
```diff
@@ -331,22 +334,29 @@ private void setGroupsValidation(final RolloutGroupsValidation validation) {
...
                 continue;
             }
             row.resetError();
-            final Long count = groupsValidation.getTargetsPerGroup().get(i);
-            if (count != null && count > maxTargets) {
-                row.setError(i18n.getMessage(MESSAGE_ROLLOUT_MAX_GROUP_SIZE_EXCEEDED, maxTargets));
-                setValidationStatus(ValidationStatus.INVALID);
+            if (validation != null) {
```
bogdan-bondar: You could use the _groupsValidation_ member field here, so you could remove the _validation_ method parameter

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/rollout/rollout/DefineGroupsLayout.java
```diff
@@ -463,12 +473,18 @@ private void removeGroupRow(final GroupRow groupRow) {
...
-                    new FloatRangeValidator(message, 0F, 100F).validate(value);
+                    rangeValidator = new FloatRangeValidator(message, 0F, 100F);
+                }
+                else if (value instanceof Integer) {
+                    rangeValidator = new IntegerRangeValidator(message, 0, 100);
                 }
-                if (value instanceof Integer) {
-                    new IntegerRangeValidator(message, 0, 100).validate(value);
+                else {
+                    throw new Validator.InvalidValueException(i18n.getMessage("message.rollout.field.value.type"));
```
bogdan-bondar: I would use the _message.enter.number_ here as before, because the entered value is not empty and not a number in this case. And in case the value itself is empty (null), I would define the message like "Field can not be empty"

hawkbit-ui/src/main/resources/messages.properties
```diff
@@ -616,6 +616,7 @@ message.rollout.name.empty = Please enter a name for Rollout
 message.correct.invalid.value = Please correct invalid values
 message.enter.number = Please enter number
 message.rollout.field.value.range = Value should be in range {0} to {1}
+message.rollout.field.value.type = Invalid type of value
```
bogdan-bondar: remove this (see comments above)

schabdo:
@hawkbit-bot verify please

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/rollout/rollout/DefineGroupsLayout.java
Nkyn: Non-numeric values are handled somewhere else in the first place, showing the message "Could not convert value to Float". Thus, the Exception here is never thrown anyway (or only in unexpected cases). But to avoid introducing a new message string I'll just change this as you suggested. 
An empty field is already handled by throwing the EmptyValueException and showing the _message.enter.number_.

PR #824, opened 2019-04-12
Feature fixed timezone

laverman:
In order to display timestamps in the UI, hawkBit tries to find out the time zone of the client (browser). Therefore, the GMT offset is taken and mapped to a valid time zone. However, since multiple time zones have the same GMT offset, hawkBit may select the wrong time zone.

In such scenarios, it might be off help to fix the time zone to a certain zone, for all clients. Ignoring their reported offset. 

This feature is disabled by default and can be used by setting the following property:

`hawkbit.server.ui.fixedTimeZone=Europe/Berlin`

### DIFF
--- a/hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/UiProperties.java
+++ b/hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/UiProperties.java
@@ -25,7 +25,7 @@ public class UiProperties implements Serializable {
 
     private boolean gravatar;
 
-    private String timezone;
+    private String fixedTimeZone;
 
     private final Localization localization = new Localization();
 
@@ -45,12 +45,12 @@ public class UiProperties implements Serializable {
         this.gravatar = gravatar;
     }
 
-    public String getTimezone() {
-        return timezone;
+    public String getFixedTimeZone() {
+        return fixedTimeZone;
     }
 
-    public void setTimezone(final String timezone) {
-        this.timezone = timezone;
+    public void setFixedTimeZone(final String fixedTimeZone) {
+        this.fixedTimeZone = fixedTimeZone;
     }
 
 
--- a/hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/UiProperties.java
+++ b/hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/UiProperties.java
@@ -9,7 +9,6 @@
 package org.eclipse.hawkbit.ui;
 
 import java.io.Serializable;
-import java.sql.Time;
 import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
@@ -51,6 +50,9 @@ public class UiProperties implements Serializable {
     }
 
 
+    /**
+     * Time zone information
+     */
     public static class Timezone implements Serializable {
         private static final long serialVersionUID = 1L;
 
--- a/hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/UiProperties.java
+++ b/hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/UiProperties.java
@@ -9,6 +9,7 @@
 package org.eclipse.hawkbit.ui;
 
 import java.io.Serializable;
+import java.sql.Time;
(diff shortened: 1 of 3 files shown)


hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/UiProperties.java
```diff
@@ -43,6 +45,31 @@ public void setGravatar(final boolean gravatar) {
...
 
+    public Timezone getTimezone() {
+        return timezone;
+    }
+
+
+    /**
+     * Time zone information
+     */
+    public static class Timezone implements Serializable {
```
laverman: Use String instead of Class

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/utils/SPDateTimeUtil.java
```diff
@@ -41,12 +42,28 @@ private SPDateTimeUtil() {
...
+    }
+
     /**
      * Get browser time zone.
      *
      * @return TimeZone
      */
     public static TimeZone getBrowserTimeZone() {
+
+        if (!fixedTimeZoneProperty.isEmpty()) {
```
laverman: Check NULL

laverman: Use SpringUtils

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/utils/SPDateTimeUtil.java
```diff
@@ -41,12 +42,28 @@ private SPDateTimeUtil() {
...
+        SPDateTimeUtil.fixedTimeZoneProperty = fixedTimeZoneProperty;
+    }
+
     /**
      * Get browser time zone.
      *
      * @return TimeZone
      */
     public static TimeZone getBrowserTimeZone() {
+
```
laverman: Adapt JavaDoc

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/UiProperties.java
```diff
@@ -25,6 +25,8 @@
 
     private boolean gravatar;
 
+    private String timezone;
```
bogdan-bondar: I would call it `fixedTimezone` or `customTimezone`, in order to show, that we are setting it explicitly

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/utils/SPDateTimeUtil.java
```diff
@@ -41,12 +43,29 @@ private SPDateTimeUtil() {
...
-     * Get browser time zone.
+     * Get browser time zone or fixed time zone if configured
      *
      * @return TimeZone
      */
     public static TimeZone getBrowserTimeZone() {
+
+
+        if (StringUtils.isEmpty(fixedTimeZoneProperty)) {
+            return TimeZone.getTimeZone(fixedTimeZoneProperty);
```
bogdan-bondar: Please, test the valid/invalid `fixedTimeZoneProperty`

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/utils/SPDateTimeUtil.java
```diff
@@ -41,12 +43,29 @@ private SPDateTimeUtil() {
...
     /**
-     * Get browser time zone.
+     * Get browser time zone or fixed time zone if configured
      *
      * @return TimeZone
      */
     public static TimeZone getBrowserTimeZone() {
+
+
+        if (StringUtils.isEmpty(fixedTimeZoneProperty)) {
```
bogdan-bondar: add ! :)

laverman: 🙈 good catch ... regression ...

PR #829, opened 2019-04-29
Add multi-user support

stschake:
This allows optionally configuring multiple static users with varying
permissions. If used, Spring Security user/password are ignored.
Otherwise, the old behavior is retained.

This is a version for core hawkBit of https://github.com/eclipse/hawkbit-extensions/pull/43

### DIFF
--- a/hawkbit-autoconfigure/src/main/java/org/eclipse/hawkbit/autoconfigure/security/InMemoryUserManagementAutoConfiguration.java
+++ b/hawkbit-autoconfigure/src/main/java/org/eclipse/hawkbit/autoconfigure/security/InMemoryUserManagementAutoConfiguration.java
@@ -9,6 +9,8 @@
 package org.eclipse.hawkbit.autoconfigure.security;
 
 import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashMap;
 import java.util.List;
 
 import org.eclipse.hawkbit.im.authentication.MultitenancyIndicator;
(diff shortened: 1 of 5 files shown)


schabdo:
@hawkbit-bot verify please

hawkbit-autoconfigure/src/main/java/org/eclipse/hawkbit/autoconfigure/security/InMemoryUserManagementAutoConfiguration.java
```diff
@@ -57,17 +65,62 @@ public void configure(final AuthenticationManagerBuilder auth) throws Exception
...
     UserDetailsService userDetailsService() {
-        final InMemoryUserDetailsManager inMemoryUserDetailsManager = new InMemoryUserPrincipalDetailsManager();
-        inMemoryUserDetailsManager.setAuthenticationManager(null);
-        final SecurityProperties.User user = securityProperties.getUser();
-        final UserBuilder userBuilder = User.builder().username(user.getName()).password(user.getPassword())
-                .authorities(PermissionUtils.createAllAuthorityList());
-        final List<String> roles = user.getRoles();
-        if (!roles.isEmpty()) {
-            userBuilder.roles(roles.toArray(new String[roles.size()]));
+        final String defaultTenant = "DEFAULT";
```
schabdo: May I ask you to extract tenant as constant and reuse it within `TenantDaoAuthenticationProvider#createSuccessAuthentication` (PR [line 142](https://github.com/eclipse/hawkbit/blob/master/hawkbit-autoconfigure/src/main/java/org/eclipse/hawkbit/autoconfigure/security/InMemoryUserManagementAutoConfiguration.java#L89)) as well?

schabdo:
Thanks a lot for all the effort you put in this contribution. Works like a charm!  Please have a look at the open review comments so that we're good to go for master

PR #836, opened 2019-05-09
put JAVA options in dockerfile in a env variable 

monotek:
put JAVA options in a env variable so it can be overwritten on container start be using "-e"


monotek:
OK, seems this makes no sense. We thought command line arguments overwrite env vars but it works with: 

sudo docker run -it --rm -e JAVA_TOOL_OPTIONS=-Xms1024m hawkbit/hawkbit-update-server:0.3.0M2

PR #838, opened 2019-05-09
fix the logic that marks distributionSet as complete

smy4kor:
A DistributionSet (DS) can only be assigned to a target, when it is marked as complete.
A DS is marked as complete, when all requirements of the selected DS-Type are fulfilled.
A DS-Type determines which SoftwareModules (SM) of which SM-Type can be (optional) or have to be (required) contained.

When specifying a DS-Type that contains only optional SMs, a DS is marked as complete even without any SM assigned. This shall be changed.

A DS shall be only considered as complete, when:

SMs of the required SM-Types are assigned
At least one SM is assigned, in case all SM-Types are optional

### DIFF
--- a/hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/model/JpaDistributionSetType.java
+++ b/hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/model/JpaDistributionSetType.java
@@ -10,6 +10,7 @@ package org.eclipse.hawkbit.repository.jpa.model;
 
 import java.util.Collections;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
@@ -209,8 +210,12 @@ public class JpaDistributionSetType extends AbstractJpaNamedEntity implements Di
 
     @Override
     public boolean checkComplete(final DistributionSet distributionSet) {
-        return distributionSet.getModules().stream().map(SoftwareModule::getType).collect(Collectors.toList())
-                .containsAll(getMandatoryModuleTypes());
+        List<SoftwareModuleType> smTypes = distributionSet.getModules().stream().map(SoftwareModule::getType)
+                .collect(Collectors.toList());
+        if (smTypes.isEmpty()) {
+            return false;
+        }
+        return smTypes.containsAll(getMandatoryModuleTypes());
     }
 
     @Override


schabdo:
@hawkbit-bot verify please

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/model/JpaDistributionSetType.java
```diff
@@ -209,8 +210,12 @@ public void setKey(final String key) {
 
     @Override
     public boolean checkComplete(final DistributionSet distributionSet) {
-        return distributionSet.getModules().stream().map(SoftwareModule::getType).collect(Collectors.toList())
-                .containsAll(getMandatoryModuleTypes());
+        List<SoftwareModuleType> smTypes = distributionSet.getModules().stream().map(SoftwareModule::getType)
+                .collect(Collectors.toList());
+        if (smTypes.size() == 0) {
```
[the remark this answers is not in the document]
smy4kor: Done.

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/model/JpaDistributionSetType.java
```diff
@@ -209,8 +210,12 @@ public void setKey(final String key) {
 
     @Override
     public boolean checkComplete(final DistributionSet distributionSet) {
-        return distributionSet.getModules().stream().map(SoftwareModule::getType).collect(Collectors.toList())
-                .containsAll(getMandatoryModuleTypes());
+        List<SoftwareModuleType> smTypes = distributionSet.getModules().stream().map(SoftwareModule::getType)
```
a-sayyed: could be simplified:
```
final Set<SoftwareModuleType> mandatoryModuleTypes = getMandatoryModuleTypes();
return distributionSet.getModules().stream().map(SoftwareModule::getType).allMatch(type -> mandatoryModuleTypes.contains(type));
```

smy4kor: This will not work in 1 specific case. When distributionSet.getModules() is empty and mandatoryModuleTypes is also empty then allMatch returns true. I just tried it.

I guess that allMatch assumes true and looks for a false. And in above, no iteration happens.

a-sayyed: You are right, we can avoid that by adding:
```
        if (smTypes.isEmpty()) {
            return false;
        }
```
before the stream call

smy4kor: I asume you are referring to something like this:
```
 if (distributionSet.getModules().isEmpty()) {
            return false;
        }
        final Set<SoftwareModuleType> mandatoryModuleTypes = getMandatoryModuleTypes();
        return distributionSet.getModules().stream().map(SoftwareModule::getType)
                .allMatch(type -> mandatoryModuleTypes.contains(type));
```

This has a problem. It returns false even when I assign a software module to this type. In theory, this is a case where a DS type has no mandatory SM type. But it has a SM assigned. So it must be marked complete right?

smy4kor: Its for 2 resons:
- An incomplete DS should not be visible in deployment view.
- If all software modules are removed from a given DS, then this DS should be marked incomplete.

a-sayyed: Ok, now I got it.
Maybe a good idea here is to introduce some unit tests? specially if a Model class has some logic like this

a-sayyed: You can maybe use the `DistributionSetTypeManagementTest` test, and test the different use  cases  where this method is called, here it's called in `constructor`, `addModule`, and `removeModule`

smy4kor: Discussed with @stefbehl to add the test cases later as an improvement. The tests will be added in the next sprint.

PR #842, opened 2019-05-17
Fix rollout distribution bar style in UI

StefanKlt:
The progress bars in the rollouts view were not coloured correctly due to a bug in the generation of their html class

Signed-off-by: Stefan Klotz <stefan.klotz@bosch-si.com>


schabdo:
@hawkbit-bot verify please

PR #854, opened 2019-06-19
Feature enable cors

bogdan-bondar:
* Enabled cors in security configuration, added corresponding properties
* Added test for cors validation


hawkbit-security-core/src/main/java/org/eclipse/hawkbit/security/HawkbitSecurityProperties.java
```diff
@@ -68,6 +73,69 @@ public Clients getClients() {
...
+
+        public void setEnabled(final boolean enabled) {
+            this.enabled = enabled;
+        }
+
+        public List<String> getAllowedOrigins() {
+            return allowedOrigins;
+        }
+
+        public void setAllowedOrigins(List<String> allowedOrigins) {
```
dobleralex: add final

PR #861, opened 2019-07-08
Use FilterParanms instead if single method parameters

vigier:
Center logic distributed in different places in the FilterParam-class
Avoid use of `null`

Signed-off-by: Peter Vigier <Peter.Vigier@bosch-si.com>

### DIFF
--- a/hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/management/targettable/TargetBeanQuery.java
+++ b/hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/management/targettable/TargetBeanQuery.java
@@ -31,7 +31,6 @@ import org.springframework.data.domain.PageRequest;
 import org.springframework.data.domain.Slice;
 import org.springframework.data.domain.Sort;
 import org.springframework.data.domain.Sort.Direction;
-import org.springframework.util.CollectionUtils;
 import org.springframework.util.StringUtils;
 import org.vaadin.addons.lazyquerycontainer.AbstractBeanQuery;
 import org.vaadin.addons.lazyquerycontainer.QueryDefinition;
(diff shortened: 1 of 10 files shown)


schabdo:
@hawkbit-bot verify please

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/management/targettable/TargetBeanQuery.java
```diff
@@ -76,16 +71,20 @@ public TargetBeanQuery(final QueryDefinition definition, final Map<String, Objec
...
-            noTagClicked = (Boolean) queryConfig.get(SPUIDefinitions.FILTER_BY_NO_TAG);
-            distributionId = (Long) queryConfig.get(SPUIDefinitions.FILTER_BY_DISTRIBUTION);
-            searchText = (String) queryConfig.get(SPUIDefinitions.FILTER_BY_TEXT);
-            targetFilterQueryId = (Long) queryConfig.get(SPUIDefinitions.FILTER_BY_TARGET_FILTER_QUERY);
+            String searchText = (String) queryConfig.get(SPUIDefinitions.FILTER_BY_TEXT);
             if (!StringUtils.isEmpty(searchText)) {
                 searchText = String.format("%%%s%%", searchText);
             }
+
+            filter = new FilterParams(
```
schabdo: Potential NPE in case `queryConfig` is null or empty => `filter` will be null. I think before this couldn't happen due to null-checks on primitive types. I'd suggest do set an empty filter in `case queryConfig` is null or empty.

schabdo:
I'm afraid the current changes broke at least the target filter element box on "Deployment Management" view - no filtering is possible anymore.

PR #867, opened 2019-07-17
fixed typos in server error codes and messages

singrob:
Fixed typos in Server-Error codes and messages. 

### DIFF
--- a/hawkbit-core/src/main/java/org/eclipse/hawkbit/exception/SpServerError.java
+++ b/hawkbit-core/src/main/java/org/eclipse/hawkbit/exception/SpServerError.java
@@ -22,7 +22,7 @@ public enum SpServerError {
     /**
     *
     */
-    SP_REPO_ENTITY_ALREADY_EXISTS("hawkbit.server.error.repo.entityAlreadyExists",
+    SP_REPO_ENTITY_ALREADY_EXISTS("hawkbit.server.error.repo.entitiyAlreayExists",
             "The given entity already exists in database"),
 
     /**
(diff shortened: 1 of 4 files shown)


hawkbit-core/src/main/java/org/eclipse/hawkbit/exception/SpServerError.java
```diff
@@ -17,12 +17,12 @@
...
     *
     */
-    SP_REPO_GENERIC_ERROR("hawkbit.server.error.repo.genericError", "unknown error occured"),
+    SP_REPO_GENERIC_ERROR("hawkbit.server.error.repo.genericError", "unknown error occurred"),
 
     /**
     *
     */
-    SP_REPO_ENTITY_ALRREADY_EXISTS("hawkbit.server.error.repo.entitiyAlreayExists",
+    SP_REPO_ENTITY_ALREADY_EXISTS("hawkbit.server.error.repo.entityAlreadyExists",
```
schabdo: I would really love to fix this ugly typos in the message keys, but I'm afraid this is part of the API exposed to all clients out there in the wild. I can't tell who is relying on this specific error code shipped within a REST response. So better leave the key as it is. However fixing message text is highly appreciated!

schabdo: Just to be on same page: `hawkbit.server.error.repo.entityAlreadyExists` is the message key. The rest is totally fine to fix.

singrob: Got it, will update the PR

schabdo: Awesome, thanks a lot!

PR #871, opened 2019-07-29
fixed NullPointerException in deployment view

singrob:

### DIFF
--- a/hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/management/DeploymentView.java
+++ b/hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/management/DeploymentView.java
@@ -407,12 +407,12 @@ public class DeploymentView extends AbstractNotificationView implements BrowserW
     protected Map<Class<?>, RefreshableContainer> getSupportedPushEvents() {
         final Map<Class<?>, RefreshableContainer> supportedEvents = Maps.newHashMapWithExpectedSize(10);
 
-        if (targetTableLayout != null) {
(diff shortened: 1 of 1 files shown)


hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/management/DeploymentView.java
```diff
@@ -407,11 +407,15 @@ private void showOrHideFilterButtons(final int browserWidth) {
     protected Map<Class<?>, RefreshableContainer> getSupportedPushEvents() {
         final Map<Class<?>, RefreshableContainer> supportedEvents = Maps.newHashMapWithExpectedSize(10);
 
-        supportedEvents.put(TargetCreatedEventContainer.class, targetTableLayout.getTable());
-        supportedEvents.put(TargetDeletedEventContainer.class, targetTableLayout.getTable());
+        if (targetTableLayout != null) {
```
schabdo: I'd prefer to use same logic as already in place in other methods of this class. What do you think:
```
if (permChecker.hasTargetReadPermission()) {
    supportedEvents.put(TargetCreatedEventContainer.class, targetTableLayout.getTable());
    supportedEvents.put(TargetDeletedEventContainer.class, targetTableLayout.getTable());
}
```

singrob: I was thinking about performance, since reading the security context is quite slow in comparison to a simple null check. 
But you are right, to keep it consistent within the class I should reuse the "permChecker".
I'll update the PR :)

PR #873, opened 2019-07-30
Reverse proxy support

fksu:
Added reverse proxy support in download links via environment variables.


schabdo:
Thanks for your PR. May I ask you what exactly do you want to achieve? A lot of the properties are already configurable so maybe (depending on your use-case) they are already sufficient

fksu:
We have a HawkBit installed in a Docker container hosted at a virtual machine and behind a reverse proxy. We use SERVER_SERVLET_CONTEXT_PATH environment variable to specify the sub-path to be forwarded to the correct virtual machine.

It works for most of the configuration UI (e.g. logouts do not work), but not for the download links: they remain in the format defined in the hawkbit-artifactdl-defaults.properties and they get wrong fields since the communication between the reverse proxy and the virtual machine (on the same host, but both VMs) is HTTP.

The configuration is a little complicated now since I have to recompile HawkBit whenever I change hawkbit-artifactdl-defaults.properties, so it is not good for automatic updates.

schabdo:
I see, thanks for shedding light on your setup of hawkBit. Did I understand it correct? 

```
                            +---------+
                HTTP /a     |         |
                +---------->+ hawkBit |
                |           |         |
           +----+----+      +---------+
 HTTPS /   |         |
+----------+  Proxy  |
           |         |
           +----+----+      +---------+
                |           |         |
                +---------->+ hawkBit |
                HTTP /b     |         |
                            +---------+
```

Now the download URL should be: `https://myhostname:8080/{a or b}/....`, right?

fksu:
Exactly! Except for /b points to a different service (such as Jira). Thank you for depicting.

schabdo:
Got it! Next try: 

```
         +---------+         +-----------+
 HTTPS / |         | HTTP /a |           |
+------->+  Proxy  +-------->+  hawkBit  |
         |         |         |           |
         +---------+         +-----------+
```
Download URL: `https://myhostname:<proxyPort>/a/...`, right?

fksu:
Actually proxyPort is always 443 since we use the default HTTPS port. I just didn't want to complicate the code by checking empty vs unset values of HAWKBIT_PORTREQUEST.

dasty:
Also if it helps, here are the links to our initial discussions on Gitter:

https://gitter.im/eclipse/hawkbit?at=5cdeb6569404bf2aed60d6fd
https://gitter.im/eclipse/hawkbit?at=5cdec9f40ac9852a95f3bb34

we configured "server.servlet.context-path=/hawkbit" and we were able to administrate hawkbit and thinking that we are done as everything seemed to be working. But than we noticed that hawkbit is sending URLs like this to our devices http://extern.streamunlimited.com:8080/some-path which were wrong because
1) there is nothing running on our reverse proxy server on port 8080
2) our reverse proxy is only properly forwarding https request and not http requests
3) to hawkbit server only URLs within "/hawkbit" are forwarded.

Therefore posted patches allow us to configure URL send to devices 
**from wrong** http://extern.streamunlimited.com:8080/some-path
**to correct** https://extern.streamunlimited.com/hawkbit/some-path

fksu:
Any update? Is there anything I could help with, @schabdo?

schabdo:
Sorry, I got distracted. Thanks for asking. Yes, now I remember the discussion we had on Gitter. 
Your PR looks good. I'd have done almost the same to tackle the proxy path issue. Maybe we could simplify it a bit more - do you really need the regex in the path?

What about adding another placeholder to the map within `PropertyBasedArtifactUrlHandler#getReplaceMap` ?

```
private static final String CONTEXT_PATH_PLACEHOLDER = "contextPath";

// ....

replaceMap.put(CONTEXT_PATH_PLACEHOLDER, protocol.getContextPath());  
```
=> `contextPath` property needs to be added to `ArtifactUrlHandlerProperties#UrlProtocol`

Than you should be able to just set:

```
hawkbit.artifact.url.protocols.download-http.ref={protocol}://{hostname}:{port}/{contextPath}/{tenant}/controller/v1/{controllerId}/softwaremodules/{softwareModuleId}/artifacts/{artifactFileName}
```
Either via environment variable or in `application.properties`

fksu:
No problem. I agree with the solution via contextPath, but I would keep the regex and application.properties changed to avoid another environment variable.

schabdo:
> I would keep the regex and application.properties changed to avoid another environment variable.

Exactly, you can put anything what suits you best in your `application.properties`

schabdo:
I think the easiest way is to set in in your `application.properties`, so just like this:

```
server.servlet.context-path=/hawkbit
hawkbit.artifact.url.protocols.download-http.ref={protocol}://{hostname}:{port}/{contextPath}/{tenant}/controller/v1/{controllerId}/softwaremodules/{softwareModuleId}/artifacts/{artifactFileName}
hawkbit.artifact.url.protocols.download-http.contextPath=${server.servlet.context-path}
```

fksu:
That's not exactly what I meant, I already send it in SERVER_SERVLET_CONTEXT_PATH environment variable.

I would like to implement getContextPath() correctly: the easiest way would be returning System.getenv("SERVER_SERVLET_CONTEXT_PATH"), but I would like to infer it properly from protocol variable.

fksu:
Thanks for your suggestion. I will incorporate it into the code and amend the commit.

fksu:
Theoretically, it would be possible to avoid the regex; but using `contextPath` could be confusing then, since `${server.servlet.context-path}` should start and not end with a slash to avoid double slashes in the links, although double slashes work OK in this case (and most cases but cookies with paths could cause troubles).

It is also possible to use either `/hawkbit` or `/hawkbit/` for `${server.servlet.context-path}`, but both `hawkbit/` and `hawkbit` are invalid (at least for Apache). And we also have to consider the users without `contextPath`, so replacing only the field would be incorrect as well.

schabdo:
Thanks for bringing this to my attention! I think I found a way to set correct path by only adding this to your `application.properties`:
```
server.servlet.context-path=/hawkbit
hawkbit.artifact.url.protocols.download-http.ref={protocol}://{hostname}:{port}${server.servlet.context-path}/{tenant}/controller/v1/{controllerId}/softwaremodules/{softwareModuleId}/artifacts/{artifactFileName}
``` 
This should do the trick without any code changes

fksu:
Of course, I know you can solve the whole settings in `application.properties` but the point is you have to recompile it then what excludes automatic updates on servers. Although this restraint can be overcome by setting it via environment variables I assume the support of reverse proxies should be automatic in recent tools.

fksu:
OK, I was able to change it via an environment variable now. Although, it would be better if HawkBit supported a reverse proxy by default (in the code).

ntx-ben:
@schabdo Hi, I'm trying to accomplish the same thing. Running 0.3.0M5. Is this now in M7?

PR #884, opened 2019-08-13
Feature event publishing

bogdan-bondar:
* Refactored code in order to use EventPublisherHolder where possible instead of auto-wiring application event publisher and bus properties explicitly. 

This solves the problem of Service Matcher not updating its service Id in case of Bus Properties change (e.g. getting new random value for bus Id after EnvironmentChange event).


schabdo:
@hawkbit-bot verify please

hawkbit-repository/hawkbit-repository-core/src/main/java/org/eclipse/hawkbit/repository/model/helper/EventPublisherHolder.java
```diff
@@ -10,26 +10,29 @@
...
 import org.springframework.context.ApplicationEventPublisher;
 
 /**
- * A singleton bean which holds the event publisher to have to the cache manager
- * in beans not instantiated by spring e.g. JPA entities or
- * CacheFieldEntityListener which cannot be autowired.
+ * A singleton bean which holds the event publisher and service origin Id in
+ * order to publish remote application events. It can be used in the cache
+ * manager and beans not instantiated by spring e.g. JPA entities or
+ * CacheFieldEntityListener which cannot be auto-wired.
```
schabdo: May I ask you to remove none existing interface `CacheFieldEntityListener`?

PR #885, opened 2019-08-19
Security fix for UI auto refresh

aktivk:
Currently UI event bus only accepts basic auth token to verify tenant information and this fix will also work with oauth2 based token


schabdo:
@hawkbit-bot verify please

schabdo:
@aktivk thanks a lot for your contribution! Seems like you haven't signed the ECA. Please do so otherwise we can't accept your contribution

schabdo:
Thanks 👍

PR #887, opened 2019-09-16
improved documentation of ddi-rootcontroller resource's links list

singrob:


hawkbit-rest/hawkbit-rest-docs/src/test/java/org/eclipse/hawkbit/rest/ddi/documentation/DdiApiModelProperties.java
```diff
@@ -108,7 +108,7 @@
 
     static final String FILENAME = "file name of artifact";
 
-    static final String TARGET_CONFIG_DATA = "configuration data as key/value list";
+    static final String TARGET_CONFIG_DATA = "Configuration data as key/value list. Only showed for initial configuration, after UI request or a successful update";
```
stefbehl: I think this is (still) misleading because this link cannot be used to retrieve config data... how about sth like this:
_Link which is provided to the provisioning target or device to push its configuration (aka "controller attributes") to the server. Only shown for the initial configuration, after a successful update action, or if requested explicitly (e.g. via the management UI)._

Issue #786, opened 2019-01-25
Allow to download an artifact from the UI


diegorondini:
Once uploaded, there is now way to download an artifact from the UI.

It would be nice to be able to download the artifact from the UI.

PR #905, opened 2019-10-29
Allow to download an artifact from the UI (#786)

AnbazhakanSubramaniam:
This feature allows to download the artifact from the Artifact details
table

Signed-off-by: usb1cob <Anbazhakan.Subramaniam@in.bosch.com>

### DIFF
--- a/hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/artifacts/details/ArtifactDetailsLayout.java
+++ b/hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/artifacts/details/ArtifactDetailsLayout.java
@@ -8,6 +8,13 @@
  */
 package org.eclipse.hawkbit.ui.artifacts.details;
 
+import java.io.File;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.channels.Channels;
+import java.nio.channels.FileChannel;
+import java.nio.channels.ReadableByteChannel;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
@@ -16,8 +23,10 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 
+import org.eclipse.hawkbit.artifact.repository.model.AbstractDbArtifact;
 import org.eclipse.hawkbit.repository.ArtifactManagement;
 import org.eclipse.hawkbit.repository.SoftwareModuleManagement;
+import org.eclipse.hawkbit.repository.model.Artifact;
 import org.eclipse.hawkbit.repository.model.SoftwareModule;
 import org.eclipse.hawkbit.ui.artifacts.event.ArtifactDetailsEvent;
 import org.eclipse.hawkbit.ui.artifacts.event.SoftwareModuleEvent;
@@ -109,7 +118,7 @@ public class ArtifactDetailsLayout extends VerticalLayout {
 
     /**
      * Constructor for ArtifactDetailsLayout
-     * 
+     *
      * @param i18n
      *            VaadinMessageSource
      * @param eventBus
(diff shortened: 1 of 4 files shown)


schabdo:
@hawkbit-bot verify please

schabdo:
Thanks a lot for your contribution! We'll look into this and provide feedback. This will fix #786

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/artifacts/details/ArtifactDetailsLayout.java
```diff
@@ -264,19 +273,87 @@ private void addGeneratedColumnButton(final Table table) {
...
+                                String fileDownloadLoc = home + "\\Downloads\\" + artifact.get().getFilename();
+                                File outputFile = new File(fileDownloadLoc);
+                                InputStream inputStream = file.get().getFileInputStream();
+                                readChannel = Channels.newChannel(inputStream);
+                                fileOutputStream = new FileOutputStream(outputFile);
+                                writeChannel = fileOutputStream.getChannel();
+                                writeChannel.transferFrom(readChannel, 0, Long.MAX_VALUE);
+                                uINotification.displaySuccess(
+                                        i18n.getMessage("message.artifact.downloaded", fileDownloadLoc));
+                            } catch (Exception e) {
```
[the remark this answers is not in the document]
AnbazhakanSubramaniam: Both the review points are handled , Please re-review

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/artifacts/details/ArtifactDetailsLayout.java
```diff
@@ -264,19 +273,87 @@ private void addGeneratedColumnButton(final Table table) {
...
+                    if (yes) {
+                        Optional<Artifact> artifact = this.artifactManagement.get(id);
+
+                        Optional<AbstractDbArtifact> file = artifactManagement
+                                .loadArtifactBinary(artifact.get().getSha1Hash());
+                        if (file.isPresent()) {
+                            FileChannel writeChannel = null;
+                            ReadableByteChannel readChannel = null;
+                            FileOutputStream fileOutputStream = null;
+                            try {
```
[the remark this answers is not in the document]
AnbazhakanSubramaniam: "try" has been replaced with "try-with-resources"

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/artifacts/details/ArtifactDetailsLayout.java
```diff
@@ -264,19 +273,87 @@ private void addGeneratedColumnButton(final Table table) {
...
+                                        .displayValidationError(i18n.getMessage("message.artifact.downloaded.failure"));
+                            } finally {
+                                try {
+                                    if (writeChannel != null) {
+                                        writeChannel.close();
+                                    }
+                                    if (readChannel != null) {
+                                        readChannel.close();
+                                    }
+                                } catch (IOException e) {
```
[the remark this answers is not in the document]
AnbazhakanSubramaniam: Review point is handled

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/artifacts/details/ArtifactDetailsLayout.java
```diff
@@ -264,19 +273,87 @@ private void addGeneratedColumnButton(final Table table) {
...
+        downloadIcon.addClickListener(event -> downloadArtifact((Long) itemId, fileName));
+        return downloadIcon;
+    }
 
+    private void downloadArtifact(final Long id, final String fileName) {
+        final ConfirmationDialog confirmDialog = new ConfirmationDialog(
+                i18n.getMessage("caption.download.artifact.confirmbox"),
+                i18n.getMessage("message.download.artifact", fileName),
+                i18n.getMessage(UIMessageIdProvider.BUTTON_YES),
+                i18n.getMessage(UIMessageIdProvider.BUTTON_NO), yes -> {
```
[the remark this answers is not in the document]
AnbazhakanSubramaniam: The method "downloadArtifact" has been refactored.

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/artifacts/details/ArtifactDetailsLayout.java
```diff
@@ -264,19 +276,77 @@ private void addGeneratedColumnButton(final Table table) {
...
+                } catch (IOException e) {
+                    uINotification.displayValidationError(i18n.getMessage("message.artifact.download.failure"));
+                    LOG.error("Artifact reading during download caused an exception", e);
+                }
+            }
+        }
+    }
 
+    private void writeToOutputFile(String fileName, ReadableByteChannel readChannel) {
+        String fileDownloadLoc = System.getProperty("user.home") + "\\Downloads\\" + fileName;
```
laverman:
```suggestion
        final String fileDownloadLoc = System.getProperty("user.home") + File.separator + "Downloads" + File.separator + fileName;
```
The backslash as file separator will not work on unix systems.

laverman:
Hi @AnbazhakanSubramaniam , thanks for this contribution! 🙂 I just tried it out 👍  

One remark about the action flow: As a user, I would expect, that I can select to which directory the file is downloaded (analog to the upload). Would you mind extending the PR in this respect? 

There should be a Vaadin 7 functionality for file downloading available, which you could use for that. 

That would be great! 🙂

AnbazhakanSubramaniam:
Hi **@laverman** , Thanks for the review and the feedback! 🙂

I just tried using Vaadin 7 file downloading feature (FileDownloader) and found that it doesn't support "File / Location Chooser" option unlike FileUpload 😕, instead it downloads to the default "download" folder as its in my previous implementation.
The reason behind this behavior is due to the restriction of local file system access from the browsers.

Would you provide your comments on this?

bogdan-bondar:
Hi @AnbazhakanSubramaniam !

I would still suggest using the `FileDownloader` with `StreamResource` and `ErrorHandler`, so that vaadin is responsible for handling the communication with the client (browser). One of the benefits of using this approach is saving the file to the browser-configured download path (that could not be always "Download" folder, in fact on some machines this folder could be just not present). 

The second benefit is MIME-type resolution based on the file name, that could be manually overriden if needed by the `StreamResource`. 

Looking forward to your feedback on these points!

AnbazhakanSubramaniam:
Hi **@bogdan-bondar** , thanks for the feedback. Yes, i would also prefer the vaadin way of downloading (browser-configured download path). instead of letting the browser to access the file system in client machine.

**@laverman** Please provide your feedback on this to proceed further

laverman:
Hi @AnbazhakanSubramaniam - I'd agree 👍 It would be great to adapt the PR in this respect.
Thanks.

AnbazhakanSubramaniam:
Hi @laverman - Thanks for the feedback, the changes have been adapted.

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/artifacts/details/ArtifactDetailsLayout.java
```diff
@@ -264,19 +269,59 @@ private void addGeneratedColumnButton(final Table table) {
...
+
+    private Button getArtifactDownloadButton(final Table table, final Object itemId) {
+        final String fileName = (String) table.getContainerDataSource().getItem(itemId)
+                .getItemProperty(PROVIDED_FILE_NAME).getValue();
+        final Button downloadIcon = SPUIComponentProvider.getButton(
+                fileName + "-" + UIComponentIdProvider.ARTIFACT_FILE_DOWNLOAD_ICON, "",
+                i18n.getMessage(UIMessageIdProvider.ARTIFACT_DOWNLOAD), ValoTheme.BUTTON_TINY + " " + "blueicon", 
+                true, FontAwesome.DOWNLOAD, SPUIButtonStyleNoBorder.class);
+        downloadIcon.setData(itemId);
+        FileDownloader fileDownloader = new FileDownloader(createStreamResource((Long) itemId));
```
bogdan-bondar: could you please still use the confirmation dialog here rather than extending the download icon directly, so you could use previously defined ids and messages

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/artifacts/details/ArtifactDetailsLayout.java
```diff
@@ -264,19 +269,59 @@ private void addGeneratedColumnButton(final Table table) {
...
+        fileDownloader.extend(downloadIcon);
```
bogdan-bondar: Additionally could you set the error handler for FileDownloader to notify user in case of failed streaming

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/artifacts/details/ArtifactDetailsLayout.java
AnbazhakanSubramaniam: Hi @bogdan-bondar - unfortunately, there are no direct ways to bring a customized confirmation dialog along with filedownloader. Please correct me if my understanding is wrong. Anyway, the browser shows a confirmation dialog to open / save the file.

Suggested solution is as below,
	Create an 'invisiblebutton' and extend with filedownloader
	Add an action listener to the download button and invoke 'invisiblebutton.click()' on button event.

So, Shall i proceed with above recommendation? or can we stick to the default confirmation dialog from the browser? Please suggest.

bogdan-bondar: Ok, let us rely upon the confirmation dialog coming from browser. Still I would suggest adding the error handler for FileDownloader and removing the stale messages from message.properties as well as BUTTON_YES and BUTTON_NO ids

AnbazhakanSubramaniam: Thanks for the confirmation, I have added error handled for file downloader and also removed unused strings from messages.properties.

AnbazhakanSubramaniam:
This PR is no more valid as it is handled by PR #944

PR #919, opened 2019-12-04
Fix already assigned targets

sebastian-firsching:
When assigning multiple targets to a distribution set, not existing targets are silently ignored. The already assigned count was not calculated correctly in this case. This pull request is a fix for this calculation.


schabdo:
@hawkbit-bot verify please

hawkbit-repository/hawkbit-repository-jpa/src/test/java/org/eclipse/hawkbit/repository/jpa/DeploymentManagementTest.java
```diff
@@ -647,6 +648,36 @@ public void multipleAssignmentsToTargetOnlyAllowedInMultiAssignMode() {
                 deploymentManagement.assignDistributionSets(Arrays.asList(targetToDS0, targetToDS1)))).isEqualTo(2);
     }
 
+    @Test
+    @Description("Assign distribution set to not existing target.")
+    public void assignDistributionSetToNotExistingTarget() {
+        final String notExistingId = "notExistingTarget";
+        // prepare distribution set
```
sebastian-firsching: use a own method here

hawkbit-repository/hawkbit-repository-jpa/src/test/java/org/eclipse/hawkbit/repository/jpa/DeploymentManagementTest.java
```diff
@@ -647,6 +648,36 @@ public void multipleAssignmentsToTargetOnlyAllowedInMultiAssignMode() {
                 deploymentManagement.assignDistributionSets(Arrays.asList(targetToDS0, targetToDS1)))).isEqualTo(2);
     }
 
+    @Test
+    @Description("Assign distribution set to not existing target.")
```
sebastian-firsching: What is the expected behaviour?

hawkbit-repository/hawkbit-repository-jpa/src/test/java/org/eclipse/hawkbit/repository/jpa/DeploymentManagementTest.java
```diff
@@ -647,6 +648,36 @@ public void multipleAssignmentsToTargetOnlyAllowedInMultiAssignMode() {
...
+        // assign targets to DS
+        final List<DeploymentRequest> deploymentRequests = new ArrayList<>();
+        for (final String controllerId : knownTargetIds) {
+            deploymentRequests.add(new DeploymentRequest(controllerId, createdDs.getId(), ActionType.FORCED, 0, null,
+                    null, null, null));
+        }
+        final List<DistributionSetAssignmentResult> assignDistributionSetsResults = deploymentManagement
+                .assignDistributionSets(deploymentRequests);
+        for (final DistributionSetAssignmentResult assignDistributionSetsResult : assignDistributionSetsResults) {
+            assertEquals(0, assignDistributionSetsResult.getAlreadyAssigned());
```
sebastian-firsching: use assertThat instead of assertEquals

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaDeploymentManagement.java
```diff
@@ -282,16 +282,21 @@ private DistributionSetAssignmentResult assignDistributionSetToTargets(final Lon
         final List<String> targetIds = targetsWithActionType.stream().map(TargetWithActionType::getControllerId)
                 .distinct().collect(Collectors.toList());
 
+        final long notExistingTargets = targetIds.stream()
+                .filter(controllerId -> !targetRepository.existsByControllerId(controllerId)).count();
```
bogdan-bondar: The problem with this approach is that it will execute N database queries for each provided controllerId that could lead to increased latency. I would suggest filtering out non-existing controllerIds by repository itself, so at the end it looks like this:
```suggestion
                final List<String> existingTargetIds = Lists.partition(targetIds, Constants.MAX_ENTRIES_IN_STATEMENT)
                .stream().map(targetRepository::filterNonExistingControllerIds).flatMap(List::stream)
                .collect(Collectors.toList());
```
using it further and getting the total size from `existingTargetIds.size()`

The target repository implementation would then look like:
```suggestion
    @Query("SELECT t.controllerId FROM JpaTarget t WHERE t.controllerId IN ?1")
    List<String> filterNonExistingControllerIds(Iterable<String> controllerIds);
```

And finally the test should look similar to:
```suggestion
        final List<String> providedTargetIds = Lists.newArrayList("1", "notExistingTarget", "2");
        testdataFactory.createTarget("1");
        testdataFactory.createTarget("2");

        List<String> filteredIds = targetRepository.filterNonExistingControllerIds(providedTargetIds);
        
        assertThat(filteredIds).hasSize(2);
        assertThat(filteredIds).containsExactlyInAnyOrder("1", "2");
```

schabdo:
@hawkbit-bot verify please

PR #925, opened 2020-01-16
Fix empty controller name

sebastian-firsching:
When no name is provided while creating a target, the controllerId is taken as name. But the length limits for controllerId and name differ. This fix truncates the controllerId when taken as name, as the limit for the controllerId is larger than the one for the name.

### DIFF
--- a/hawkbit-rest/hawkbit-mgmt-resource/src/test/java/org/eclipse/hawkbit/mgmt/rest/resource/MgmtTargetResourceTest.java
+++ b/hawkbit-rest/hawkbit-mgmt-resource/src/test/java/org/eclipse/hawkbit/mgmt/rest/resource/MgmtTargetResourceTest.java
@@ -677,11 +677,15 @@ public class MgmtTargetResourceTest extends AbstractManagementApiIntegrationTest
 
         final String targetList = JsonBuilder.targets(Arrays.asList(target), false);
 
+        final String expectedTargetName = randomString.substring(0,
+                Math.min(JpaTarget.CONTROLLER_ID_MAX_SIZE, NamedEntity.NAME_MAX_SIZE));
+
         mvc.perform(post(MgmtRestConstants.TARGET_V1_REQUEST_MAPPING).content(targetList)
                 .contentType(MediaType.APPLICATION_JSON)).andExpect(status().isCreated())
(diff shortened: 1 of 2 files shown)


schabdo:
@hawkbit-bot verify please

hawkbit-rest/hawkbit-mgmt-resource/src/test/java/org/eclipse/hawkbit/mgmt/rest/resource/MgmtTargetResourceTest.java
```diff
@@ -667,6 +668,22 @@ public void getInstalledDistributionSetOfTargetIsEmpty() throws Exception {
...
+        final String randomString = RandomStringUtils.randomAlphanumeric(JpaTarget.CONTROLLER_ID_MAX_SIZE);
+
+        final Target target = entityFactory.target().create().controllerId(randomString).build();
+
+        final String targetList = JsonBuilder.targets(Arrays.asList(target), false);
+
+        mvc.perform(post(MgmtRestConstants.TARGET_V1_REQUEST_MAPPING).content(targetList)
+                .contentType(MediaType.APPLICATION_JSON)).andExpect(status().isCreated())
+                .andExpect(jsonPath("[0].controllerId", equalTo(randomString)))
+                .andExpect(jsonPath("[0].name", equalTo(randomString.substring(0,
```
AmmarBikic: Here you are testing the response for proper values in its body, which is fine. Would it be even better to not only check it via the response body, but additionally check the targets name value by calling the target management (so we are sure its not only the response body which looks fine)? This could look like:
`assertThat(targetManagement.getByControllerID(randomString).get().getName()).isEqualTo(randomString.substring(0, Math.min(JpaTarget.CONTROLLER_ID_MAX_SIZE, NamedEntity.NAME_MAX_SIZE)));`

By adding this, you could then also exclude `randomString.substring(0, Math.min(JpaTarget.CONTROLLER_ID_MAX_SIZE, NamedEntity.NAME_MAX_SIZE))` to a variable like `String expectedTargetName`.

sebastian-firsching: Implemented

PR #927, opened 2020-01-20
Add docker for M6

laverman:
Signed-off-by: Jeroen Laverman <jeroen.laverman@bosch-si.com>


hawkbit-runtime/docker/generator/generate.sh
```diff
@@ -9,7 +9,7 @@ set -euxo pipefail
 # http://www.eclipse.org/legal/epl-v10.html
 #
 
-HAWKBIT_VERSION="0.3.0M5"
+HAWKBIT_VERSION="0.3.0M6"
```
schabdo: Please update driver version to `2.5.3` and JRE version to `8u212-jre-alpine`

PR #928, opened 2020-01-22
Sort controller attributes alphabetically

sebastian-firsching:
With this commit, the controller attributes are sorted alphabetically in the controller detail view in the UI.

Signed-off-by: Sebastian Firsching <sebastian.firsching@bosch-si.com>

### DIFF
--- a/hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/management/targettable/TargetDetails.java
+++ b/hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/management/targettable/TargetDetails.java
@@ -8,18 +8,13 @@
  */
 package org.eclipse.hawkbit.ui.management.targettable;
 
-import com.vaadin.server.FontAwesome;
-import com.vaadin.shared.ui.label.ContentMode;
-import com.vaadin.ui.Button;
-import com.vaadin.ui.Button.ClickEvent;
-import com.vaadin.ui.Component;
-import com.vaadin.ui.HorizontalLayout;
-import com.vaadin.ui.Label;
-import com.vaadin.ui.TextField;
-import com.vaadin.ui.UI;
-import com.vaadin.ui.VerticalLayout;
-import com.vaadin.ui.Window;
-import com.vaadin.ui.themes.ValoTheme;
+import java.net.URI;
+import java.util.List;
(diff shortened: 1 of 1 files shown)


schabdo:
@hawkbit-bot verify please

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/management/targettable/TargetDetails.java
```diff
@@ -259,7 +263,10 @@ private void updateAttributesLayout(final String controllerId) {
 
     private void updateAttributesLabelsList(final VerticalLayout attributesLayout,
             final Map<String, String> attributes) {
-        for (final Map.Entry<String, String> entry : attributes.entrySet()) {
+        final List<Entry<String, String>> sortedEntries = attributes.entrySet().stream()
```
sebastian-firsching: Investigate whether it is possible to get the attributes sorted from database

dobleralex: In case DB approach won't work here are a few suggestions to improve this code section:
- Consider to put the map into a `TreeMap` constructed with a custom comparator and using the `putAll()` method to add all attributes. `TreeMaps` are always sorted by their key. This is a style decision so fully up to you, streams are also fine.
- Replace your custom lowerCase compare with the String objects `compareToIgnoreCase()`
- In case you keep the Stream use `Comparator.comparing()` to create the comparator, this saves you from retriving the keys manually, this is not necessary for the `TreeMap` method
- use the `forEach((key, value) -> ...` to iterate over entries, either on the `TreeMap` object or directly on your stream instead of the collect call.

schabdo:
@hawkbit-bot verify please

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/management/targettable/TargetDetails.java
sebastian-firsching: The database approach did not work here, as the controller attributes are saved as own table and retrieved as map. 
Instead the map is put into a TreeMap to sort the entries.

PR #931, opened 2020-02-05
Show correct notification message for single and multiple notifications.

herdt-michael:
Show correct notification message for single and multiple notifications.


sebastian-firsching:
Looks good!

PR #936, opened 2020-02-17
Feature prioritisation of campaigns ui

ramannas:
This feature contains UI enhancement for the prioritisation of campaigns. Adding a new field called Weight at the creation of an action at deployment management, rollout, and auto assignment of target filter query. 

Also the added weights at the time of creation/updation will be reflecting at the their respective tables. Weights can also be edited if needed (subjective to conditions).


schabdo:
@hawkbit-bot verify please

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaTenantConfigurationManagement.java
```diff
@@ -208,11 +213,39 @@ private static void assertMultiAssignmentsValueChange(final String key, final Jp
...
 
+    private void assertMultiAssignmentDefaultWeightValueChange(final String key,
+            final JpaTenantConfiguration valueChange) {
+        if (MULTI_ASSIGNMENTS_WEIGHT_DEFAULT.equals(key) && isMultiAssignmentEnabled() && !validateWeight(valueChange)) {
+            LOG.debug("The value {} cannot be set.", key);
+            throw new TenantConfigurationValueChangeNotAllowedException();
+        }
+    }
+
+    private boolean isMultiAssignmentEnabled() {
```
StefanKlt: since the method is called is...Enabled, I would not throw an exception but return false in case it it not

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaTenantConfigurationManagement.java
```diff
@@ -188,6 +192,7 @@
     private void assertValueChangeIsAllowed(final String key, final JpaTenantConfiguration valueChange) {
```
StefanKlt: Is the key String here really needed? Is it not always equal to the key of the JpaTenantConfiguration object? This would make the assertion methods simpler, including the newly added one

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaTenantConfigurationManagement.java
```diff
@@ -208,11 +213,39 @@ private static void assertMultiAssignmentsValueChange(final String key, final Jp
...
+
+    private boolean isMultiAssignmentEnabled() {
+        if (getConfigurationValue(MULTI_ASSIGNMENTS_ENABLED).getValue().equals(Boolean.FALSE)) {
+            LOG.debug("Cannot proceed operation because multi assignment is disabled.");
+            throw new MultiAssignmentIsNotEnabledException();
+        }
+        return true;
+    }
+
+    private boolean validateWeight(final JpaTenantConfiguration valueChange) {
```
StefanKlt: can validateWeight and validateAndParseWeight be merged to one single method?

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/utils/TenantConfigHelper.java
```diff
@@ -43,11 +45,21 @@ public static TenantConfigHelper usingContext(final SystemSecurityContext system
...
         return systemSecurityContext.runAsSystem(() -> tenantConfigurationManagement
                 .getConfigurationValue(MULTI_ASSIGNMENTS_ENABLED, Boolean.class).getValue());
     }
+
+    /**
+     * Returns the wanted configuration
+     *
+     * @return configuration value
+     */
+    public <T extends Serializable> T getConfigValue(final String key, final Class<T> valueType) {
```
StefanKlt: This only seems to be used once to get the multiassignment default weight. Maybe making it private and use it in isMultiAssignmentsEnabled and offer a public getMultiAssignmentsSefaultWeight would be better. That way we would not offer a public generic method that runs with system privileges, which might be called by mistake. (Maybe also name it getConfigValueWithSystemPrivileges)

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/utils/WeightValidationHelper.java
```diff
@@ -49,19 +49,33 @@ public static WeightValidationHelper usingContext(final SystemSecurityContext sy
...
     }
 
+    /**
+     * Validating weights associated with all the {@link DeploymentRequest}s accepting null weight
+     *
+     * @param deploymentRequests
+     *            the {@linkplain List} of {@link DeploymentRequest}s
+     */
+    public void validateAcceptNullWeight(final List<DeploymentRequest> deploymentRequests) {
+        final long assignmentsWithWeight = deploymentRequests.stream()
```
StefanKlt: code seems duplicated from validate(List<DeploymentRequest>). what aboout two static methods containsAssignmentWithWeight(...) and containsAssignmentWithWeight (...)?

schabdo:
@ramannas any updates or progress on this PR?

ramannas:
> @ramannas any updates or progress on this PR?

I should be able to pick this up right after my current tasks on Tags :)

schabdo:
Sounds great! Looking forward to finally see last missing piece for prioritisation heading towards master

PR #937, opened 2020-02-20
Rollouts view should not require users to have SYSTEM_ADMIN permission

herdt-michael:
Read tenant configuration value from system context to prevent permission issues while loading the rollouts view.

Signed-off-by: Michael Herdt <Michael.Herdt2@bosch-si.com>


schabdo:
@hawkbit-bot verify please

PR #943, opened 2020-03-04
Add spring exception handler for FileStreamingFailedException

sebastian-firsching:
This pull request adds a new exception handler for FileStreamingException that will be called by Spring-Framework. As the file streaming has already begun in case the FileStreamingException occurs, a Json response can't be returned. This is the reason why the FileStreamingException needs its own exception handler.

Signed-off-by: Sebastian Firsching <sebastian.firsching@bosch-si.com>


schabdo:
@hawkbit-bot verify please

PR #945, opened 2020-03-06
Streamline hawkBit logout behaviour (OIDC vs. Credentials) 

Nkyn:
When oidc configurations are set in the properties file, currently the auth provider login screen is shown even when logged in with a basic auth user.

Instead, even when oidc is configured, the user should be redirected to the same login page he previously used to login.


schabdo:
@hawkbit-bot verify please

schabdo:
@hawkbit-bot verify please

PR #946, opened 2020-03-06
Show username in tables

Nkyn:
Currently the user who created objects in UI tables (e.g. targets, target filters) are displayed inconsistently, depending on which user management mechanism is used (e.g. `spring.security.user`, `hawkbit.server.im.users[0]`) and which user is logged in. 
The username or loginname however is a consistent value that's always provided and used to login. 
Thus this username is now shown in the tables' Created By / Modified By columns instead of firstname, lastname.


schabdo:
@hawkbit-bot verify please

PR #947, opened 2020-03-12
Fix xss remained parts

AmmarBikic:
Through analysis of the SnakeYAML parser vulnerability, some remained Cross Site Scripting relevant tooltips were found. This tooltips were disabled by this pull request, as they are always handled as rendered rich text. It is planned to enable them again after Vaadin 8 migration, as tooltips will provide content type settings.


schabdo:
@hawkbit-bot verify please

PR #949, opened 2020-03-17
Add permissions with roles for spring.security users.

herdt-michael:
Add assigned permissions with roles spring.security users, instead of assigning all permissions each time without roles.

Signed-off-by: Michael Herdt <michael.herdt2@bosch-si.com>

### DIFF
--- a/hawkbit-autoconfigure/src/main/java/org/eclipse/hawkbit/autoconfigure/security/InMemoryUserManagementAutoConfiguration.java
+++ b/hawkbit-autoconfigure/src/main/java/org/eclipse/hawkbit/autoconfigure/security/InMemoryUserManagementAutoConfiguration.java
@@ -93,11 +93,26 @@ public class InMemoryUserManagementAutoConfiguration extends GlobalAuthenticatio
         if (userPrincipals.isEmpty()) {
             final String name = securityProperties.getUser().getName();
             final String password = securityProperties.getUser().getPassword();
-            userPrincipals.add(new UserPrincipal(name, password, name, name, name, null, DEFAULT_TENANT,
-                    PermissionUtils.createAllAuthorityList()));
+            List<GrantedAuthority> authorityList;
+            if (securityProperties.getUser().getRoles().isEmpty()) {
+                authorityList = PermissionUtils.createAllAuthorityList();
+            } else {
+                authorityList = createAuthoritiesFromList(securityProperties.getUser().getRoles());
+            }
+            userPrincipals
+                    .add(new UserPrincipal(name, password, name, name, name, null, DEFAULT_TENANT, authorityList));
         }
 
         return new FixedInMemoryUserPrincipalUserDetailsService(userPrincipals);
+    }
+
+    private static List<GrantedAuthority> createAuthoritiesFromList(final List<String> userAuthorities) {
+        List<GrantedAuthority> grantedAuthorityList = new ArrayList<>(userAuthorities.size());
+        for (final String permission : userAuthorities) {
+            grantedAuthorityList.add(new SimpleGrantedAuthority(permission));
+            grantedAuthorityList.add(new SimpleGrantedAuthority("ROLE_" + permission));
+        }
+        return grantedAuthorityList;
     }
 
     private static class FixedInMemoryUserPrincipalUserDetailsService implements UserDetailsService {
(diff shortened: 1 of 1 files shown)


schabdo:
@hawkbit-bot verify please

hawkbit-autoconfigure/src/main/java/org/eclipse/hawkbit/autoconfigure/security/InMemoryUserManagementAutoConfiguration.java
```diff
@@ -93,13 +93,28 @@ UserDetailsService userDetailsService() {
         if (userPrincipals.isEmpty()) {
             final String name = securityProperties.getUser().getName();
             final String password = securityProperties.getUser().getPassword();
-            userPrincipals.add(new UserPrincipal(name, password, name, name, name, null, DEFAULT_TENANT,
-                    PermissionUtils.createAllAuthorityList()));
+            List<GrantedAuthority> authorityList;
+            if (securityProperties.getUser().getRoles().isEmpty()) {
```
schabdo: Better extract `securityProperties.getUser().getRoles()` to a local variable and put it next to the username and password variable. Than your if-statement looks cleaner:

```java
if(roles.isEmpty()){
   authorityList = PermissionUtils.createAllAuthorityList();
} else {
   authorityList = createAuthoritiesFrom(roles);
} 
```

hawkbit-autoconfigure/src/main/java/org/eclipse/hawkbit/autoconfigure/security/InMemoryUserManagementAutoConfiguration.java
```diff
@@ -93,13 +93,28 @@ UserDetailsService userDetailsService() {
...
+            }
+            userPrincipals
+                    .add(new UserPrincipal(name, password, name, name, name, null, DEFAULT_TENANT, authorityList));
         }
 
         return new FixedInMemoryUserPrincipalUserDetailsService(userPrincipals);
     }
 
+    private static List<GrantedAuthority> createAuthoritiesFromList(final List<String> userAuthorities) {
+        List<GrantedAuthority> grantedAuthorityList = new ArrayList<>(userAuthorities.size());
```
schabdo: This looks almost the same as above (lines 78-82) can't you reuse same method there somehow as well?

hawkbit-autoconfigure/src/main/java/org/eclipse/hawkbit/autoconfigure/security/InMemoryUserManagementAutoConfiguration.java
```diff
@@ -93,13 +90,29 @@ UserDetailsService userDetailsService() {
...
             final String password = securityProperties.getUser().getPassword();
-            userPrincipals.add(new UserPrincipal(name, password, name, name, name, null, DEFAULT_TENANT,
-                    PermissionUtils.createAllAuthorityList()));
+            final List<String> roles = securityProperties.getUser().getRoles();
+            List<GrantedAuthority> authorityList;
+            if (roles.isEmpty()) {
+                authorityList = PermissionUtils.createAllAuthorityList();
+            } else {
+                authorityList = createAuthoritiesFromList(roles);
+            }
```
a-sayyed: for unary conditions like this, I prefer to use the conditional operator:
```
List<GrantedAuthority> authorityList = roles.isEmpty() ? PermissionUtils.createAllAuthorityList() : createAuthoritiesFromList(roles);
```
But that's of course a coding style and in no way a requirement ;)

hawkbit-autoconfigure/src/main/java/org/eclipse/hawkbit/autoconfigure/security/InMemoryUserManagementAutoConfiguration.java
a-sayyed: Isn't it the case that 'previous' lines 78-82 were refactored to be this method?

schabdo: @herdt-michael was really fast by doing this

PR #951, opened 2020-03-23
Fix treshold column headers

AmmarBikic:
This pull request provides small adaptations in the threshold column headers inside the Rollout view to provide a better understanding of the column values, as they contain percentage values instead of absolute ones.


schabdo:
@hawkbit-bot verify please

PR #959, opened 2020-04-16
Fix NPE issues in RSQLUtility

dobleralex:
This adds a check to ensure the field path of a RSQL query is not null, otherwise throw an exception that can be used to display an error.


laverman:
@hawkbit-bot please verify

PR #961, opened 2020-04-20
Feature mitigate splunk dmflogs

ramannas:
Spunk DMF logs were getting polluted with unnecessary stack traces because of invalid target attributes. To fix the issue, a try-catch has been added to handle the unnecessary logs and warn message has been added.

### DIFF
--- a/hawkbit-dmf/hawkbit-dmf-amqp/src/main/java/org/eclipse/hawkbit/amqp/AmqpMessageHandlerService.java
+++ b/hawkbit-dmf/hawkbit-dmf-amqp/src/main/java/org/eclipse/hawkbit/amqp/AmqpMessageHandlerService.java
@@ -10,7 +10,6 @@ package org.eclipse.hawkbit.amqp;
 
 import static org.eclipse.hawkbit.repository.RepositoryConstants.MAX_ACTION_COUNT;
 import static org.eclipse.hawkbit.tenancy.configuration.TenantConfigurationProperties.TenantConfigurationKey.MULTI_ASSIGNMENTS_ENABLED;
-import static org.springframework.util.StringUtils.hasText;
 
 import java.io.Serializable;
 import java.net.URI;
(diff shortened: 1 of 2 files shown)


hawkbit-dmf/hawkbit-dmf-amqp/src/main/java/org/eclipse/hawkbit/amqp/AmqpMessageHandlerService.java
```diff
@@ -324,10 +323,15 @@ private void updateAttributes(final Message message) {
         final DmfAttributeUpdate attributeUpdate = convertMessage(message, DmfAttributeUpdate.class);
         final String thingId = getStringHeaderKey(message, MessageHeaderKey.THING_ID, THING_ID_NULL);
 
+        try {
         controllerManagement.updateControllerAttributes(thingId, attributeUpdate.getAttributes(),
                 getUpdateMode(attributeUpdate));
+        } catch (final InvalidTargetAttributeException ex) {
```
schabdo, line 329: If you catch this exception here the DMF message sent via AMQP will be acknowledged and silently "ignored" instead of rejected into dead letter queue. So publisher (in this case the user of DMF API) won't get any feedback that attributes are invalid. Not sure if this is intended

ramannas: If so, are we obligated to wrap this exception into AmqpRejectAndDontRequeueException with intended message, so that the publisher is notified and also the message is rejected into dead letter queue?

schabdo: Wrapping the exception would only address this specific call to update attributes, it won't do you any good for other exceptions / other calls somewhere else in the code base. So you will still see logs polluted from different origin. In my opinion it's better to fix the root cause of the log message. I assume that `ConditionalRejectingErrorHandler` is the source of log statement you want to suppress, right?

ramannas:
Closing the PR as there has been no progress on it from a long time. Shall create new one once we have a correct solution.

Issue #916, opened 2019-11-20
"Custom Filter" list doesn't update after creating new filter


Informatic:
Hey, just a quick UI-related bug report.

After creating a new filter on "Target Filters" page, I don't see a newly created filter on "Custom Filter" side of "Deployment" page, unless I refresh the page.

schabdo:
Thanks for filing an issue for that. Indeed custom filters are not displayed at the "Deployment" page right away after they have been created. It seems that the custom filter component is the only part of this page which is not yet push enabled (by means automatically refreshed on entity changes)

schabdo:
Thanks for tackling this bug, highly appreciated! 

@bogdan-bondar may I asked you to have a look at the PR and provide feedback?

PR #970, opened 2020-06-05
Custom Filter refreshed in management page - #916

ajaybalakumar:
Add, modify and remove events handled for custom filter and the
management page is updated accordingly
https://github.com/eclipse/hawkbit/issues/916

Signed-off-by: Ajay Balakumar Jeyabalakrishnan <ajaybalakumar.jeyabalakrishnan@in.bosch.com>


schabdo:
@hawkbit-bot verify please

schabdo:
Thanks for tackling this bug, highly appreciated!

@bogdan-bondar may I asked you to have a look at the PR and provide feedback?

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/management/targettable/TargetTable.java
```diff
@@ -219,6 +220,12 @@ void onEvent(final TargetFilterEvent filterEvent) {
 
     @EventBusListenerMethod(scope = EventScope.UI)
     void onEvent(final ManagementUIEvent managementUIEvent) {
+      if(managementUIEvent == ManagementUIEvent.REFRESH_TARGETS_ON_FILTER_UPDATE)
```
bogdan-bondar, line 223: you need to first check if the table is filtered by the custom filter at all, otherwise you will unnecessarily refresh the table each time you modify some custom target filter elsewhere. I would also execute all of the logic in `UI.getCurrent().access` scope and combine tags conditionals with custom filter conditional (see above) and for both send the _ManagementUIEvent.TARGET_TABLE_FILTER_ event in order to update _CountMessageLabel_

PR #989, opened 2020-07-31
Added config setting for attribute search.

Sergey-G-dev:
Currently, when running a target search by text, the target attributes are also subject to search. This slows down the search significantly if there is a large number of targets and their attributes. It is also not needed when searching by target ID. So I've added a configuration option in the configuration dashboard to exclude the target attributes to reduce the database load and speed up the search.


Sergey-G-dev:
CircleCI complains about missing license headers, although I've added and double-checked the headers and those seem to be correct. Can anyone check?

[WARNING] Missing header in: /home/circleci/hawkBit/hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/tenantconfiguration/search/TargetSearchConfigurationItem.java
[WARNING] Missing header in: /home/circleci/hawkBit/hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/tenantconfiguration/TargetSearchConfigurationView.java

PR #998, opened 2020-08-24
Make use of GitHub actions

ctron:
This PR allows to use GitHub actions in additions to CircleCI.

Everyone already has GitHub actions when using GitHub. So this makes it easier for people to contribute. As CircleCI forces you into creating an account and granting full access to all public repositories, just for viewing the test failures.

Signed-off-by: Jens Reimann <jreimann@redhat.com>

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,0 +1,31 @@
+name: Java CI
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Set up JDK 1.8
+        uses: actions/setup-java@v1
+        with:
+          java-version: 1.8
+
+      - name: Cache local Maven repository
+        uses: actions/cache@v2
+        with:
+          path: ~/.m2/repository
+          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+          restore-keys: |
+            ${{ runner.os }}-maven-
+
+      - name: Build with Maven
+        run: mvn verify license:check javadoc:javadoc --batch-mode
+
+      - name: Publish Test Report
+        uses: scacap/action-surefire-report@v1
+        with:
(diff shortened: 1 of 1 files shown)


floruschbaschan:
Thanks a lot for your contribution to Hawkbit (#998). We will look at it and then give feedback. But please keep in mind that currently many colleagues are having vacation so it might take it little bit longer as usual.

schabdo:
Thanks a lot @ctron, highly appreciated! Indeed GitHub actions don't require anything and are well integrated. So for sure we could add them additionally to CircleCI and gain some experience. Depending of the outcome of [this discussion](https://discuss.circleci.com/t/authenticate-with-docker-to-avoid-impact-of-nov-1st-rate-limits) they may become our default CI very fast :)

**I do have only one remark:** Since `AmqpAuthenticationMessageHandlerIntegrationTest` tries to connect to a RabbitMQ broker (which is available on CircleCI) we will encounter errors in the output of Surefire. These errors are shown as "Check failure" annotations at the end of each action run, however build will succeed. FMPOV it would be cool to have this fixed before we introduce GitHub actions. From my own experience false-positives often lower the acceptance rate of something new :)

Long story short: I'll try to fix the mentioned test somehow but this will need more time. Meanwhile if somebody could lend a hand here this would awesome!

schabdo:
I think I found a better way to get the mentioned issue addressed. Luckily there is a RabbitMQ action available which could be leveraged to execute the AMQP integration tests in a proper manner and consider the test results within a build (like we do on CircleCI). This should do the trick:

```
- uses: nijel/rabbitmq-action@v1.0.0
        with:
          rabbitmq version: '3.7-alpine'
```

@ctron having the RabbitMQ in place we're good to go for master. May I ask you to add this to this PR?

diegorondini:
@ctron are you able to implement the requested changes?

ctron:
> @ctron are you able to implement the requested changes?

Right now I am a bit busy with other stuff. However, feel free to amend the PR.

avgustinmm:
outdated

PR #1007, opened 2020-09-09
Add Update Factory by Kynetics as hawkbit provider

diegorondini:
Signed-off-by: Diego Rondini <diego.rondini@kynetics.com>


schabdo:
Good catch @diegorondini. The link is already part of the GitHub `readme.md` but missing in the docs. Thanks for adding.

PR #1013, opened 2020-09-23
Actions created by Autoassignment or rollout should have the value createdBy of the user

a-sayyed:


hawkbit-core/src/main/java/org/eclipse/hawkbit/tenancy/TenantAware.java
```diff
@@ -39,6 +39,8 @@
      */
     <T> T runAsTenant(final String tenant, TenantRunner<T> tenantRunner);
 
+    <T> T runAsTenantAsUser(String tenant, String username, TenantRunner<T> tenantRunner);
```
a-sayyed, line 42: add description

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/RepositoryApplicationConfiguration.java
```diff
@@ -753,15 +753,17 @@ EventEntityManager eventEntityManager(final TenantAware aware, final EntityManag
...
+     * @param systemSecurityContext
+     *            the {@link SystemSecurityContext}
      * @return a new {@link AutoAssignChecker}
      */
     @Bean
-    @ConditionalOnMissingBean
-    AutoAssignExecutor autoAssignExecutor(final TargetFilterQueryManagement targetFilterQueryManagement,
-                                         final TargetManagement targetManagement, final DeploymentManagement deploymentManagement,
-                                         final PlatformTransactionManager transactionManager) {
+    @ConditionalOnMissingBean AutoAssignExecutor autoAssignExecutor(
```
a-sayyed, line 761: wrong format

hawkbit-security-core/src/main/java/org/eclipse/hawkbit/security/SystemSecurityContext.java
```diff
@@ -117,6 +117,28 @@ public SystemSecurityContext(final TenantAware tenantAware) {
         }
     }
 
+    public <T> T runAsSystemAsTenantAsUser(final Callable<T> callable, final String tenant, final String username) {
```
a-sayyed, line 120: add description

hawkbit-autoconfigure/src/main/java/org/eclipse/hawkbit/autoconfigure/security/OidcUserManagementAutoConfiguration.java
```diff
@@ -193,7 +193,7 @@ public void onAuthenticationSuccess(final HttpServletRequest request, final Http
             final AbstractAuthenticationToken token = (AbstractAuthenticationToken) authentication;
             token.setDetails(new TenantAwareAuthenticationDetails(defaultTenant, false));
 
-            systemSecurityContext.runAsSystemAsTenant(systemManagement::getTenantMetadata, defaultTenant);
+            systemSecurityContext.runAsSystemAsTenantAsUser(systemManagement::getTenantMetadata, defaultTenant, authentication.getName());
```
schabdo, line 196: I'm not sure if this is correct here. Since this will bootstrap the initial setup from my point of view it should not be audited with the name of the user which logged in the first time. It's a pure technical one time setup thus it rather belongs to the "system" user

schabdo: Since there is another context switch in `JpaSystemManagement#createInitialTenantMetaData` back to the tenant I think `runAsSystemAsTenantAsUser` is not required here

hawkbit-security-core/src/main/java/org/eclipse/hawkbit/security/SystemSecurityContext.java
schabdo, line 120: Not sure if this is intended or the name is just a bit misleading: This method will give the provided username "superpower" by means all permissions! It is not running with the permissions assigned to the user itself

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/autoassign/AutoAssignChecker.java
```diff
@@ -146,7 +152,8 @@ private int runTransactionalAssignment(final TargetFilterQuery targetFilterQuery
                             targetFilterQuery.getAutoAssignWeight().orElse(null), PAGE_SIZE);
                     final int count = deploymentRequests.size();
                     if (count > 0) {
-                        deploymentManagement.assignDistributionSets(deploymentRequests, actionMessage);
+                        systemSecurityContext.runAsSystemAsTenantAsUser(() -> deploymentManagement.assignDistributionSets(deploymentRequests, actionMessage),
```
schabdo, line 155: Not sure if this is intended: With this change you won't be able to distinguish between a manual assignment and an automatic assignment from the auto scheduler. It will always be the poor guy who created the filter

a-sayyed: that is true, do you think it makes sense to prepend something like `autoassign:` or something like that to the username?

a-sayyed: would it make sense to track the enabling of the autoassignment, something like an extra field `auto_assign_created_by` and use that user instead of the creator of the filter?

hawkbit-security-core/src/main/java/org/eclipse/hawkbit/security/SystemSecurityContext.java
a-sayyed: you're right, maybe a better approach is to just `runAsTenantAsUser` and read the permissions for that user from somewhere. The system scope is not needed I think

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/autoassign/AutoAssignChecker.java
a-sayyed: what do you think about creating a field `auto_assign_triggered_by` or `auto_assign_enabled_by` in `TargetFilterQuery` table, and `action_triggered_by` filed in `Action` table?

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaRolloutManagement.java
```diff
@@ -837,29 +837,30 @@ private long executeFittingHandler(final long rolloutId) {
         LOGGER.debug("handle rollout {}", rolloutId);
         final JpaRollout rollout = rolloutRepository.findById(rolloutId)
                 .orElseThrow(() -> new EntityNotFoundException(Rollout.class, rolloutId));
+        return systemSecurityContext.runAsSystemAsTenantAsUser(() -> {
```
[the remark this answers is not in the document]
a-sayyed: what happens in the case/state of `waitingForApproval` or it was 'auto-approved'? we need to verify that

a-sayyed:
closing this PR in favor of https://github.com/eclipse/hawkbit/pull/1017/files

PR #1017, opened 2020-09-24
added triggeredBy for rollouts and autoassignment

a-sayyed:
To improve traceability of created actions, it should be possible to track who triggered an action (directly or indirectly).
This PR adds a new field to the Action entity that corresponds to the user who initiated/triggered an assignment, via manual/auto assignment or via a Rollout.


hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaRolloutManagement.java
```diff
@@ -624,6 +624,7 @@ private void createScheduledAction(final Collection<Target> targets, final Distr
             action.setStatus(Status.SCHEDULED);
             action.setRollout(rollout);
             action.setRolloutGroup(rolloutGroup);
+            action.setTriggeredBy(rollout.getCreatedBy());
```
a-sayyed, line 627: todo: verfiy when rollout is in state waiting approval, who should be the trigggerer here?

schabdo:
@hawkbit-bot verify please

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaDeploymentManagement.java
```diff
@@ -417,12 +428,12 @@ private void setAssignedDistributionSetAndTargetUpdateStatus(final AbstractDsAss
...
 
-    private List<JpaAction> createActions(final Collection<TargetWithActionType> targetsWithActionType,
-            final List<JpaTarget> targets, final AbstractDsAssignmentStrategy assignmentStrategy,
-            final JpaDistributionSet set) {
+    private List<JpaAction> createActions(final String triggeredBy, final Collection<TargetWithActionType> targetsWithActionType,
+            final List<JpaTarget> targets, final AbstractDsAssignmentStrategy assignmentStrategy, final JpaDistributionSet set) {
 
         return targetsWithActionType.stream().map(twt -> assignmentStrategy.createTargetAction(twt, targets, set))
-                .filter(Objects::nonNull).map(actionRepository::save).collect(Collectors.toList());
+                .filter(Objects::nonNull).peek(a -> a.setTriggeredBy(triggeredBy))
```
schabdo, line 435: Why don't you set `triggeredBy` within the strategy? Modification after action has been created is not a good idea if someone else uses this method

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/model/JpaAction.java
```diff
@@ -137,6 +137,9 @@
     @Column(name = "external_ref", length = Action.EXTERNAL_REF_MAX_LENGTH)
     private String externalRef;
 
+    @Column(name = "triggered_by", updatable = false, nullable = false, length = 64)
```
schabdo, line 140: Shouldn't this rather be `updatable = true`?

a-sayyed: no the action should have only 1 triggerer, why should this value be updatable?

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/model/JpaTargetFilterQuery.java
```diff
@@ -77,6 +77,9 @@
     @Column(name = "auto_assign_weight", nullable = true)
     private Integer autoAssignWeight;
 
+    @Column(name = "auto_assign_triggered_by", nullable = true, length = 64)
+    private String autoAssignTriggeredBy;
```
schabdo, line 81: Not sure if this is a good wording here. The trigger is the scheduler/system. Isn't more about who configured the auto assignment?

schabdo: @bogdan-bondar do you have a better suggestion?

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/model/JpaAction.java
schabdo: Indeed, you're right. The action won't be changed after its creation

schabdo:
What about tests if the field is filled correctly by different use cases?

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/model/JpaTargetFilterQuery.java
a-sayyed: would `auto_assign_enabled_by` fit more here?

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaDeploymentManagement.java
```diff
@@ -182,33 +182,41 @@ protected JpaDeploymentManagement(final EntityManager entityManager, final Actio
...
 
-        return assignDistributionSets(deploymentRequests, null, offlineDsAssignmentStrategy);
+        return assignDistributionSets(tenantAware.getCurrentUsername(), deploymentRequests, null,
+                offlineDsAssignmentStrategy);
     }
 
     @Override
     @Transactional(isolation = Isolation.READ_COMMITTED)
-    public List<DistributionSetAssignmentResult> assignDistributionSets(
+    public List<DistributionSetAssignmentResult> assignDistributionSets(final String triggeredBy,
```
schabdo, line 191: So I miss something or is this method never called from outside? ... just by the one below without the `triggeredBy ` in the signature?

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaDeploymentManagement.java
```diff
@@ -182,33 +182,41 @@ protected JpaDeploymentManagement(final EntityManager entityManager, final Actio
                 .map(entry -> DeploymentManagement.deploymentRequest(entry.getKey(), entry.getValue()).build())
                 .collect(Collectors.toList());
 
-        return assignDistributionSets(deploymentRequests, null, offlineDsAssignmentStrategy);
+        return assignDistributionSets(tenantAware.getCurrentUsername(), deploymentRequests, null,
```
schabdo, line 185: I think `assignDistributionSets(deploymentRequests, null, offlineDsAssignmentStrategy)` should work as well

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaDeploymentManagement.java
a-sayyed: yes, you're right, will make it private

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaRolloutManagement.java
a-sayyed: I still believe that the triggerer here would be the creator. If I want to do something but have to wait for approval, I'm still the triggerer of the action, not the approver. Without the creator, there would be no action

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaDeploymentManagement.java
```diff
@@ -182,33 +182,39 @@ protected JpaDeploymentManagement(final EntityManager entityManager, final Actio
...
 
     @Override
     @Transactional(isolation = Isolation.READ_COMMITTED)
     public List<DistributionSetAssignmentResult> assignDistributionSets(
             final List<DeploymentRequest> deploymentRequests) {
-        return assignDistributionSets(deploymentRequests, null);
+        return assignDistributionSets(tenantAware.getCurrentUsername(), deploymentRequests);
+    }
+
+    private List<DistributionSetAssignmentResult> assignDistributionSets(final String triggeredBy,
```
a-sayyed, line 196: inline method

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/model/JpaTargetFilterQuery.java
bogdan-bondar: `autoAssignInitiatedBy`? Same can be used for the action

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaDeploymentManagement.java
bogdan-bondar: If you mark it as private I assume you would also want to remove the @Transactional annotation because it will be not taken into consideration by Spring proxy. Nevertheless you can wrap it in explicit transaction if you think it is required here

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaRolloutManagement.java
a-sayyed: @bogdan-bondar what do you think?

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/model/JpaTargetFilterQuery.java
a-sayyed: sounds good for me, do you agree @schabdo ?

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaTargetFilterQueryManagement.java
```diff
@@ -248,8 +249,10 @@ public TargetFilterQuery updateAutoAssignDS(final AutoAssignDistributionSetUpdat
             // auto-assign distribution set when creating a target filter query
             assertMaxTargetsQuota(targetFilterQuery.getQuery());
             final JpaDistributionSet ds = findDistributionSetAndThrowExceptionIfNotFound(update.getDsId());
+            final String initiatedBy = TenantAwareHolder.getInstance().getTenantAware().getCurrentUsername();
```
bogdan-bondar, line 252: I would inject the TenantAware directly via constructor injection here. Additionally I would inline the call into #setAutoAssignInitiatedBy

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/autoassign/AutoAssignChecker.java
```diff
@@ -146,7 +147,10 @@ private int runTransactionalAssignment(final TargetFilterQuery targetFilterQuery
                             targetFilterQuery.getAutoAssignWeight().orElse(null), PAGE_SIZE);
                     final int count = deploymentRequests.size();
                     if (count > 0) {
-                        deploymentManagement.assignDistributionSets(deploymentRequests, actionMessage);
+                        final String initiatedBy = StringUtils.isEmpty(targetFilterQuery.getAutoAssignInitiatedBy()) ?
```
bogdan-bondar, line 150: Could you please move it to dedicated private static method?

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/model/JpaAction.java
```diff
@@ -137,6 +137,9 @@
     @Column(name = "external_ref", length = Action.EXTERNAL_REF_MAX_LENGTH)
     private String externalRef;
 
+    @Column(name = "initiated_by", updatable = false, nullable = false, length = 64)
```
bogdan-bondar, line 140: If you set this field as non-nullable, what happens to already existing actions in the DB, is a migration script required? Could you also define the length of this field in `Action` interface

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/model/JpaTargetFilterQuery.java
```diff
@@ -77,6 +77,9 @@
     @Column(name = "auto_assign_weight", nullable = true)
     private Integer autoAssignWeight;
 
+    @Column(name = "auto_assign_initiated_by", nullable = true, length = 64)
```
bogdan-bondar, line 80: could you move the length value to the `TargetFilterQuery` interface

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/model/JpaAction.java
a-sayyed: no migration needed for old actions, since this value is not updatable
I also tested it with the following code path, works without issues
```java
@Bean
ApplicationRunner testOldActions(final TenantAware tenantAware, final ActionRepository actionRepository) {
        return args -> {
            tenantAware.runAsTenant("DEFAULT", () -> {
                actionRepository.findAll().forEach(a -> {
                    if (a.isActive()) {
                        a.setActionType(Action.ActionType.DOWNLOAD_ONLY);
                        actionRepository.save(a);
                    }
                });
                return null;
            });
        };
    }
```

hawkbit-repository/hawkbit-repository-jpa/src/main/resources/db/migration/SQL_SERVER/V1_12_16__add_action_initiated_by___SQL_SERVER.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE sp_action ADD initiated_by VARCHAR(64) NOT NULL;
+ALTER TABLE sp_target_filter_query ADD COLUMN auto_assign_initiated_by VARCHAR(64);
```
bogdan-bondar, line 2: I guess it should be ADD instead of ADD COLUMN for SQL Server, could you check it?

a-sayyed: good catch!

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/model/JpaTargetFilterQuery.java
schabdo: Yes, thanks @bogdan-bondar

Issue #983, opened 2020-07-29
Optimization of AutoAssignment Checks


Sergey-G-dev:
Hi all,

in our organization we are using hawkBit to manage a large number of target devices (currently around 1,2 Million and the number is growing). We want to make sure that every time a new target device connects and registers, the latest firmware is provided to that device as quickly as possible. To achieve that, we run the AutoAssigment Checker with a time interval of 3 minutes. This behavior is not optimal for us, because, on one hand, the customer has to wait up to 3 minutes to get the update. On the other hand, having that many targets saved in the database and executing TargetFilterQueries every 3 minutes causes a constantly high database load. To improve this, we would like to schedule a regular AutoAssigment Check only once a day, but whenever a target uploads its attributes to hawkBit, the AutoAssigment Check would be triggered only for that one specific target.

So it would go like this: target pushes its "configData to hawkBit", hawkBit fetches the list of the TargetFilterQueries with assigned DistributionSet from the database, then tries to match the properties from the TFQ-string against the target's attributes. If it finds a match, then the corresponding DistributionSet would be assigned to that target.

So far I haven't found any way to match the TFQ directly against the JpaTarget's properties, without querying the database. One of the possible ways to do this would be to implement some kind of a matcher that would parse the TFQs and try to match it against the given JpaTarget. Do you think that would be feasible? Or would you recommend a different approach?

Thanks and best regards,
Sergey

KONGDejing:
I want to ask a question, If i want to install hawkbit , what the detailed steps are for the new gay.  I search for the internet and find nothing about this.

PR #1024, opened 2020-10-09
Trigger Distribution Set assignment check on updating target config data

Sergey-G-dev:
This is an attempt to solve the issue described here: https://github.com/eclipse/hawkbit/issues/983

Summary: whenever a target uploads its configuration over REST endpoint '.../configData' a check for Distribution Set assignment is triggered based on existing Target Filters. This makes AutoAssignmentChecker and Scheduler obsolete. The database load is reduced significantly since there is no need to execute the assignment check regularly.

Signed-off-by: Sergey Gerasimov <sergey.gerasimov@devolo.de>

### DIFF
--- a/hawkbit-dmf/hawkbit-dmf-amqp/src/test/java/org/eclipse/hawkbit/integration/AmqpMessageHandlerServiceIntegrationTest.java
+++ b/hawkbit-dmf/hawkbit-dmf-amqp/src/test/java/org/eclipse/hawkbit/integration/AmqpMessageHandlerServiceIntegrationTest.java
@@ -924,12 +924,12 @@ public class AmqpMessageHandlerServiceIntegrationTest extends AbstractAmqpServic
         assertThat(target).isPresent();
 
         // verify the DS was assigned to the Target
-        final DistributionSet assignedDistributionSet = ((JpaTarget) target.get()).getAssignedDistributionSet();
+        final DistributionSet assignedDistributionSet = target.get().getAssignedDistributionSet();
         assertThat(assignedDsId).isNotNull();
         assertThat(assignedDistributionSet.getId()).isEqualTo(assignedDsId);
 
         // verify that the installed DS was not affected
-        final JpaDistributionSet installedDistributionSet = ((JpaTarget) target.get()).getInstalledDistributionSet();
+        final DistributionSet installedDistributionSet = target.get().getInstalledDistributionSet();
         if (installedDsId == null) {
             assertThat(installedDistributionSet).isNull();
         } else {
--- a/hawkbit-repository/hawkbit-repository-api/src/main/java/org/eclipse/hawkbit/repository/ControllerManagement.java
+++ b/hawkbit-repository/hawkbit-repository-api/src/main/java/org/eclipse/hawkbit/repository/ControllerManagement.java
@@ -385,6 +385,9 @@ public interface ControllerManagement {
     Target updateControllerAttributes(@NotEmpty String controllerId, @NotNull Map<String, String> attributes,
             UpdateMode mode);
 
+    @PreAuthorize(SpringEvalExpressions.IS_CONTROLLER)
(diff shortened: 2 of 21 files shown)


schabdo:
@hawkbit-bot verify please

bogdan-bondar:
Thanks a lot for your contribution! I have one remark though: "This makes AutoAssignmentChecker and Scheduler obsolete", not really. Target Filter isn't scoped to Target attributes only, this involves Target name, metadata, description, etc. as well. So as an example: if new Target matching the Filter (say by name) comes online and being registered in Hawkbit per plug-and-play mechanism it should receive the DS assignment irregardless of whether it sends attributes or not.

Sergey-G-dev:
You are correct, @bogdan-bondar . My choice of words wasn't very precise. By "attributes" I meant not only the actual attributes, but all the relevant fields as defined in the class `TargetFields.java`, since those are used by the filters. Still, I don't see how the AutoAssignmentScheduler/Checker is usefull after this change, because it does exactly the same thing, just at a scheduled time interval.

Sergey-G-dev:
@hawkbit-bot verify please

schabdo:
@hawkbit-bot verify please

schabdo:
@hawkbit-bot verify please

Sergey-G-dev:
After rebasing this branch on top of the current master, CircleCI complains about 2 failed Unit/Integration tests, but the log messages make no sense to me:

```
[ERROR]   ControllerManagementTest.controllerReportsActionFinishedForDownloadOnlyActionThatIsNotActive Did not receive the expected amount of events form class org.eclipse.hawkbit.repository.event.remote.entity.TargetUpdatedEvent Expected: 3 but was: 3
[ERROR]   ControllerManagementTest.updateTargetAttributesWithDifferentUpdateModes Did not receive the expected amount of events form class org.eclipse.hawkbit.repository.event.remote.entity.TargetCreatedEvent Expected: 1 but was: 1
```

Running the build on my local machine works fine with all tests passing. Any ideas?

schabdo:
> Running the build on my local machine works fine with all tests passing. Any ideas?

I triggered the build 5 times in a row (cf. [PR builds](https://app.circleci.com/pipelines/github/eclipse/hawkbit?branch=pull%2F1024)) without error. I assume that this was some kind of flakiness on CircleCI side

bogdan-bondar:
Hi @Sergey-G-dev ! Eventually I got the chance to take a look at your PR, sorry for the delay. I really like your idea of checking if there is an assignment available based on the Target change rather than querying all the targets periodically for each target filter. However, from my point of view, the current implementation is not quite there yet. 

First of all, you trigger the assignment check **only** when targets update their attributes: `DdiRootController#putConfigData -> controllerManagement.triggerDistributionSetAssignmentCheck`, but what if a target does not send its attributes at all, or if my filter is based on targets' metadata only? I would rather react on TargetCreated/TargetUpdated events and perform the filter matching in the background thread of some autoAssignmentExecutor not to block the async communication. However this approach should also be carefully evaluated comparing the amount/frequency of TargetCreated/TargetUpdated events and filter matching performance/resource consumption, probably a cache should be introduced for the target filter queries to minimise the database calls. 

Second notice concerns the introduction of many duplications under ControllerManagement and DeploymentManagement and re-implementing the logic of AutoAssignChecker/Scheduler. I would suggest to follow the separation of concerns pattern and just adapt/extend the classes that are already responsible for the auto-assignment as opposed to moving all of the logic into the Management layer. Moreover overloading should be used where possible for effectively the same methods with different signature (e.g. list/single arguments). We could discuss additional minor remarks after the concrete structure is agreed upon.

In case you have any questions/concerns feel free to contact me!

avgustinmm:
This is something that could be useful - having assignment before the auto assignment scheduler run. However, the PR is outdated (conflicts), there are concerns raised (and not answered). 
I'll close the PR. 
If there is an interest in this PR to be pushed further please let us know - we would reopen it and conversation could be continued

Issue #1029, opened 2020-10-17
Misleading unnecessary code in RSQLUtility - unused operator "=li="


Sergey-G-dev:
When working on https://github.com/eclipse/hawkbit/issues/1023 I've found a suspicious line of code in `RSQLUtility.java`, method `static Node parseRsql(final String rsql)`, line No. 156:
`operators.add(new ComparisonOperator("=li=", false))`

It adds the `=li=` as a valid operator to the RSQL parser, so that no exception is thrown by the parser during validation so that this appears to be a valid syntax. But the RSQL Visitor is not aware of this operator and throws a "Unkown operator exception" if it encounters this operator when traversing the nodes . 

So I would suggest to remove this line since it has no further meaning except confusing the users, making them think that `=li=` is a valid operator. I'll create a PR shortly.

PR #1030, opened 2020-10-17
Removed misleading code from RSQLUtility - unused operator '=li='

Sergey-G-dev:
This is a fix for https://github.com/eclipse/hawkbit/issues/1029

Summary. The unused RSQL operator is removed from the list of valid operators, as it is not implemented by the RSQL visitor and can only confuse users when valitating the RSQL filter syntax, making them think that this is a valid operator.


schabdo:
Indeed, well spotted @Sergey-G-dev! Seems like this line of code is there from the very early days 😄

Issue #955, opened 2020-03-31
Java 11 support


schabdo:
As Java 8-10 will be obsolete soon, hawkBit has to support Java 11

#### Preconditions
- [ ] Migration to Vaadin 8
- [ ] Release 0.3.0 (on Java 8)

#### ToDo
- [ ] Migrate hawkBit
- [ ] Migrate hawkBit Extensions
- [ ] Migrate hawkBit Examples
- [ ] Update Dokerfiles
- [ ] CircleCI build on JDK 11
- [ ] Update Documentation

bantu:
Attempting to compile master at eae34e65ccf7e79ddbab65b77f04ddecee1ed4db using the `openjdk-11-jdk` package provided by Debian 10 fails with
```
[INFO] Compiling module org.eclipse.hawkbit.ui.AppWidgetSet
[INFO]    [ERROR] Hint: Check that your module inherits 'com.google.gwt.core.Core' either directly or indirectly (most often by inheriting module 'com.google.gwt.user.User')
```
On closer inspection this seems to be caused by GWT.

Maybe the Java 8 source code can be compiled using OpenJDK 11 after updating GWT. This would be a first step towards moving to Java 11 entirely.

schabdo:
Exactly! I already prepared some stuff to achieve exactly what you're proposing (cf. [#674](https://github.com/eclipse/hawkbit/pull/674)). However back in the days (as you mentioned) Vaadin 7 was always the show-stopper. Since a couple of days `master` is on Vaadin 8, we are one step further: Now WidgetSet could be complied based on JDK 11 (with source level to 1.8). However it seems that the XML bind dependency/module is missing for the tests:

```
[ERROR] javax/xml/bind/annotation/XmlElement
[ERROR] org.apache.maven.surefire.booter.SurefireBooterForkException: There was an error in the forked process
[ERROR] javax/xml/bind/annotation/XmlElement
[ERROR] 	at org.apache.maven.plugin.surefire.booterclient.ForkStarter.fork(ForkStarter.java:656)
```

avgustinmm:
Obsolete, hawkBit is supporting Java17 already

PR #1038, opened 2020-10-26
Java 11 support for hawkBit

schabdo:
Changes allow to bump Java source level from `8` to `11` passing all tests on JDK 11. 
However I do not enforce the switch to Java 11 yet (cf. [Java 11 roadmap](https://github.com/eclipse/hawkbit/issues/955)).

Main changes:

* Latest stable EclipseLink version fixing invalid JPA type serialization
* Fixed compiler and JPA spec violation which is enforced since Java 9 (cf. [JDK-8157181](https://bugs.openjdk.java.net/browse/JDK-8157181))
* Fixed NPE in `eclipselink-maven-plugin` while parsing JPA annotations


vigier (approved):
:+1: looks good - looking forward to switch to Java 11 :smile: :coffee:

PR #1042, opened 2020-11-06
Fixup force delete action documentation

stschake:
This was doing a GET request and didn't even specify
the action ID.

Signed-off-by: Stefan Schake <stefan.schake@devolo.de>


schabdo:
Well spotted @stschake! Seems like test is only failing because the action is not in state `canceled` hence it's not allowed to force quite the action

stschake:
Finally got a minute to update this, should be good now.

schabdo:
Is there anything else missing? If not I'll have a closer look. Just asking due to the draft state of PR

stschake:
Marked it as ready now, thanks for checking!

schabdo:
Sure, here we go ...

PR #1044, opened 2020-11-13

stefbehl:
Execute rollouts and auto assignments in the correct user context instead of the system context. Use the user from `org.eclipse.hawkbit.repository.model.Rollout#getCreatedBy()` and `org.eclipse.hawkbit.repository.model.TargetFilterQuery#getAutoAssignInitiatedBy()` to build the user context. As a consequence, the initatedBy property of the actions (`org.eclipse.hawkbit.repository.model.Action` entities) that are created in the context of the rollout /auto assignment is also initialized with this user.

Signed-off-by: Stefan Behl <stefan.behl@bosch.io>


hawkbit-security-core/src/main/java/org/eclipse/hawkbit/security/InMemoryUserAuthoritiesResolver.java
```diff
@@ -0,0 +1,31 @@
+/**
+ * Copyright (c) 2020 Bosch.IO GmbH and others.
```
bogdan-bondar, line 2: I guess you forgot the empty line after Copyright (compare with UserAuthoritiesResolver.java) ;)

hawkbit-autoconfigure/src/main/java/org/eclipse/hawkbit/autoconfigure/security/SecurityAutoConfiguration.java
```diff
@@ -33,18 +41,42 @@
...
+     *
+     * @return    an {@link InMemoryUserAuthoritiesResolver} bean
+     */
+    @Bean
+    public UserAuthoritiesResolver inMemoryAuthoritiesResolver(final SecurityProperties securityProperties,
+            final MultiUserProperties multiUserProperties) {
+        final Map<String, List<String>> usersToPermissions = multiUserProperties.getUsers().stream() //
+                .collect(Collectors.toMap(User::getUsername, User::getPermissions));
+
+        usersToPermissions.put(securityProperties.getUser().getName(), securityProperties.getUser().getRoles());
```
bogdan-bondar, line 77: I guess we need to check if there are any users defined within the multiUserProperties first and add the spring security user only in opposite case. Otherwise if we define the user with the same name as spring security user, he will get spring security roles/permissions (see InMemoryUserManagementAutoConfiguration#userDetailsService)

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaRolloutManagement.java
```diff
@@ -1147,4 +1152,26 @@ private void assertActionsPerTargetQuota(final Target target, final int requeste
         QuotaHelper.assertAssignmentQuota(target.getId(), requested, quota, Action.class, Target.class,
                 actionRepository::countByTargetId);
     }
+
+    private void runInUserContext(final BaseEntity rollout, final Runnable handler) {
```
bogdan-bondar, line 1156: Due to isolated/generic nature of this method, I would extract it to either DeploymentHelper or even SystemManagement passing the userName parameter instead of BaseEntity. It will then help removing the duplication with AutoAssignChecker and potentially other classes

hawkbit-security-core/src/main/java/org/eclipse/hawkbit/security/InMemoryUserAuthoritiesResolver.java
```diff
@@ -0,0 +1,31 @@
...
+
+    private final Map<String, List<String>> usernamesToPermissions;
+
+    public InMemoryUserAuthoritiesResolver(final Map<String, List<String>> usernamesToPermissions) {
+        this.usernamesToPermissions = usernamesToPermissions;
+    }
+
+    @Override
+    public Collection<String> getUserAuthorities(final String tenant, final String username) {
+        return usernamesToPermissions.get(username);
```
bogdan-bondar, line 29: In my opinion we need to leave a comment here explaining why tenant parameter is ignored. Additionally, just to be on a safe side, should we verify if username is valid (not null and not empty) and if it is better to return empty list instead of null, what do you think?

hawkbit-security-core/src/main/java/org/eclipse/hawkbit/security/SecurityContextTenantAware.java
```diff
@@ -59,44 +72,62 @@ public String getCurrentUsername() {
     }
 
     @Override
-    public <T> T runAsTenant(final String tenant, final TenantRunner<T> callable) {
+    public <T> T runAsTenant(final String tenant, final TenantRunner<T> tenantRunner) {
+        final SecurityContext originalContext = SecurityContextHolder.getContext();
```
bogdan-bondar, line 76: Could you please extract the wrapping context change/rollback to its own method e.g. runInContext(SecurityContext, TenantRunner), thanks!

bogdan-bondar (changes requested):
Great job! I have left only some minor suggestions ;)

hawkbit-autoconfigure/src/main/java/org/eclipse/hawkbit/autoconfigure/security/SecurityAutoConfiguration.java
stefbehl: Good point. How about adding the Spring security user only if it is not yet contained in the map AFTER processing multiUserProperties?

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaRolloutManagement.java
stefbehl: Makes sense! I would design the generic helper method to accept a callback (basically a supplier) which can be invoked to obtain the "fallback user". Because - if the user context is already in place, there is no need to determine the fallback user.

hawkbit-autoconfigure/src/main/java/org/eclipse/hawkbit/autoconfigure/security/SecurityAutoConfiguration.java
a-sayyed: I think this is still ok, if we define a user with the same username as the spring user, that user will not be taken into consideration. The Spring security user should always be there, and should always have the predefined set of authorities related to him.

What we can do, we can check if a user is defined with the same name as the spring security user, then we can throw an exception for example. But I wouldn't make the spring security user optional

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaRolloutManagement.java
stefbehl: Moved to DeploymentHelper#runInNonSystemContext

hawkbit-autoconfigure/src/main/java/org/eclipse/hawkbit/autoconfigure/security/SecurityAutoConfiguration.java
bogdan-bondar: As I've already mentioned above, please, take a look at InMemoryUserManagementAutoConfiguration#userDetailsService implementation. It is either-or semantic, meaning that in case we define the users by ourselves via multiUserProperties spring security user will be omitted in user management completely. So in my opinion we should not add the spring security user at all to the map in case multiUserProperties is not empty for the current implementation. Alternatively, we could change the logic in InMemoryUserManagementAutoConfiguration to ALWAYS setup the spring security user

a-sayyed: ok I was not aware that this is the current behavior of the `InMemoryUserManagement`, somehow was expecting it to configure 'additional' users, and not override the spring security user. I'm fine with either changing the autoconfiguration or adapting the user permissions here

stefbehl: ok, I will fix that accordingly to keep the current semantics. Thanks guys!

bogdan-bondar (approved):
Thanks @stefbehl ! Looks good!

hawkbit-core/src/main/java/org/eclipse/hawkbit/tenancy/UserAuthoritiesResolver.java
```diff
@@ -0,0 +1,30 @@
...
+    /**
+     * User authorities/roles lookup based on the tenant and the username
+     *
+     * @param tenant
+     *            The tenant that this user belongs to
+     * @param username
+     *            The username of the user
+     * @return a {@link Collection} of authorities/roles for this user
+     */
+    Collection<String> getUserAuthorities(String tenant, String username);
```
a-sayyed, line 29: since we now call this always within a system context, should we annotate this method with the appropriate security annotation?
`@PreAuthorize(SpringEvalExpressions.IS_SYSTEM_CODE)`

schabdo:
@hawkbit-bot verify please

stefbehl:
Replaced with https://github.com/eclipse/hawkbit/pull/1100

PR #1046, opened 2020-11-16
CSV export of target data from Target Filter View.

Sergey-G-dev:
Added a CSV export functionality to Target Filter Details View.

Column separator is hardcoded to ";" (semicolon). Column entries are put into double quotes. Can be opened by Excel with one click.

CSVexport


hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/filtermanagement/TargetDataExporter.java
```diff
@@ -0,0 +1,80 @@
...
+import java.text.SimpleDateFormat;
+import java.util.Arrays;
+import java.util.Date;
+import java.util.List;
+import java.util.Optional;
+
+/**
+ * Utility class to provide formatted Target Data to be used in file exports
+ */
+public class TargetDataExporter {
```
bogdan-bondar, line 23: In order to implement the above mentioned batch processing you could use the following variant of the exporter:
```suggestion
/**
 * Utility class to export targets of the {@link ProxyTargetFilterQuery} as CSV.
 */
public class TargetDataCsvExporter {
    private static final Logger LOG = LoggerFactory.getLogger(TargetFilterAddUpdateLayout.class);

    private static final List<String> HEADER_KEYS = Arrays.asList("Name", "Description", "Status", "Created By",
            "Created Date", "Modified By", "Modified Date");

    private final String separator;
    private final SimpleDateFormat dateFormat;
    private final TargetManagement targetManagement;

    public TargetDataCsvExporter(final TargetManagement targetManagement, final String separator,
            final String dateFormatPattern) {
        this.targetManagement = targetManagement;
        this.separator = separator;
        this.dateFormat = new SimpleDateFormat(dateFormatPattern);
    }

    public void writeTargetsByFilterIdTo(final Long targetFilterId, final OutputStream out) {
        try {
            addHeader(out);
            convertTargets(targetFilterId, out);

            out.flush();
            out.close();
        } catch (final IOException e) {
            LOG.error("Exception during csv conversion: ", e);
        }
    }

    private void addHeader(final OutputStream out) throws IOException {
        out.write(HEADER_KEYS.stream().collect(Collectors.joining(separator)).concat("\n").getBytes());
    }

    private void convertTargets(final Long targetFilterId, final OutputStream out) throws IOException {
        Pageable pageable = PageRequest.of(0, SPUIDefinitions.PAGE_SIZE);
        Slice<Target> targets;
        do {
            targets = targetManagement.findByTargetFilterQuery(pageable, targetFilterId);
            out.write(targets.stream().map(this::convertTarget).collect(Collectors.joining("\n")).getBytes());
        } while ((pageable = targets.nextPageable()) != Pageable.unpaged());
    }

    private String convertTarget(final Target target) {
        return Arrays
                .asList(target.getControllerId(), target.getDescription(), target.getUpdateStatus().name(),
                        target.getCreatedBy(), toReadableDate(target.getCreatedAt()), target.getLastModifiedBy(),
                        toReadableDate(target.getLastModifiedAt()))
                .stream().map(Strings::nullToEmpty).map(TargetDataCsvExporter::quote)
                .collect(Collectors.joining(separator));
    }

    private String toReadableDate(final long millis) {
        return dateFormat.format(new Date(millis));
    }

    private static String quote(final String input) {
        return String.format("\"%s\"", input);
    }
}
```

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/filtermanagement/TargetFilterAddUpdateLayout.java
```diff
@@ -135,9 +151,33 @@ public ComponentContainer getRootComponent() {
...
+
+    protected void removeDownloaderExtension(){
+        if(fileDownloader != null){
+            fileDownloader.remove();
+        }
+    }
+
+    private StreamResource getStreamResourceToCSV(final ProxyTargetFilterQuery proxyEntity){
+
+        StreamSource streamSource =  () -> {
```
bogdan-bondar, line 168: And start the processing in its own thread using the ui executor:
```suggestion
        final StreamSource streamSource = () -> {
            final PipedInputStream in = new PipedInputStream();

            try {
                final PipedOutputStream out = new PipedOutputStream(in);
                final TargetDataCsvExporter csvExporter = new TargetDataCsvExporter(targetManagement, ";",
                        "E MMM dd HH:mm:ss z yyyy");

                SpringContextHolder.getInstance().getBean("uiExecutor", Executor.class)
                        .execute(() -> csvExporter.writeTargetsByFilterIdTo(targetFilterQuery.getId(), out));
            } catch (final IOException ex) {
                LOG.warn("Creating piped Stream failed: ", ex);
            }

            return in;
        };

        return new StreamResource(streamSource, targetFilterQuery.getName() + ".csv");
```

bogdan-bondar (changes requested):
Please see the comments

Sergey-G-dev:
Thanks for your feedback and suggestions! I've refactored the whole feature as you suggested.
However, writing a unit test now seems more complicated than the business code itself. I'm a little bit unsure what's the right approach here, so perhaps you can provide some advice. As for my understanding, we need to create a test class that extends 
`AbstractJpaIntegrationTest` class. Then we need to create several Targets and some TargetFilters. Then we need to create a reference object (expected result), which is of type `OutputStream`. Then finally we need to compare the produced `OutputStream` object with the expected (reference) object. Is that somewhat correct or am I thinking too complicated?

bogdan-bondar:
Sorry for the late reply and thanks a lot for adapting the PR! Regarding the test: it depends if you would like to implement the Integration test or a Unit test :) I would suggest to go with the Unit test here by mocking the targetManagement (the only real dependency of TargetDataCsvExporter) and using the ByteArrayOutputStream as your OutputStream. In that case you could just make targetManagement#findByTargetFilterQuery to return some targets and verify if they were converted correctly by using ByteArrayOutputStream#toByteArray(). Moreover, you could even spy on the OutputStream and verify the OutputStream#write calls and their arguments, but that is up to you ;) In case you have further questions/concerns feel free to contact me!

krishna-devolo:
Hi @bogdan-bondar , my name is Krishna and I am a new employee at devolo. I will take over this (and other) PR from Sergey. 

I am facing an issue implementing your suggestion for unit testing. How do I get the targetManagement#findByTargetFilterQuery to return targets? The only way I am aware of is by using TestdataFactory to create targets, but that does not work: The test class for TargetDataCsvExporter is in a different module (hawkbit-ui) and even if I add hawkbit-repository jpa as a module depedency, I cannot import AbstractJpaIntegrationTest (to be able to use testdataFactory).

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/filtermanagement/TargetDataCsvExporter.java
```diff
@@ -0,0 +1,93 @@
...
+    private void addHeader(final OutputStream out) throws IOException {
+        out.write(HEADER_KEYS.stream().collect(Collectors.joining(separator)).concat("\n").getBytes());
+    }
+
+    private void convertTargets(final Long targetFilterId, final OutputStream out) throws IOException {
+        Pageable pageable = PageRequest.of(0, SPUIDefinitions.PAGE_SIZE);
+        Slice<Target> targets;
+
+        do {
+            targets = targetManagement.findByTargetFilterQuery(pageable, targetFilterId);
```
bogdan-bondar, line 72: This will work only for the already saved target filters. In case you create a new one and try to export the targets it would fails with `EntityNotFoundException`. 
In order to fix it you would need to use the `targetManagement.findByRsql` method passing the query string instead of id as a method parameter of `convertTargets`;

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/filtermanagement/TargetFilterAddUpdateLayout.java
```diff
@@ -135,9 +157,46 @@ public ComponentContainer getRootComponent() {
...
+        final StreamSource streamSource = () -> {
+            final PipedInputStream in = new PipedInputStream();
+
+            try {
+                final PipedOutputStream out = new PipedOutputStream(in);
+                final TargetDataCsvExporter csvExporter = new TargetDataCsvExporter(targetManagement, ";",
+                        "E MMM dd HH:mm:ss z yyyy");
+
+                SpringContextHolder.getInstance().getBean("uiExecutor", Executor.class)
+                        .execute(() -> csvExporter.writeTargetsByFilterId(proxyEntity.getId(), out));
```
bogdan-bondar, line 183: I would rather use `csvExporter.writeTargetsByFilterId(proxyEntity.getQuery(), out));`, see the comment above

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/filtermanagement/TargetFilterAddUpdateLayout.java
```diff
@@ -135,9 +157,46 @@ public ComponentContainer getRootComponent() {
...
+
+            } catch (final IOException ex) {
+                LOG.warn("Creating piped Stream failed: ", ex);
+            }
+            return in;
+        };
+
+        return new StreamResource(streamSource, proxyEntity.getName() + ".csv");
```
bogdan-bondar, line 191: It should be validated if the name is not empty (could be the case for Add target filter window). In that case either use default name or adapt the name dynamically upon change

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/filtermanagement/TargetFilterAddUpdateLayout.java
```diff
@@ -135,9 +157,46 @@ public ComponentContainer getRootComponent() {
...
+    }
+
     private void addValueChangeListeners() {
         searchButton.addClickListener(event -> onSearchIconClick());
-        autoCompleteComponent.addValidationListener((valid, message) -> searchButton.setEnabled(valid));
+        autoCompleteComponent.addValidationListener((valid, message) -> {
+            searchButton.setEnabled(valid);
+            exportButton.setEnabled(valid);
```
bogdan-bondar, line 198: Check if we need to additionally verify if name is present before activating the button

bogdan-bondar (changes requested):

bogdan-bondar:
Hi @krishna-devolo ! You don't need to use neither `TestdataFactory` nor `AbstractJpaIntegrationTest`, both of this classes assume there is a Spring Application Context present by auto wiring required dependencies. For your changes however it is sufficient to only unit test the `TargetDataCsvExporter`. In order to do it you can just manually create test `JpaTarget`s and mock the `targetManagement`, something like `when(targetManagement.findByRsql(any(), anyString()))
                .thenReturn(new PageImpl<>(Collections.singletonList(targetInPage1), PageRequest.of(0, 1), 2))
                .thenReturn(new PageImpl<>(Collections.singletonList(targetInPage2), PageRequest.of(1, 1), 2));`, where `targetInPage1` and `targetInPage2` are created manually by `new JpaTarget("targetInPage1")` and populating all the fields as necessary

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/filtermanagement/TargetFilterAddUpdateLayout.java
krishna-devolo: All good suggestions - thanks! 

By name, are you referring to the fileName of CSV export here? If so, we do not need the fileName to be present for activating the button. The button only needs decoration elements, e.g., caption/label and tooltip.

bogdan-bondar:
No, it does not. The reason you are getting the `NullPointerException` is due to a missing `SecurityTokenGenerator`. Try using `new JpaTarget("targetInPage1", null); `

stefbehl:
Closing this PR as it was replaced with https://github.com/eclipse/hawkbit/pull/1194

PR #1047, opened 2020-11-16
DB and RabbitMQ integration tests and PostgreSQL testing/bug fixing

### LINKED ISSUE #1045: JPA Repository broken with SQLServer after Eclipselink update
Hi,

the recent Eclipselink upgrade from #1038 unfortunately broke the SQLServer support.

It causes massive exceptions `com.microsoft.sqlserver.jdbc.SQLServerException: ORDER BY items must appear in the select list if SELECT DISTINCT is specified.`

The reason is the change introduced with https://github.com/eclipse-ee4j/eclipselink/pull/300 which causes the error. 

I am open for suggestions. We either drop SQLServer support or revert the Eclipselink upgrade.

I reported this in Eclipselink as well https://github.com/eclipse-ee4j/eclipselink/issues/963

Kai

CC: @schabdo

kaizimmerm:
Hi,

because of #1045 I worked a bit on an integration test matrix. Leveraging Azure Pipelines up to 10 jobs in parallel capability it can be done in reasonable time. See https://dev.azure.com/kaizimmerm/eclipse/_build/results?buildId=66&view=results  I also created a .devcontainer config (I cannot live without these nowadays :))

While at it I also added integration tests for the PostgreSQL support. However, the tests showed quite a lot of issues that needed fixing. 

What is left is a general different behavior in RSQL queries as PostgreSQL is handling numbers not like strings. That is ok imho but I had to make some of the tests aware of that behavior.

I reverted also the Eclipselink version bump from @schabdo as this would fix #1045 for the time being. Any thoughts on that one?

Kai


schabdo:
@hawkbit-bot verify please

schabdo:
LGTM, thanks for spending the effort on this! Regarding the EclipseLink version I commented [here](https://github.com/eclipse/hawkbit/issues/1045#issuecomment-730302598). 
Looking forward to see another contribution integrating the pipeline in hawkBit as well :)

PR #1056, opened 2020-11-27
Show Controller ID within target, filter and rollouts view

herdt-michael:
Add the controllerId column to the targetGrid of the deployment, rollout target and filter search result view.

### DIFF
--- a/hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/common/builder/GridComponentBuilder.java
+++ b/hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/common/builder/GridComponentBuilder.java
@@ -18,6 +18,7 @@ import java.util.stream.Collectors;
 
 import org.eclipse.hawkbit.ui.common.data.proxies.ProxyIdentifiableEntity;
 import org.eclipse.hawkbit.ui.common.data.proxies.ProxyNamedEntity;
+import org.eclipse.hawkbit.ui.common.data.proxies.ProxyTarget;
 import org.eclipse.hawkbit.ui.common.grid.support.DeleteSupport;
 import org.eclipse.hawkbit.ui.utils.SPUIStyleDefinitions;
 import org.eclipse.hawkbit.ui.utils.UIMessageIdProvider;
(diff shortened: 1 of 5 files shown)


hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/filtermanagement/TargetFilterTargetGrid.java
```diff
@@ -83,4 +86,9 @@ public void addColumns() {
 
         getColumns().forEach(column -> column.setHidable(true));
     }
+
+    private Column<ProxyTarget, String> addControllerIdColumn() {
+        return GridComponentBuilder.addColumn(this, ProxyTarget::getControllerId).setId(TARGET_CONTROLLER_ID)
```
bogdan-bondar, line 91: In my opinion it is better to just extend the GridComponentBuilder with #addControllerIdColumn in order not to duplicate it three times for each grid. It is kind of a special case, that ideally should not be part of GridComponentBuilder, however due to the fact there are multiple target grids used in different places it appears to be the most suitable solution

bogdan-bondar (changes requested):
Looks good! Just a small remark ;)

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/filtermanagement/TargetFilterTargetGrid.java
herdt-michael: Had the same considerations, since the builder only creates columns that each type has, I did not add the controllerId to the builder. But to avoid duplicated code, I will move it into the `GridComponentBuilder`.

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/common/builder/GridComponentBuilder.java
```diff
@@ -120,6 +121,22 @@ public static Button buildLink(final String idSuffix, final String idPrefix, fin
...
+     *            to add the column to
+     * @param i18n
+     *            message source for internationalization
+     * @param columnId
+     *            column ID
+     * @return the created column
+     */
+    public static Column<ProxyTarget, String> addControllerIdColumn(final Grid<ProxyTarget> grid,
+            final VaadinMessageSource i18n, final String columnId) {
+        return addColumn(i18n, grid, ProxyTarget::getControllerId, "header.controllerId", columnId, 100D);
```
bogdan-bondar, line 137: Could you please define the constant for minWidth (100D) parameter not to duplicate it for each method, thanks!

bogdan-bondar (approved):

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/common/builder/GridComponentBuilder.java
herdt-michael: Good idea. Done ✅

schabdo:
@hawkbit-bot verify please

PR #1061, opened 2020-12-11
Fix database migration scripts

StefanKlt:
There is a bug in the database migration scripts version 1_12_16 for PostgreSQL, Microsoft SQL Server and DB2.
The script for DB2 does not work, the scripts for PostgreSQL and Microsoft SQL Server only work if the sp_action table is empty. The reason is that a column was added that does not allow a null value, which results in an error.
To fix that a default value was added to the column. The default value is an empty string and the default is removed from the column after is is created. That way we get the same behaviour that we have when using MySQL.


bogdan-bondar (approved):
Thanks a lot for the fix Stefan!

PR #1074, opened 2021-01-22
Do not accept valid token headers by default

### LINKED ISSUE #1066: SecurityToken Auth works per default without enabeling it in de Management UI
Hi,
I use Hawkbit for the first time and in an explorative way.
Reading the docs, one should enable the usage of security token in the Authentication Configuration in the system config. 
In my case, this flow works without enabeling it.
Is that a default behaviour right now?
Because in the docs it is explicitly menioned that this feature has to be enabled.

schabdo:
The default DDI configuration excepted valid target and gateway token headers even the option to allow them was not set explicitly within settings menu of the management UI. Documentation clearly state that both need to be enabled first  

Fixes #1066


stefbehl (approved):
Looks good. Thanks Dominic.

PR #1077, opened 2021-02-05
Fix NPE that occurs on invalid user input in editfields

Nkyn:
Fix NPE that occurs in UI on invalid user input in editfields

Signed-off-by: Natalia Kislicyn <natalia.kislicyn@bosch.io>


schabdo:
@hawkbit-bot verify please

PR #1079, opened 2021-02-09
Document and explain DB migration issue

StefanKlt:
A bug in a DB migration script was fixed. The issue and the steps a user has to follow in case of an error are now documented in the MIGRATION.md.


schabdo:
Thanks for explaining and preparing the required migration step for the upcoming `M7` milestone

PR #1092, opened 2021-03-03
Improve error handling for the UI

herdt-michael:
In this PR on the one hand the error handling for the user interface is improved and on the other hand a good template for further error handling is created. By adding a generic ConditionalErrorHandler it is possible to have multiple handlers that can process an error event based on a condition.


herdt-michael:
Will be improved with https://github.com/eclipse/hawkbit/pull/1106

PR #1095, opened 2021-03-11

stefbehl:
Execute rollouts and auto assignments in the correct user context instead of the system context. Use the user from org.eclipse.hawkbit.repository.model.Rollout#getCreatedBy() and org.eclipse.hawkbit.repository.model.TargetFilterQuery#getAutoAssignInitiatedBy() to build the user context. As a consequence, the initatedBy property of the actions (org.eclipse.hawkbit.repository.model.Action entities) that are created in the context of the rollout /auto assignment is also initialized with this user.

Signed-off-by: Stefan Behl stefan.behl@bosch.io


stefbehl:
Duplicate

PR #1097, opened 2021-03-16
Feature/remove junit5 leftovers

floruschbaschan:
`spring-cloud-stream-binder-test` has added the JUnit4 dependency to `hawkbit-repository-test`. Therefore modules that include this dependency can still make use of JUnit4. This PR will replace all JUnit4 leftovers with JUnit5.

### DIFF
--- a/hawkbit-repository/hawkbit-repository-core/src/test/java/org/eclipse/hawkbit/event/BusProtoStuffMessageConverterTest.java
+++ b/hawkbit-repository/hawkbit-repository-core/src/test/java/org/eclipse/hawkbit/event/BusProtoStuffMessageConverterTest.java
@@ -65,12 +65,11 @@ public class BusProtoStuffMessageConverterTest {
     @Description("Verifies that a MessageConversationException is thrown on missing event-type information encoding")
     public void missingEventTypeMappingThrowsMessageConversationException() {
         final DummyRemoteEntityEvent dummyEvent = new DummyRemoteEntityEvent(targetMock, "applicationId");
-        try {
-            underTest.convertToInternal(dummyEvent, new MessageHeaders(new HashMap<>()), null);
-            Assertions.fail("Missing MessageConversationException for un-defined event-type");
-        } catch (final MessageConversionException e) {
-            // expected exception
-        }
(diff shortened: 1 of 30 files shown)


hawkbit-repository/hawkbit-repository-core/src/test/java/org/eclipse/hawkbit/event/BusProtoStuffMessageConverterTest.java
```diff
@@ -65,12 +65,11 @@ public void successfullySerializeAndDeserializeEvent() {
...
-        try {
-            underTest.convertToInternal(dummyEvent, new MessageHeaders(new HashMap<>()), null);
-            Assertions.fail("Missing MessageConversationException for un-defined event-type");
-        } catch (final MessageConversionException e) {
-            // expected exception
-        }
+        final MessageHeaders messageHeaders = new MessageHeaders(new HashMap<>());
+
+        org.junit.jupiter.api.Assertions.assertThrows(MessageConversionException.class,
+                () -> underTest.convertToInternal(dummyEvent, messageHeaders, null),
```
bogdan-bondar, line 71: I guess you could use the assertJ here as it is already imported and present. Moreover, there are currently a lot of inconsistencies whether we use Junit5 Jupiter exception assertions or the ones provided by assertJ. I don't expect we fix all of the inconsistency issues all together, and definitely not in the scope of this PR, however I would try to at least keep the consistency within the test class (e.g. see PreAuthTokenSourceTrustAuthenticationProviderTest.java )

bogdan-bondar (changes requested):
Just a small remark, otherwise looks good, thanks Florian!

hawkbit-repository/hawkbit-repository-core/src/test/java/org/eclipse/hawkbit/event/BusProtoStuffMessageConverterTest.java
floruschbaschan: Yes you were right. I just replaced all  Junit5 Jupiter exception assertions by the one from assertJ since they offer more flexibility.

bogdan-bondar (approved):
Looks good! Thanks a lot Florian!

PR #1098, opened 2021-03-16
Abort input stream if needed

herdt-michael:
Add a method to cancel an Artifact Input Stream before it is closed.


herdt-michael:
Changes contained in https://github.com/eclipse/hawkbit-extensions/pull/71 are sufficient to fix the problem.

PR #1100, opened 2021-03-17

stefbehl:
Execute rollouts and auto assignments in the correct user context instead of the system context. Use the user from `org.eclipse.hawkbit.repository.model.Rollout#getCreatedBy()` and `org.eclipse.hawkbit.repository.model.TargetFilterQuery#getAutoAssignInitiatedBy()` to build the user context. As a consequence, the initatedBy property of the actions (`org.eclipse.hawkbit.repository.model.Action` entities) that are created in the context of the rollout /auto assignment is also initialized with this user.

Signed-off-by: Stefan Behl stefan.behl@bosch.io


hawkbit-security-core/src/main/java/org/eclipse/hawkbit/security/SecurityContextTenantAware.java
```diff
@@ -59,44 +79,68 @@ public String getCurrentUsername() {
...
 
-    private static SecurityContext buildSecurityContext(final String tenant) {
+    private static SecurityContext buildSystemSecurityContext(final String tenant) {
+        return buildUserSecurityContext(tenant, SYSTEM_USER, SYSTEM_AUTHORITIES);
+    }
+
+    private static <T> T runAsSystem(final TenantRunner<T> tenantRunner) {
+        final SecurityContext currentContext = SecurityContextHolder.getContext();
+        try {
+            SystemSecurityContext.setSystemContext(currentContext);
```
bogdan-bondar, line 111: I have a feeling like `SecurityContextTenantAware` and `SystemSecurityContext` belong to the same domain of Security context management and it feels weird to me how they interact with each other. I would definitely not address it in the scope of this PR, just wanted to hear your opinion on the matter?

bogdan-bondar (changes requested):
Great job, Stefan! Left a couple of small remarks ;)

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/utils/DeploymentHelper.java
```diff
@@ -110,4 +118,33 @@ public static void successCancellation(final JpaAction action, final ActionRepos
...
+     * @param handler
+     *            The handler to be invoked in the right user context.
+     * @param username
+     *            Callback to obtain the real user the user context should be
+     *            established for.
+     * @param tenantAware
+     *            The {@link TenantAware} bean to determine the current tenant
+     *            context.
+     */
+    public static void runInNonSystemContext(@NotNull final Runnable handler, @NotNull final Supplier<String> username,
```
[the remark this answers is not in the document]
stefbehl: I think the current method name is fine because the method does not switch the given user, if the current user is already a non system user.

hawkbit-security-core/src/main/java/org/eclipse/hawkbit/security/SecurityContextTenantAware.java
stefbehl: I totally agree. Both classes are heavily intertwined. I think some of the methods were marked static just to avoid the circular reference... ;-)

bogdan-bondar (approved):

bogdan-bondar:
@hawkbit-bot verify please

PR #1101, opened 2021-03-17
Fix the creation of the Usage Report

herdt-michael:
Currently, a page request is created for a maximum 1000 tenants to generate the usage report. The existing forEachTenant method is now used to create the report for all existing tenants.


StefanKlt (approved):
Looks good, thanks @herdt-michael 👍

PR #1106, opened 2021-03-23
UI error handling refactoring

bogdan-bondar:
General refactoring of UI error handling

* Introduced possibility to inject UI error handler via DI into Hawkbit UI
* Refactored default error handling
* Provided possibility to define own ui error details extractors


hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/error/extractors/AbstractSingleUiErrorDetailsExtractor.java
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    /**
+     * Extracts single ui error details from given error.
+     * 
+     * @param error
+     *            error to extract details from
+     * @return ui error details if found
+     */
+    protected abstract Optional<UiErrorDetails> findDetails(Throwable error);
```
stefbehl, line 34: Would it make sense to have a default implementation here? It could simply take the simple name of the exception class as caption and the exception message as description.

stefbehl (approved):
Looks good. Thanks Bogdan!

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/error/extractors/AbstractSingleUiErrorDetailsExtractor.java
bogdan-bondar: Not really, it is just a helper class for error details extractors that should specialise themselves on particular exception. The default case (when no extractors are found for the given exception) is managed in HawkbitUIErrorHandler#showGenericErrorNotification. There we log the detailed exception and only show generic localised error notification to the user. In theory we could use the exception class name/message in notification, however it would break the user-facing localisation so I would not do that

PR #1108, opened 2021-04-01
Refactoring TargetFilterQueryManagement

herdt-michael:
Use TargetManagement for the quota related count operation when activating auto assignment instead of performing this operation directly on the repository.


bogdan-bondar:
@hawkbit-bot verify please

PR #1109, opened 2021-04-09
Fix output folder for REST docu tests

Nkyn:
The output folder for the REST documentation snippets was not set properly by the child classes anymore. All snippets were put in the same default folder "output" that was set as the resource name in the abstract class. 
Now every child class has to implement the abstract method `getResourceName()` to define an own output folder for the corresponding resource again. 

Signed-off-by: Natalia Kislicyn <natalia.kislicyn@bosch.io>

### DIFF
--- a/hawkbit-rest/hawkbit-rest-docs/src/test/java/org/eclipse/hawkbit/rest/documentation/AbstractApiRestDocumentation.java
+++ b/hawkbit-rest/hawkbit-rest-docs/src/test/java/org/eclipse/hawkbit/rest/documentation/AbstractApiRestDocumentation.java
@@ -65,7 +65,7 @@ import io.qameta.allure.Feature;
  * Parent class for all Management API rest documentation classes.
  *
  */
-@Feature("Documentation Verfication - API")
+@Feature("Documentation Verification - API")
 @ExtendWith(RestDocumentationExtension.class)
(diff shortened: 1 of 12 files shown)


hawkbit-rest/hawkbit-rest-docs/src/test/java/org/eclipse/hawkbit/rest/documentation/AbstractApiRestDocumentation.java
```diff
@@ -80,17 +80,17 @@
...
 
-    protected String resourceName = "output";
-
     protected RestDocumentationResultHandler document;
 
     protected String arrayPrefix;
 
     protected String host = "management-api.host";
 
+    public abstract String getResourceName();
```
Nkyn, line 89: add description

bogdan-bondar:
@hawkbit-bot verify please

bogdan-bondar (approved):
Thanks for the fix!

hawkbit-rest/hawkbit-rest-docs/src/test/java/org/eclipse/hawkbit/rest/documentation/AbstractApiRestDocumentation.java
Nkyn: done.

PR #1112, opened 2021-04-22
PopUp dialog for canceling queued uploads

singrob:
There is currently a bug within the drag and drop multi file upload. If a user starts a multi-file-upload and switches the view, the file upload get stuck. The reason for this is that the queued upload requests are cancelled on client side but not on backend.

To prevent this a new pop up confirmation box is introduced by this PR. The confirmation box warns the user that they are about to switch the view and therefore clearing the queued uploads. On Confirmation the queued uploads are cancelled (on client and backend side).

Signed-off-by: Robert Sing <robert.sing@bosch-si.com>


hawkbit-ui/src/main/resources/messages.properties
```diff
@@ -161,6 +161,8 @@ caption.maintenancewindow.schedule = Schedule
 caption.maintenancewindow.duration = Duration
 caption.maintenancewindow.timezone = Time Zone (from GMT)
 
+caption.file.upload.inprogress = File upload in progress
```
Nkyn, line 164: caption of the dialog should indicate more clearly what is going to happen when clicking OK. 
e.g. "Stop file upload and leave page"
Or something shorter.

hawkbit-ui/src/main/resources/messages.properties
```diff
@@ -496,6 +498,7 @@ message.key.deleted.or.notAllowed = {0} with key {0} was deleted or you are not
 message.forcedTime.cannotBeEmpty = Forced time can not be empty
 message.forcedTime.missing = Force time is not specified for the time-forced assignment
 message.scheduledTime.cannotBeEmpty = Scheduled time can not be empty
+message.clear.file.upload.queue.confirm = There is still at least one queued file upload pending. By leaving this view the upload queue will be cleared. \nAre you sure you want to leave this view and cancel the pending file upload?
```
Nkyn, line 501: The text uses the word "cancel". But in the dialog itself the Ok button confirms this text and the cancel button does the opposite. 
Replace "cancel" in the text with something else, e.g. "stop"

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/artifacts/UploadArtifactView.java
```diff
@@ -260,4 +270,30 @@ void destroy() {
...
+            confirmDeleteDialog.getWindow().bringToFront();
+        } else {
+            event.navigate();
+        }
+    }
+
+    private boolean isAnyUploadInUploadQueue() {
+        return artifactUploadState.getAllFileUploadProgressValuesFromOverallUploadProcessList().stream()
+                .map(FileUploadProgress::getFileUploadStatus)
+                .anyMatch(fileUploadStatus -> fileUploadStatus.equals(UPLOAD_STARTED));
```
Nkyn, line 296: Shouldn't we also check for the status `UPLOAD_IN_PROGRESS`?

singrob: The files with an upload status of `UPLOAD_IN_PROGRESS` will still be uploaded, even when the view is changed. Only queued up (and not yet started) file uploads will be removed.

hawkbit-ui/src/main/resources/messages.properties
singrob: DONE

hawkbit-ui/src/main/resources/messages.properties
singrob: DONE

Nkyn (approved):
Looks good 👍

PR #1114, opened 2021-04-28
Removed flyway downgrade

ramannas:
Removed the flyway downgrade so that the spring default is picked-up

Signed-off-by: Shruthi Manavalli Ramanna <shruthimanavalli.ramanna@bosch-si.com>

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -145,9 +145,7 @@
       <rabbitmq.http-client.version>3.5.0.RELEASE</rabbitmq.http-client.version>
       <!-- Newer version contains new rule that we violate. Needs to be fixed first -->
       <hibernate-validator.version>6.0.20.Final</hibernate-validator.version>
-      <!-- Newer community version does not support MySQL 5.6 anymore -->
-      <flyway.version>5.2.4</flyway.version>
-      <!-- Spring boot version overrides - END -->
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -145,9 +145,7 @@
       <rabbitmq.http-client.version>3.5.0.RELEASE</rabbitmq.http-client.version>
       <!-- Newer version contains new rule that we violate. Needs to be fixed first -->
       <hibernate-validator.version>6.0.20.Final</hibernate-validator.version>
-      <!-- Newer community version does not support MySQL 5.6 anymore -->
-      <flyway.version>5.2.4</flyway.version>
```
a-sayyed, line 149: and we don't need to track this anymore? or would we end up with the same situation with future versions?

a-sayyed (approved):
looks good, thanks @ramannas :+1:

pom.xml
ramannas: Yes, since we moved to MySQL 5.7 pinning flyway version is not anymore required and can rely on spring specific version.
**would we end up with the same situation with future versions** -incase Flyway decides to remove support for 5.7 in their future versions then YES... we might have to do something similar like this until we move to their supported version :)

Issue #1023, opened 2020-10-08
Unnecessary SQL queries when editing target filters.


Sergey-G-dev:
**Problem description.**
A frequent use case is to modify a target filter. To do that, we have to navigate to "Target Filters" view and select a specific filter by clicking on its name. After that a detailed filter view is presented where the RSQL filter can be edited. At the same time, a database query is generated (based on the RSQL string) and performed and a list of matching targets is fetched from the DB and displayed on the page. But this operation is not necessary for the purpose of modifying a filter. It slows down the workflow and can increase the database load significantly, depending on complexity of the filter and the number of matching targets. 

**Improvement idea.**
A possible improvement would be to change the default behavior and not auto-load the list of matching targets when editing a filter. It would be still possible to fetch and display the matching targets by clicking on the magnifying glass symbol near the save button.

**Attempt to solve.**
After some code analysis and debugging sessions I found a method call which is resposible for performing the SQL query and populating the view with the retrieved targets. That is `targetFilterAddUpdateLayout.filterTargetListByQuery(proxyEntity.getQuery());` in the class `TargetFilterDetailsGridHeader.java`, line 144. After removing this line the view is not populated with the matching targets anymore. But the SQL query is still being generated and performed (in fact, two queries are being performed - one usual SELECT and another SELECT COUNT(..)). This does not make sense at all, because at this point it is unnecessary and the results are not presented anywhere. So I did some further investigation and discovered that there are several listeners attached to `AutoCompleteTextFieldComponent` class as well as its field `TextField queryTextField`. Apparently one of those listeners performs the query on each text change event in the Query Text Field (as long as it is valid). This looks like unnecessary overhead. I spent a good amount of time investigating the attached listeners, but until now I wasn't able to find the code that performs the SQL query. 

So if anyone could give me a hint where to look for that code, I would greatly appreciate it.

Sergey-G-dev:
[...]

PR #1121, opened 2021-05-25
Consider Target Fields when validating RSQL filters

stefbehl:
Based on feedback from (#1023)
Added valid TargetFields to RSQL validation when editing Target Filter.
Validation now also considers allowed Target Fields when validating RSQL filters.

Signed-off-by: Sergey Gerasimov <sergey.gerasimov@devolo.de>
Signed-off-by: Stefan Behl <stefan.behl@bosch.io>


bogdan-bondar (approved):
Thanks @stefbehl for taking it over! In general I don't have any other remarks except of one:
previously a bug was introduced in `MgmtRolloutResource#create` where `targetFilterQueryManagement#verifyTargetFilterQuerySyntax` is used. However, if you take a look at implementation of `JpaTargetFilterQueryManagement#verifyTargetFilterQuerySyntax` you would notice that it doesn't check anything but rather only creates `RSQLSpecification` via `RSQLUtility#parse`. Additionally, I guess while using MGMT Api for creating a target filter in `MgmtTargetFilterQueryResource#createFilter` validation is not being performed at all. In my opinion we can verify and fix it within this PR as well, what do you think?

PR #1124, opened 2021-05-31
ControllerPreAuthenticatedSecurityHeaderFilter: add ingress-nginx auth-tls-* annotations support

sergeysedoy97:
It is possible to enable Client Certificate Authentication using additional annotations in Ingress Rule.

The following headers are sent to the upstream service according to the `auth-tls-*` annotations:

* `ssl-client-issuer-dn`: The issuer information of the client certificate.
Example: "CN=My CA"
* `ssl-client-subject-dn`: The subject information of the client certificate.
Example: "CN=My Client"

This PR adds support for this annotations without any global changes, you just need to overwrite the following properties:
`hawkbit.server.ddi.security.rp.cnHeader` and `hawkbit.server.ddi.security.rp.sslIssuerHashHeader`.


avgustinmm:
this commit is too old and outdated - has a conflicts 
doesn't contain documentation for users of that feature.
it also made a little bit complicated treating of CN header - could be CN or DN which is not good for security matters.
Also it seems that mapping of DN to CN could be made at nginx level - https://eclipse.dev/hawkbit/concepts/authentication/
So, I'd close the PR. 
@sergeysedoy97, if you still have interest this PR to become a part of hawkBit I could reopen it and we could proceed with conversations.

PR #1125, opened 2021-06-02
Feature/speedup quota tests

a-sayyed:
Reduced the number of entities created in tests to hit quota limits faster


floruschbaschan (commented):
I just found some minor improvements. In addition, while checking the quota logs I noticed in `tooMuchDeploymentActionMessagesInFeedback` inside of `DdiDeploymentBaseTest` we are creating nearly the double amount of the allowed quota:
```
        for (int i = 0; i < 51; i++) {
            messages.add(String.valueOf(i));
        }
```
Here we could also use the default value defined within the `QuotaManagement`:
```
        for (int i = 0; i < quotaManagement.getMaxMessagesPerActionStatus() + 1; i++) {
            messages.add(String.valueOf(i));
        }
```

hawkbit-rest/hawkbit-ddi-resource/src/test/java/org/eclipse/hawkbit/ddi/rest/resource/DdiDeploymentBaseTest.java
```diff
@@ -567,7 +567,7 @@ public void tooMuchDeploymentActionMessagesInFeedback() throws Exception {
                 .get(0);
 
         final List<String> messages = new ArrayList<>();
-        for (int i = 0; i < 51; i++) {
```
floruschbaschan, line 570: You need to reach the limit here:
`quotaManagement.getMaxMessagesPerActionStatus() + 1`

hawkbit-repository/hawkbit-repository-jpa/src/test/java/org/eclipse/hawkbit/repository/jpa/RolloutGroupManagementTest.java
```diff
@@ -44,11 +44,11 @@ public void nonExistingEntityAccessReturnsNotPresent() {
...
             + " by means of throwing EntityNotFoundException.")
     @ExpectEvents({ @Expect(type = RolloutDeletedEvent.class, count = 0),
-            @Expect(type = RolloutGroupCreatedEvent.class, count = 10),
-            @Expect(type = RolloutGroupUpdatedEvent.class, count = 10),
+            @Expect(type = RolloutGroupCreatedEvent.class, count = 5),
+            @Expect(type = RolloutGroupUpdatedEvent.class, count = 5),
             @Expect(type = DistributionSetCreatedEvent.class, count = 1),
             @Expect(type = SoftwareModuleCreatedEvent.class, count = 3),
-            @Expect(type = RolloutUpdatedEvent.class, count = 1), @Expect(type = TargetCreatedEvent.class, count = 10),
+            @Expect(type = RolloutUpdatedEvent.class, count = 1), @Expect(type = TargetCreatedEvent.class, count = 5),
```
floruschbaschan, line 51: `TestdataFactory` calculates the number of targets: `quotaManagement.getMaxTargetsPerRolloutGroup() * quotaManagement.getMaxRolloutGroupsPerRollout()`. Therefore you need to adapt the number of `TargetCreatedEvent` to:

`@Expect(type = TargetCreatedEvent.class, count = 125`

floruschbaschan (changes requested):
Those findings will fix the test failures

floruschbaschan (approved):
LGTM, thanks @a-sayyed

floruschbaschan (approved):

floruschbaschan (approved):

hawkbit-repository/hawkbit-repository-jpa/src/test/java/org/eclipse/hawkbit/repository/jpa/ConcurrentDistributionSetInvalidationTest.java
```diff
@@ -38,35 +48,54 @@
...
+            rolloutManagement.handleRollouts();
+            return 0;
+        }, tenant)).start();
+
         // wait until at least one RolloutGroup is created, as this means that
         // the thread has started and has acquired the lock
-        Awaitility.await().until(() -> tenantAware.runAsTenant(tenant, () -> systemSecurityContext
+        Awaitility.await().atMost(Duration.FIVE_SECONDS)
+                .pollInterval(Duration.ONE_HUNDRED_MILLISECONDS)
+                .until(() -> tenantAware.runAsTenant(tenant, () -> systemSecurityContext
```
bogdan-bondar, line 98: could we also use `systemSecurityContext.runAsSystemAsTenant` as above?

hawkbit-repository/hawkbit-repository-jpa/src/test/java/org/eclipse/hawkbit/repository/jpa/ConcurrentDistributionSetInvalidationTest.java
```diff
@@ -38,35 +48,54 @@
...
+        @Primary
+        RolloutGroupRepository slowRolloutGroupRepository(final RolloutGroupRepository groupRepo) {
+            final RolloutGroupRepository slowGroupRepo = mock(RolloutGroupRepository.class, delegatesTo(groupRepo));
+
+            doAnswer(invocation -> {
+                final JpaRolloutGroup group = invocation.getArgument(0);
+                if (group.getId() == null) {
+                    return groupRepo.save(group);
+                }
+                TimeUnit.SECONDS.sleep(2);
```
bogdan-bondar, line 72: could you please elaborate/comment why 2 seconds are chosen here. Is there really no better approach as to wait before saving, doesn't it mask the real problem of quota calculation?..

bogdan-bondar (commented):
Thanks Ahmed, nice job! Just left a couple of points to clarify

hawkbit-repository/hawkbit-repository-jpa/src/test/java/org/eclipse/hawkbit/repository/jpa/ConcurrentDistributionSetInvalidationTest.java
a-sayyed: I did not change the logic, I only added the `atMost()` and the `pollInterval()` to have a faster feedback

hawkbit-repository/hawkbit-repository-jpa/src/test/java/org/eclipse/hawkbit/repository/jpa/ConcurrentDistributionSetInvalidationTest.java
a-sayyed: the idea here is to simulate a slow group repository, therefore I added the sleep construct 2 seconds are just a fitting number.
I am not sure I got your point about the quota calculation.
The test was successful previously because we create a big number of entities and groups, and so the ds invalidation could succeed, but since we reduced this number, the test is too fast for a concurrent invalidatoin

PR #1131, opened 2021-06-09
Added userinfo endpoint to validate basic auth

aktivk:
added userinfo API endpoint to verify the basic authentication credentials and returns the user information with the tenant and username


stefbehl (changes requested):
Looks good. Thanks Anand!
Would be great if you could add a unit test which also covers the error cases. Thanks!

stefbehl (approved):
Looks good. Thanks Anand!

PR #1136, opened 2021-07-02
Publish target assign event

herdt-michael:
Throw the TargetAssignDistributionSetEvent even if there are actions in CANCELING state present. Filter the actions on the receiver side. In this case at the DMF.


hawkbit-dmf/hawkbit-dmf-amqp/src/test/java/org/eclipse/hawkbit/integration/AbstractAmqpServiceIntegrationTest.java
```diff
@@ -217,6 +216,20 @@ protected Message createAndSendPingMessage(final String correlationId, final Str
         return message;
     }
 
+    protected void sendActionUpdateStatus(final DmfActionUpdateStatus actionStatus) {
```
bogdan-bondar, line 219: please unify/replace with `createAndSendActionStatusUpdateMessage`

bogdan-bondar (changes requested):

bogdan-bondar (approved):

Issue #1138, opened 2021-07-07
Run hawkBit Update Server with services as Docker Stack or as Docker Compose fails due to MySQL container issue.


mgilroy:
Hi,

Following the getting started  guide https://www.eclipse.org/hawkbit/gettingstarted I've been unable to run either the 
instructions to run hawkBit Update Server with services as Docker Stack or as Docker Compose

docker stack deploy -c docker-compose-stack.yml hawkbit

or

docker-compose up -d

The issue appears to be the inclusion of the following line in the hawkbit-runtime/docker/docker-compose.yml file:

MYSQL_USER: "root" 

Removing this line and both docker commands work. 

The default username and password of admin:admin are also not stated anywhere in the getting started guide.

floruschbaschan:
Hi @mgilroy , yes you are right. It looks like it is no longer working like this since [this mysql PR](https://github.com/docker-library/mysql/pull/749) was merged. `MYSQL_USER: "root"` is not required anymore since it is created by default (see https://hub.docker.com/_/mysql/). I would suggest to add a respective comment and remove the `MYSQL_USER` property.

Indeed, also the documentation has room for improvement like mentioning the default username and password in the getting started guide. Would you mind contributing these small fixes? :)

avgustinmm:
Outdated. There are many changes on dockers and docker compose.
Closing. If still relevant - open new issue.

PR #1139, opened 2021-07-08
Remove mysql_user and update documentation

mgilroy:
Should resolve issue #1138.

Removed the MYSQL_USER value from docker and replaced with a comment. Added a comment in the docker-compose.yml and docker-compose-stack.yml to include the default login details, also added the login details as a table to the getting started guide.

### DIFF
--- a/docs/content/gettingstarted.md
+++ b/docs/content/gettingstarted.md
@@ -31,6 +31,13 @@ In addition, the following vendors offer free trial accounts for their Eclipse h
 | MySQL |  | &#10003; | &#10003; | 
 | RabbitMQ |  | &#10003; | &#10003; | 
 
+HawkBit Update Server docker container login credentials:
+
+| Username | Password |
+|---|---|
+| admin | admin |
+| hawkBit Update Server
+
 ### A: Run hawkBit Update Server as Docker Container
 
 Start the hawkBit Update Server as a single container
--- a/hawkbit-runtime/docker/docker-compose-stack.yml
+++ b/hawkbit-runtime/docker/docker-compose-stack.yml
@@ -31,6 +31,8 @@ services:
 
   # ---------------------
   # hawkBit simulator
+  # username: admin
+  # password: admin
   # ---------------------
   simulator:
     image: "hawkbit/hawkbit-device-simulator:latest"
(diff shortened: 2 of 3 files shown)


docs/content/gettingstarted.md
```diff
@@ -31,6 +31,13 @@ In addition, the following vendors offer free trial accounts for their Eclipse h
 | MySQL |  | &#10003; | &#10003; | 
 | RabbitMQ |  | &#10003; | &#10003; | 
 
+HawkBit Update Server docker container login credentials:
```
floruschbaschan, line 34: Only one service feels a little bit lost in this table. I would suggest to mention the default credentials in a sentence which is enriched with some additional information. Something like:

hawkBit Update Server uses `username=admin` and `password=admin` as default login credentials. They can be overridden by the environment variables `spring.security.user.name` and `spring.security.user.password` which are defined in the corresponding default [application.properties](https://github.com/eclipse/hawkbit/blob/master/hawkbit-runtime/hawkbit-update-server/src/main/resources/application.properties).

hawkbit-runtime/docker/docker-compose-stack.yml
```diff
@@ -31,6 +31,8 @@ services:
 
   # ---------------------
   # hawkBit simulator
+  # username: admin
```
floruschbaschan, line 34: IMHO it would be enough to mention the default credentials in the documentation only :)

hawkbit-runtime/docker/docker-compose.yml
```diff
@@ -41,6 +41,8 @@ services:
 
   # ---------------------
   # HawkBit service
+  # username: admin
```
floruschbaschan, line 44: IMHO it would be enough to mention the default credentials in the documentation only :)

bogdan-bondar:
@mgilroy could you please address the PR review comments so we could take the fix into the master?

hawkbit-runtime/docker/docker-compose.yml
mgilroy: Sure, made the change.

hawkbit-runtime/docker/docker-compose-stack.yml
mgilroy: Sure, made the change.

docs/content/gettingstarted.md
mgilroy: Replaced the table and pushed the change,.

hawkbit-runtime/docker/docker-compose.yml
floruschbaschan: Somehow this change was not committed

floruschbaschan:
@mgilroy it looks like one of your changes was not committed. Please have a look: https://github.com/eclipse/hawkbit/pull/1139#discussion_r685875685

PR #1144, opened 2021-07-19
docker-compose.yml: remove MYSQL_USER

diegorondini:
Recent docker images for mysql fail to start with the message:
[ERROR] [Entrypoint]: MYSQL_USER="root", MYSQL_USER and MYSQL_PASSWORD
are for configuring a regular user and cannot be used for the root user

Remove the MYSQL_USER env var as it is no longer needed:
https://stackoverflow.com/a/66910240/813810

Signed-off-by: Diego Rondini <diego.rondini@kynetics.com>


bogdan-bondar:
Duplicate/will be resolved with #1139

Issue #1142, opened 2021-07-13
Align query approaches towards using specifications


dobleralex:
### Problem

Currently the hawkbit project uses a mix of multiple approaches to embed/generate database queries into the application:

- `@Query` annotations for custom queries in repository classes ([example](https://github.com/eclipse/hawkbit/blob/a8f7f50cf902f8496c7c0e908b70fbb2c980e5ec/hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/TargetRepository.java#L92))
- Queries generated by spring from method names in repositories ([example](https://github.com/eclipse/hawkbit/blob/a8f7f50cf902f8496c7c0e908b70fbb2c980e5ec/hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/TargetRepository.java#L166))
- Use spring data specifications to parameterize queries ([example](https://github.com/eclipse/hawkbit/blob/a8f7f50cf902f8496c7c0e908b70fbb2c980e5ec/hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaTargetManagement.java#L161))

This mix makes it hard to keep an overview of the already implemented queries and can also lead to duplicates, I spotted one for example generated from method name (see [here](https://github.com/eclipse/hawkbit/blob/a8f7f50cf902f8496c7c0e908b70fbb2c980e5ec/hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/TargetRepository.java#L178)) as well as specification (see [here](https://github.com/eclipse/hawkbit/blob/a8f7f50cf902f8496c7c0e908b70fbb2c980e5ec/hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/specifications/TargetSpecifications.java#L268)).
This can cause even further problems if they derive from each other leading to different behavior where they are used.

### Suggested solution
Streamline query generation as far as possible to a single of the above solutions. Specifications are the best fit in my opinion. They simplify the repository to only the core queries. Lead to a single go to specification class per entity, making it easy to get an overview of available filters.
In addition specs increase the flexibility, they can be combined to create more complex filters as well as be extended easily later on with additional filters.

A good first step would be to start with a single entity like the `Target` I used for the above examples.

PR #1146, opened 2021-07-21
Replace NoCountPagingRepository by interface with dedicated findAllWithoutCount

dobleralex:
While trying to align query approaches based on specification as described in issue #1142, I noticed that not all find calls where handled by the appropriate repositories. Looking further I found out that in order to optimize performance some service classes pull in a generic `NoCountPagingRepository` implementation that creates new repository on each call based on generic parameters and calls a `findAll()` with a custom `readPage()` implementation. This will skip the additional `count()` call included in the pages total number and is used in cases where this is preserved externally to save the additional DB calls (e.g. in the UI for infinite scrolling).

The problem with this approach is that it places parts of the entity access actions outside the according repository, making them hard to spot and also incompatible with springs mechanisms to inject custom behavior when building the proxy implementation. In addition, generating a new instance for every `findAll()` call seems like unnecessary overhead.

As a replacement I introduced a new repository interface `NoCountSliceRepository` that defines methods for `findAllWithoutCount()`, explicitly calling for the behavior that was previously implied. To include them in all repositories I added it as an extension to hawkbits `BaseEntityRepository`. To provide an implementation to the proxy repositories created by spring jpa I introduced a `HawkbitRepositoryFactoryBean` that will use `SimpleJpaWithNoCountRepository` as a base class. 
This is an extension of the default `SimpleJpaRepository` that in addition also implements the methods of `NoCountSliceRepository` and thus provides them to the proxy repositories.


hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaDistributionSetTypeManagement.java
```diff
@@ -234,8 +232,9 @@ public DistributionSetType unassignSoftwareModuleType(final long dsTypeId, final
 
     @Override
     public Slice<DistributionSetType> findAll(final Pageable pageable) {
-        return convertPage(criteriaNoCountDao.findAll(DistributionSetTypeSpecification.isDeleted(false), pageable,
-                JpaDistributionSetType.class), pageable);
+        return convertPage(
+                distributionSetTypeRepository.findAll(DistributionSetTypeSpecification.isDeleted(false), pageable),
```
dobleralex, line 236: use nocount?

dobleralex: fixed

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/SimpleJpaWithNoCountRepository.java
```diff
@@ -0,0 +1,62 @@
...
+
+    public SimpleJpaWithNoCountRepository(final Class<T> domainClass, final EntityManager em) {
+        super(domainClass, em);
+    }
+
+    public SimpleJpaWithNoCountRepository(JpaEntityInformation<T, ?> entityInformation, EntityManager entityManager) {
+        super(entityInformation, entityManager);
+    }
+
+    public Slice<T> findAllWithoutCount(@Nullable Specification<T> spec, Pageable pageable) {
```
Nkyn, line 45: Add Override annotation to point out that it is required by the interface

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/SimpleJpaWithNoCountRepository.java
```diff
@@ -0,0 +1,62 @@
...
+        TypedQuery<T> query = getQuery(spec, pageable);
+        return pageable.isUnpaged() ? new PageImpl<>(query.getResultList()) : readPageWithoutCount(query, pageable);
+    }
+
+    public Slice<T> findAllWithoutCount(final Pageable pageable) {
```
Nkyn, line 50: Add Override annotation to point out that it is required by the interface

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/SimpleJpaWithNoCountRepository.java
dobleralex: fixed

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/SimpleJpaWithNoCountRepository.java
dobleralex: fixed

Nkyn (approved):
Looks good, thanks 👍

PR #1161, opened 2021-08-09
Feature - play/pause auto assignments

singrob:
This PR introduces play/pause functionality for target filters, to stop automatic provisioning without losing the distribution set. Pausing will only affect new assignments, already created actions are not affected.


laverman:
Hi @singrob , thanks for your contribution. As already discussed on other channels, we want to not merge this PR, but rather think of a potential alignment of the concepts of auto-assignments and rollouts first. - Sorry.

PR #1174, opened 2021-09-01
Colour property in Mgmt API response for DS/SM type

Nkyn:
In the UI, a colour can be set for distribution set type and software module type. Also via the Mgmt API it is possible to include the colour property into the request body of a POST/PUT request.
However, the colour was never part of the response for those API resources. 

This PR adds the colour property to the Mgmt API responses of the DS type and SM type resources. The rest API docs are adapted accordingly, including some other fixes.


hawkbit-rest/hawkbit-mgmt-api/src/main/java/org/eclipse/hawkbit/mgmt/rest/api/MgmtDistributionSetTypeRestApi.java
```diff
@@ -68,9 +67,9 @@
      * within.
      *
      * @param distributionSetTypeId
-     *            the ID of the module type to retrieve
+     *            the ID of the DS type to retrieve
      *
-     * @return a single softwareModule with status OK.
+     * @return a single distributionSet with status OK.
```
bogdan-bondar, line 72: distribution set type

bogdan-bondar (approved):
Looks great! Just a tiny javadoc comment

PR #1176, opened 2021-09-08
Feature invalidation of distribution sets [UI part]

blomark:
Feature to invalidate distribution sets to mark broken updates, and avoid the distribution of such.

- Invalidated distribution sets cannot be assigned to targets using rollouts, auto-assignments, or single assignments. 
- Invalidated distribution sets cannot be "valid" again, but remain invalid.
- Invalidated distribution sets are removed from all auto-assignments
- When invalidating a distribution set the user can choose to also stop running rollouts that references the distribution set that is going to be invalidated
- this PR is based on a PR that introduces distribution set invalidation on API level
- There will be a follow-up PR that will give the user the option to also stop all running actions


blomark:
incorporated and/or discussed all code review remarks

Nkyn (approved):
Looks good, Thanks 👍

bogdan-bondar:
Great job! Thanks Markus!

PR #1181, opened 2021-09-20
README.md: Add rauc-hawkbit-updater to the list of clients supporting hawkBit

ejoerns:
We've just released version 1.0 of the rauc-hawkbit-updater client that we consider to be halfway stable and usable in real-world applications, thus I would say this is a good point in time to add it to the list of clients outside the Eclipse ecosystem so that both projects can continue to benefit from each other.

Link to SW: https://github.com/rauc/rauc-hawkbit-updater
Documentation: https://rauc-hawkbit-updater.readthedocs.io/en/latest/ (still deserves some more content)

The python variant (https://github.com/rauc/rauc-hawkbit) that was primarily written for demo purposes did not receive major updates within the last time, thus adding the word 'demo' here should be appropriate to indicate that the C variant is probably the better choice when not doing prototyping.


laverman:
Thanks for contributing 👍🏻

PR #1204, opened 2021-11-12
align OIDC username attribute name lookup

mdaur:
Signed-off-by: Martin Daur <mdaur@gmx.net>
This fix aligns the behaviour hence OIDC username attribute name lookup at these two places:
- OidcBearerTokenAuthenticationFilter/doFilter
- JwtAuthoritiesOidcUserService/loadUser

Additionally, this fix checks the existence of the claim which is set by the Spring user-name-attribute property. Doing so allows OIDC scenarios in which the userNameAttributeName might not be available in case of different OAuth flows. E.g. in Azure AD preferredUsername is available for OAuth 2.0 auth code grant and OAuth 2.0 device code flow but it is not available for Oauth 2.0 client credentials grants with shared secrets or certificates.

To align with the Spring user-name-attribute property (spring.security.oauth2.client.provider) it makes sense to return the name of the OidcUser and not hardcoded the preferredUser (see SpringSecurityAuditorAware).

For backward compatibly you may configure the provider as follows (including the benefit to have a fallback to the sub as the username if there is no such claim):
spring.security.oauth2.client.provider.azure.user-name-attribute=preferred_username

Without these adjustments you may end up with "NULL" usernames at various places (e.g. Created by, Last modified by), which also breaks the UI.

### DIFF
--- a/hawkbit-autoconfigure/src/main/java/org/eclipse/hawkbit/autoconfigure/security/OidcUserManagementAutoConfiguration.java
+++ b/hawkbit-autoconfigure/src/main/java/org/eclipse/hawkbit/autoconfigure/security/OidcUserManagementAutoConfiguration.java
@@ -162,8 +162,8 @@ class JwtAuthoritiesOidcUserService extends OidcUserService {
 
         final String userNameAttributeName = clientRegistration.getProviderDetails().getUserInfoEndpoint()
                 .getUserNameAttributeName();
-        OidcUser oidcUser;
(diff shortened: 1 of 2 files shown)


pdomineaux:
Hello,
Is this pull request will fix the issue with the null in the initiated_by when trying to associate 'distribution set' to taget?

avgustinmm:
Seems outdated. With removal of legacy UI hawkbit doesn't support Oidc login.

PR #1207, opened 2021-11-23
Same origin x-frame-options header security config

bogdan-bondar:
By default Spring Security Config sets the "X-Frame-Options" header to "DENY" in order to prevent https://en.wikipedia.org/wiki/Clickjacking attack. However, this constraint should be relaxed by setting "SAME-ORIGIN" value for "/UI" routes in order to overcome file-downloading problems in Safari browser.


dobleralex (approved):
Looks good

PR #1209, opened 2021-12-06
Make entity object of event optional

StefanKlt:
A RemoteEntityEvent provides the entity that caused the event. By the time the event is consumed, the entity might already be deleted. This should be made explicit by returning an Optional to avoid NullPointerExceptions.

Signed-off-by: Stefan Klotz <stefan.klotz@bosch.io>


bogdan-bondar (approved):
Looks good!

PR #1212, opened 2021-12-10
Bump log4j version to mitigate CVE-2021-44228

dobleralex:
Alternatively it can be mitigated by running hawkbit with property `log4j2.formatMsgNoLookups=true`


stefbehl (approved):
Looks good. Thanks Alex!

stefbehl:
log4j-core is not used at the moment

PR #1220, opened 2022-01-17
DDI API: /controller/v1/{controllerid}/installedBase/{actionId}

Nkyn:
After installing an update and finishing the corresponding action successfully, the device has no chance to compare its latest installed software with the installed distribution that is reflected on hawkbit side (e.g. for a consistency check). 
Furthermore, a device might want to re-access some of the artefacts later after the installation. This is currently possible by calling the `/deploymentBase` again, but is more or less a hidden link, as the controller base in the DDI API won't show that link after the action was closed. 

This feature adds a new link to the controller base of the DDI API: `/installedBase`
Thus the device can still access the deployment operation and artefacts of the latest installed action.

### DIFF
--- a/hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaControllerManagement.java
+++ b/hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaControllerManagement.java
@@ -1065,4 +1065,24 @@ public class JpaControllerManagement extends JpaActionManagement implements Cont
     void setTargetRepository(final TargetRepository targetRepositorySpy) {
         this.targetRepository = targetRepositorySpy;
     }
+
+    @Override
+    public Optional<Action> getInstalledActionByTarget(final long targetId) {
+        throwExceptionIfTargetDoesNotExist(targetId);
+
+        final JpaTarget jpaTarget = targetRepository.findOne(TargetSpecifications.hasId(targetId)).get();
+
+        final JpaDistributionSet installedDistributionSet = jpaTarget.getInstalledDistributionSet();
+        if (null != installedDistributionSet) {
+            final List<Action> action = actionRepository.findActionByTargetAndDistributionSet(targetId,
+                    installedDistributionSet.getId());
+            if (action.isEmpty() || action.get(0).isCancelingOrCanceled()) {
+                return Optional.empty();
+            }
+            return Optional.ofNullable(action.get(0));
+        } else {
+            return Optional.empty();
+        }
+
+    }
 }
--- a/hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaControllerManagement.java
+++ b/hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaControllerManagement.java
@@ -1075,6 +1075,7 @@ public class JpaControllerManagement extends JpaActionManagement implements Cont
         afterCommit.afterCommit(() -> eventPublisherHolder.getEventPublisher().publishEvent(
                 new CancelTargetAssignmentEvent(target, actionId, eventPublisherHolder.getApplicationId())));
     }
+
     // for testing
     void setTargetRepository(final TargetRepository targetRepositorySpy) {
         this.targetRepository = targetRepositorySpy;
--- a/hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaControllerManagement.java
+++ b/hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaControllerManagement.java
@@ -1056,33 +1056,27 @@ public class JpaControllerManagement extends JpaActionManagement implements Cont
         return actionRepository.findByExternalRef(externalRef);
     }
 
-    private void cancelAssignDistributionSetEvent(final JpaTarget target, final Long actionId) {
-        afterCommit.afterCommit(() -> eventPublisherHolder.getEventPublisher().publishEvent(
-                new CancelTargetAssignmentEvent(target, actionId, eventPublisherHolder.getApplicationId())));
-    }
-
-    // for testing
-    void setTargetRepository(final TargetRepository targetRepositorySpy) {
-        this.targetRepository = targetRepositorySpy;
-    }
-
     @Override
     public Optional<Action> getInstalledActionByTarget(final long targetId) {
-        throwExceptionIfTargetDoesNotExist(targetId);
-
-        final JpaTarget jpaTarget = targetRepository.findOne(TargetSpecifications.hasId(targetId)).get();
+        final JpaTarget jpaTarget = targetRepository.findOne(TargetSpecifications.hasId(targetId))
+                .orElseThrow(() -> new EntityNotFoundException(Target.class, targetId));
 
         final JpaDistributionSet installedDistributionSet = jpaTarget.getInstalledDistributionSet();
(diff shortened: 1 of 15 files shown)


singrob (changes requested):

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaControllerManagement.java
```diff
@@ -1056,6 +1056,21 @@ public void updateActionExternalRef(final long actionId, @NotEmpty final String
...
+    @Override
+    public Optional<Action> getInstalledActionByTarget(final long targetId) {
+        final JpaTarget jpaTarget = targetRepository.findOne(TargetSpecifications.hasId(targetId))
+                .orElseThrow(() -> new EntityNotFoundException(Target.class, targetId));
+
+        final JpaDistributionSet installedDistributionSet = jpaTarget.getInstalledDistributionSet();
+        if (null != installedDistributionSet) {
+            final List<Action> action = actionRepository.findActionByTargetAndDistributionSet(targetId,
+                    installedDistributionSet.getId());
+            return Optional.ofNullable(action.get(0));
```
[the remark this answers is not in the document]
Nkyn: with the suggestion above and `Optional<Action>` as the return value, we can return `findFirstByTargetIdAndDistributionSetId `directly.

hawkbit-rest/hawkbit-ddi-resource/src/main/java/org/eclipse/hawkbit/ddi/rest/resource/DdiRootController.java
```diff
@@ -506,6 +490,53 @@ private Status handleClosedCase(final DdiActionFeedback feedback, final String c
...
+        LOG.debug("getControllerInstalledAction({})", controllerId);
+
+        final Target target = controllerManagement.getByControllerId(controllerId)
+                .orElseThrow(() -> new EntityNotFoundException(Target.class, controllerId));
+
+        final Action action = findActionWithExceptionIfNotFound(actionId);
+        if (!action.getTarget().getId().equals(target.getId())) {
+            LOG.warn(GIVEN_ACTION_IS_NOT_ASSIGNED_TO_GIVEN_TARGET, action.getId(), target.getId());
+            return ResponseEntity.notFound().build();
+        }
```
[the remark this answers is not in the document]
Nkyn: Moved to new method `verifyActionAssignedToTarget` where an `EntityNotFoundException` is thrown.
However, this produces an error body, whereas before the body was empty when 404 was returned in such a case.

hawkbit-rest/hawkbit-ddi-resource/src/main/java/org/eclipse/hawkbit/ddi/rest/resource/DdiRootController.java
```diff
@@ -492,20 +458,57 @@ private Status handleClosedCase(final DdiActionFeedback feedback, final String c
...
+    public ResponseEntity<DdiDeploymentBase> getControllerInstalledAction(@PathVariable("tenant") final String tenant,
+            @PathVariable("controllerId") final String controllerId, @PathVariable("actionId") final Long actionId,
+            @RequestParam(value = "actionHistory", defaultValue = DdiRestConstants.NO_ACTION_HISTORY) final Integer actionHistoryMessageCount) {
+        LOG.debug("getControllerInstalledAction({})", controllerId);
+
+        final Target target = findTargetWithExceptionIfNotFound(controllerId);
+        final Action action = findActionWithExceptionIfNotFound(actionId);
+        verifyActionAssignedToTarget(target, action);
+
+        if (!action.isActive() && !action.isCancelingOrCanceled()) {
```
dobleralex, line 480: Minor thing but I would suggest to reverse the condition. So move the notFound() return into the condition block and the expected successful path currently inside the block out and to the end, usually easier to read that way.

hawkbit-rest/hawkbit-ddi-resource/src/test/java/org/eclipse/hawkbit/ddi/rest/resource/DdiInstalledBaseTest.java
```diff
@@ -0,0 +1,411 @@
...
+        testdataFactory.createArtifacts(softwareModuleId);
+        // get artifacts
+        performGet(SOFTWARE_MODULE_ARTIFACTS, MediaType.parseMediaType(DdiRestConstants.MEDIA_TYPE_CBOR),
+                status().isOk(), tenantAware.getCurrentTenant(), target.getControllerId(),
+                String.valueOf(softwareModuleId));
+    }
+
+    @Test
+    @Description("Test several deployments to a controller. Checks that action is represented as installedBase after installation.")
+    public void deploymentSeveralActionsInInstalledBase() throws Exception {
```
dobleralex, line 97: It is possible to also look up previous installedBase action IDs, but your tests only check for the latest (as far as I can see). I would suggest to either extend this test or add another one were you check that looking up older installedBase actionIds works.

hawkbit-rest/hawkbit-rest-docs/src/main/asciidoc/rootcontroller-api-guide.adoc
```diff
@@ -301,6 +305,64 @@ include::../errors/409.adoc[]
...
 |===
+
+== GET /{tenant}/controller/v1/{controllerid}/installedBase/{actionId}
+
+=== Implementation notes
+
+Resource to receive information of the latest installation.
+Can be used to re-retrieve artifacts of the last finished action, for example in case a re-installation is necessary.
+The response will be of the same format as the deploymentBase operation, providing the latest action that has been finished successfully.
+As the action is already finished, no further feedback is expected.
```
dobleralex, line 316: As accessing not only the latest but all previous installed actions is possible maybe replace `latest` with `previous`.

hawkbit-rest/hawkbit-rest-docs/src/test/java/org/eclipse/hawkbit/rest/ddi/documentation/RootControllerDocumentationTest.java
```diff
@@ -450,4 +455,103 @@ public void getSoftwareModulesArtifacts() throws Exception {
                                         .description(DdiApiModelProperties.ARTIFACT_HTTP_HASHES_MD5SUM_LINK))));
     }
 
+    @Test
+    @Description("Resource to receive information of the latest installation. The response will be of same format as "
+            + "the deploymentBase operation.")
```
dobleralex, line 460: Again use term `previous` instead of `latest`

dobleralex (approved):

hawkbit-rest/hawkbit-ddi-resource/src/main/java/org/eclipse/hawkbit/ddi/rest/resource/DdiRootController.java
Nkyn: Done. Inverted the if condition

hawkbit-rest/hawkbit-ddi-resource/src/test/java/org/eclipse/hawkbit/ddi/rest/resource/DdiInstalledBaseTest.java
Nkyn: Done. Added another call to the older action to the same test

hawkbit-rest/hawkbit-rest-docs/src/main/asciidoc/rootcontroller-api-guide.adoc
Nkyn: Right, that wording makes it clearer that the endpoint can be called for any older installed action. 
Docu is adapted.

PR #1232, opened 2022-02-21
Fix installedBase link in DDI base resource

Nkyn:
The recently introduced extension of the DDI API  (`/installedBase` https://github.com/eclipse/hawkbit/pull/1220 ) had a bug, that occurred when the same DistributionSet was assigned several times to the same target. 
E.g. in case an assignment of as DS was cancelled and later re-assigned to the target and successfully finished, the DDI base resource returned the id of the already cancelled action for the `/installedBase` link, instead of the actual action id of the finished action. Following that link would return 404, since the corresponding action is in cancelled state.


sebastian-firsching (approved):
Looks good to me

PR #1235, opened 2022-03-23
Fix overdue timestamp calculation permission

bogdan-bondar:


herdt-michael (approved):
Looks good, thank @bogdan-bondar!

PR #1260, opened 2022-06-14
Improve target management

herdt-michael:
This pull request contains the following enhancements:

- Fix JOIN type when fetching multiple targets.
- Improve creation of multiple devices.
- Override saveAll repository method to retrieve a List instead of a Iterable

### DIFF
--- a/hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/BaseEntityRepository.java
+++ b/hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/BaseEntityRepository.java
@@ -9,6 +9,7 @@
 package org.eclipse.hawkbit.repository.jpa;
 
 import java.io.Serializable;
+import java.util.List;
 import java.util.Optional;
 
 import org.eclipse.hawkbit.repository.jpa.model.AbstractJpaTenantAwareBaseEntity;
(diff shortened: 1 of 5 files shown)


hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/BaseEntityRepository.java
```diff
@@ -40,4 +41,17 @@
      */
     Optional<T> findById(I id);
 
+    /**
+     * Overrides
```
bogdan-bondar, line 45: please write why exactly you need this override and give a link to JpaRepository

bogdan-bondar (commented):
Please fix the tests, either than that everything look fine! :)

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaTargetManagement.java
```diff
@@ -748,7 +748,9 @@ public Target create(final TargetCreate c) {
     @Retryable(include = {
             ConcurrencyFailureException.class }, maxAttempts = Constants.TX_RT_MAX, backoff = @Backoff(delay = Constants.TX_RT_DELAY))
     public List<Target> create(final Collection<TargetCreate> targets) {
-        return targets.stream().map(this::create).collect(Collectors.toList());
+        final List<JpaTarget> targetList = targets.stream().map(JpaTargetCreate.class::cast).map(JpaTargetCreate::build)
+                .collect(Collectors.toList());
+        return new ArrayList<>(targetRepository.saveAll(targetList));
```
bogdan-bondar, line 753: I would still return `Collections.unmodifiableList` and adapt tests instead. The reason for that is after save operation you don't want to modify the resulting list, the only reason why we return this list is to get access to DB ids

PR #1261, opened 2022-06-15
Add springdoc to management rest API

VladimirSvoboda:
This enables a swagger-ui interface to test and discover the API.
And it also notably generates an OpenAPI specification file which could
then be used to generate API clients in multiple languages to interact
with hawkBit.

Compared to https://github.com/kokuwaio/hawkbit-specs , this approach
has the advantage that the specification file can be easily updated
whenever the API is updated.
Also the generated specification is complete.

A similar approach could be used to generate the OpenAPI specification
for the other APIs as asked
[here](https://gitter.im/eclipse/hawkbit?at=5e1c2db3b720fa5b3c06a101).

Signed-off-by: Vladimir Svoboda <vsvoboda@forkoder.eu>

### DIFF
--- a/hawkbit-rest/hawkbit-mgmt-api/pom.xml
+++ b/hawkbit-rest/hawkbit-mgmt-api/pom.xml
@@ -49,5 +49,24 @@
          <artifactId>allure-junit5</artifactId>
          <scope>test</scope>
       </dependency>
+
+      <dependency>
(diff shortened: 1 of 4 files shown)


avgustinmm:
branch has conflicts. also there is undergoing implementation of support for openapi and swagger - already has open api configuration. So, I'll close that PR

PR #1265, opened 2022-07-03
Fixed typographical error in SpServerError enum

hoip:
"hawkbit.server.error.repo.entitiyAlreayExists" -> "hawkbit.server.error.repo.entitiyAlreadyExists"


avgustinmm:
PR has conflicts. Marked as _wontfix_ since almost an year and not complains. 
Also, key of the message could be used by automatics and change could be harmful. 
So, I'll close the PR. We could think further about fixing such typos

PR #1270, opened 2022-08-12
Provide target quicklinks from Target Filters and Rollout views to Deployment view

stefbehl:
Provide target quicklinks from Target Filters and Rollout views to Deployment view.
Further information will be added soon.


hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/management/targettable/TargetGrid.java
```diff
@@ -328,7 +328,7 @@ public void addColumns() {
                 Arrays.asList(addPinColumn(), addDeleteColumn()));
     }
 
-    private Column<ProxyTarget, String> addControllerIdColumn() {
+    private Column<ProxyTarget, Button> addControllerIdColumn() {
```
bogdan-bondar, line 331: controller id is only visible in the maximised Target Table view. Do you think it makes sense to expose it as a link there as we are already on Deployment page and see all the targets?

bogdan-bondar (changes requested):

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/management/targettable/TargetGrid.java
stefbehl: So far we've been using the GridComponentBuilder#buildLink abstraction /helper API to create links. Basically we are creating buttons which are visualized as links. It is also used for all other links (e.g. the rollout group links in the Rollout Management view). I decided to stick to this pattern to maintain consistency regarding look&feel.

bogdan-bondar (approved):

PR #1272, opened 2022-08-16
Cors config adaptation

bogdan-bondar:
Changed cors configuration bean definition to be reusable under different routes


StefanKlt (approved):
looks good, thanks @bogdan-bondar

PR #1276, opened 2022-09-02
Update jsoup

bogdan-bondar:


hawkbit-repository/hawkbit-repository-api/src/main/java/org/eclipse/hawkbit/repository/ValidStringValidator.java
```diff
@@ -0,0 +1,44 @@
...
+
+    private final Cleaner cleaner = new Cleaner(Safelist.none());
+
+    @Override
+    public boolean isValid(final String value, final ConstraintValidatorContext context) {
+        return StringUtils.isEmpty(value) || cleaner.isValid(stringToDocument(value));
+    }
+
+    private static Document stringToDocument(final String value) {
+        final Document xmlFragment = Jsoup.parse(value, "", Parser.xmlParser());
```
stefbehl, line 36: Can the parse call lead to an IOException? Would it make sense to map an IOException to false in the calling method (isValid)?

bogdan-bondar: done

stefbehl (approved):
Looks good. Thanks @bogdan-bondar !

PR #1278, opened 2022-09-13
Added assignment quota error UI extractor

bogdan-bondar:


hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaTargetFilterQueryManagement.java
```diff
@@ -315,9 +315,9 @@ public boolean verifyTargetFilterQuerySyntax(final String query) {
         }
     }
 
-    private void assertMaxTargetsQuota(final String query) {
-        QuotaHelper.assertAssignmentQuota(targetManagement.countByRsql(query),
-                quotaManagement.getMaxTargetsPerAutoAssignment(), Target.class, TargetFilterQuery.class);
+    private void assertMaxTargetsQuota(final String query, final String filterName) {
+        QuotaHelper.assertAssignmentQuota(filterName, targetManagement.countByRsql(query),
+                quotaManagement.getMaxTargetsPerAutoAssignment(), Target.class, TargetFilterQuery.class, null);
```
herdt-michael, line 320: This limitation looks to be incorrectly defined in my opinion. The user probably intends to apply Auto Assignment to a filter after a Rollout has been initiated. It is not unusual to address 20,000 devices with one filter here. In the end, what matters is how many assignments are performed by activating the Auto Assignment. Therefore, shouldn't we perhaps adjust the calculation of the count?

```suggestion
    private void assertMaxTargetsQuota(final String query, final String filterName, final long dsId) {
        QuotaHelper.assertAssignmentQuota(filterName, targetManagement.countByRsqlAndNonDSAndCompatible(dsId, query),
                quotaManagement.getMaxTargetsPerAutoAssignment(), Target.class, TargetFilterQuery.class, null);

```

herdt-michael (commented):
Left a comment regarding the quota check.

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaTargetFilterQueryManagement.java
herdt-michael: Adapted.

PR #1280, opened 2022-09-16
Feature/java11 build

floruschbaschan:
This PR will switch hawkbits java version to 11 and update the following dependencies: 

spring.boot.version - 2.3.7.RELEASE to 2.3.12.RELEASE
spring.cloud.version - Hoxton.SR7 to Hoxton.SR12
eclipselink.version - 2.7.9 to 2.7.11
eclipselink.maven.plugin.version 2.7.5.1 to 2.7.9.1

PR #1283, opened 2022-10-05
Make DdiActionFeedback#time value an optional property

herdt-michael:
With [a previous PR ](https://github.com/eclipse/hawkbit/pull/1280) a minor change was introduced which leads to the problem, that the "time" value is required when providing an action feedback against the DDI API. This will lead to an API break and needs to be fixed by making this value optional again. This PR restores the old behaviour and extends the tests to identify such cases in the future.


stefbehl (approved):
Looks good. Thanks @herdt-michael !

PR #1290, opened 2022-10-25
Adapted resources to reflect switch to sonarcloud

blomark:


floruschbaschan (approved):
Thanks for cleaning up @blomark

PR #1296, opened 2022-11-30
Upgraded few dependencies

ramannas:
Upgraded versions of gson, netty-all, logback, jackson-bom dependencies

Signed-off-by: Shruthi Manavalli Ramanna <shruthimanavalli.ramanna@bosch-si.com>

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -190,7 +190,7 @@
       <!-- SONATYPE-2021-1175 -->
       <logback.version>1.2.9</logback.version>
       <!-- CVE-2020-36518, CVE-2022-42003, CVE-2022-42004 -->
-      <jackson-bom.version>2.14.0-rc1</jackson-bom.version>
+      <jackson-bom.version>2.14.0</jackson-bom.version>
       <!-- Misc libraries versions - END -->
 
       <!-- Release - START -->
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -184,6 +184,13 @@
       <rsql-parser.version>2.1.0</rsql-parser.version>
       <awaitility.version>3.1.2</awaitility.version>
       <io-protostuff.version>1.5.6</io-protostuff.version>
+      <gson.version>2.8.9</gson.version>
+      <netty.version>4.1.74.Final</netty.version>
+
+      <!-- SONATYPE-2021-1175 -->
+      <logback.version>1.2.9</logback.version>
+      <!-- CVE-2020-36518, CVE-2022-42003, CVE-2022-42004 -->
+      <jackson-bom.version>2.14.0-rc1</jackson-bom.version>
```
floruschbaschan, line 193: Please update to 2.14.0 instead of using the RC

pom.xml
```diff
@@ -672,6 +679,12 @@
             <artifactId>jaxb-api</artifactId>
             <version>${jaxb-api.version}</version>
          </dependency>
+         <dependency>
+            <groupId>io.netty</groupId>
+            <artifactId>netty-all</artifactId>
```
floruschbaschan, line 684: This dependency is not required because it is already part of `<artifactId>netty-bom</artifactId>` from the `spring-boot-dependencies`

PR #1301, opened 2022-12-13
Fix Sonar issue in AbstractAmqpIntegrationTest

stefbehl:
Fix Sonar issue in AbstractAmqpIntegrationTest.

Location: hawkbit-dmf/hawkbit-dmf-rabbitmq-test/src/main/java/org/eclipse/hawkbit/rabbitmq/test/AbstractAmqpIntegrationTest.java

Finding: A "NullPointerException" could be thrown; "getQueueProperties()" can return null.
Line 76.


floruschbaschan (approved):
LGTM, thanks @stefbehl

PR #1311, opened 2023-01-13
Management REST API: Allow to filter actions by last status code

stefbehl:
This PR enhances the resource collection "actions" with a new sort- and filter-field "lastStatusCode" which allows to filter the list of actions based on the last (numeric) status code that was reported by a target device.

For example, it can be used to locate all actions which failed with a specific error code.

API usage example:
`/rest/v1/actions?q=lastStatusCode==1720`


StefanKlt (approved):
Looks good, thanks @stefbehl

PR #1313, opened 2023-01-27
Management API: Expose lastStatusCode property of action entities

stefbehl:
With this pull request, the lastStatusCode property of action entities is also exposed via Management REST API, in particular via the endpoint "/rest/v1/targets/target-id/actions/action-id".


herdt-michael (approved):
Looks good, thank you @stefbehl !

PR #1322, opened 2023-02-21
Make exposed CORS headers configurable

herdt-michael:


stefbehl (approved):
Looks good. Thanks @herdt-michael!

PR #1326, opened 2023-02-27
Fix sub entity attribute formatting

herdt-michael:
Fix formatting of the sub entity attribute by verifying the formatting against the sub entity attributes list of the related parent property enum.


stefbehl (approved):
Looks good. Thanks @herdt-michael !

PR #1329, opened 2023-03-03
Fix issues in rollout creation dialog

herdt-michael:
Introduce text input converter and validator to prevent problems with rollout group definitions.

### DIFF
--- a/hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/utils/TrimmingStringConverter.java
+++ b/hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/utils/TrimmingStringConverter.java
@@ -21,11 +21,15 @@ import com.vaadin.data.ValueContext;
 public class TrimmingStringConverter implements Converter<String, String> {
     @Override
     public Result<String> convertToModel(final String s, final ValueContext valueContext) {
-        return Result.ok(StringUtils.trimWhitespace(s));
+        return Result.ok(trimmedString(s));
     }
 
     @Override
     public String convertToPresentation(final String s, final ValueContext valueContext) {
-        return StringUtils.trimWhitespace(s);
+        return trimmedString(s);
+    }
+
+    private static String trimmedString(final String s){
+        return s == null ? "" : StringUtils.trimWhitespace(s);
     }
 }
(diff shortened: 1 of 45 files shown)


hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/utils/TrimmingStringConverter.java
```diff
@@ -0,0 +1,31 @@
...
+ */
+public class TrimmingStringConverter implements Converter<String, String> {
+    @Override
+    public Result<String> convertToModel(final String s, final ValueContext valueContext) {
+        return Result.ok(StringUtils.trimWhitespace(s));
+    }
+
+    @Override
+    public String convertToPresentation(final String s, final ValueContext valueContext) {
+        return StringUtils.trimWhitespace(s);
```
blomark, line 29: Please check if we need this trim call here. I would assume that the value from the database is already trimmed and we don't need to do it again.

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/utils/TrimmingStringConverter.java
```diff
@@ -0,0 +1,31 @@
...
+
+import com.vaadin.data.Converter;
+import com.vaadin.data.Result;
+import com.vaadin.data.ValueContext;
+
+/**
+ * Converter to trim whitespaces for an input e.g. to further convert it into an
+ * Integer, etc.
+ */
+public class TrimmingStringConverter implements Converter<String, String> {
```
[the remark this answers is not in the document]
herdt-michael: Good point! I have extended the usage of this converter with https://github.com/eclipse/hawkbit/pull/1329/commits/3573fc4c2a0a3e1d2fef22170e08b5c84beb4c3d

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/utils/TrimmingStringConverter.java
herdt-michael: Verified: Yes, it is trimmed. But we have a problem with null values here. In case the value is not present in the database, we should provide an empty string here which can be set as value on the TextField. Therefore I have extended the logic perform a null check as well.

blomark (approved):

Issue #1249, opened 2022-04-21
Limitations on Deployment HTTP Range Request Download


Bastian-Krause:
Hi,

we are using hawkBit with [RAUC](https://github.com/rauc/rauc) in [rauc-hawkbit-updater](https://github.com/rauc/rauc-hawkbit-updater). In a recent PR, [streaming installation](https://github.com/rauc/rauc-hawkbit-updater/pull/130) of update bundles is implemented: rauc-hawkbit-updater passes the authentication header and the hawkBit deployment download URL to RAUC. RAUC performs HTTP Range Requests (~128 KB range per request, due to [NBD](https://www.kernel.org/doc/html/latest/admin-guide/blockdev/nbd.html) reads per [squashfs block](https://www.kernel.org/doc/html/latest/filesystems/squashfs.html), see [RAUC's streaming docs](https://rauc.readthedocs.io/en/latest/advanced.html#http-streaming)) while installing the update bundle.

While testing this, I hit these hawkBit errors:

```
Registered DOS attack! Client [0:0:0:0:0:0:0:1] is above configured READ request threshold (200)!
Cannot assign 1 ActionStatus entities to Action '1' because of the configured quota limit 1000. Currently, there are 1000 ActionStatus entities assigned.
```

That means hawkBit's default configuration limits HTTP Range Requests effectively to ~1000 requests per action and 200 requests per second. For the time being, I disabled both limitations via:

```
--hawkbit.server.security.dos.filter.enabled=false
--hawkbit.server.security.dos.maxStatusEntriesPerAction=-1
```

While this works, I am not so sure if this is the desired approach for our use case. What do you think?

Furthermore, one ActionStatus entity per Range Request is generated, meaning a medium sized update bundle of 150 MB generates >1000 ActionStatuses when requested in 128 KB chunks. Could you elaborate why hawkBit needs to generate an ActionStatus per Range Request? I could not find reasoning in the commits/PRs. Could these Range Request ActionStatuses be dropped maybe? Or made configurable somehow?

Bastian-Krause:
A comment on this issue would still be very welcome.

I'd be willing to create a PR to change the ActionStatus behavior for HTTP Range Requests, but only if you tell me whether to drop ActionStatuses for this case entirely or make them configurable.

MartB:
@Bastian-Krause Is this still a limitation?
Would be great to get a comment from the responsible maintainers!

[...]

PR #1331, opened 2023-03-08
hawkbit-ddi-resource: do not log range requests

zyga:
HawkBit creates ActionStatus entries for each GET request associated with an
artefact of an assigned software module. Amongst others, this severely breaks
RAUC block-based adaptive updates [1], which issue a flurry of range requests
for small chunks, so that the whole image does not have to be streamed in case
of incremental system changes.
    
[1] https://rauc.readthedocs.io/en/latest/advanced.html#block-based-adaptive-update-block-hash-index

Fixes: https://github.com/eclipse/hawkbit/issues/1249

### DIFF
--- a/hawkbit-rest/hawkbit-ddi-resource/src/main/java/org/eclipse/hawkbit/ddi/rest/resource/DdiRootController.java
+++ b/hawkbit-rest/hawkbit-ddi-resource/src/main/java/org/eclipse/hawkbit/ddi/rest/resource/DdiRootController.java
@@ -201,38 +201,36 @@ public class DdiRootController implements DdiRootControllerRestApi {
             if (ifMatch != null && !HttpUtil.matchesHttpHeader(ifMatch, artifact.getSha1Hash())) {
                 result = new ResponseEntity<>(HttpStatus.PRECONDITION_FAILED);
             } else {
-                final ActionStatus action = checkAndLogDownload(requestResponseContextHolder.getHttpServletRequest(),
+                final ActionStatus maybeAction = checkAndMaybeLogDownload(requestResponseContextHolder.getHttpServletRequest(),
                         target, module.getId());
-
-                final Long statusId = action.getId();
 
                 result = FileStreamingUtil.writeFileResponse(file, artifact.getFilename(), artifact.getCreatedAt(),
                         requestResponseContextHolder.getHttpServletResponse(),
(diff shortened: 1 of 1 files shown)


zyga:
NOTE: this is failing in tests that observe download progress.

MartB:
Nice job!

I dont think its important to have these range requests logged anyway, its a thing between client and server and it should always be followed up with another action status, no matter what happens. For client integration debugging purposes its much easier to just add a logger on the fly. As it is right now, this provides 0 benefits for normal users.

zyga:
I can look at removing the database entries while keeping the progress callback intact. If you have any suggestions I'm happy to learn.

zyga:
@MartB I've pushed a small update. I think making the current download progress event work is futile, with my limited knowledge of spring, it seems that the underlying class requires an non-null object and in this case it was always the ActionStatus instance that was tied to a database entry.

avgustinmm:
has conflicts. is this logging change need still actual?

cybrox:
@avgustinmm Definitely. - In fact, I would not classify this as a _logging related change_ but as a fix for a _functional issue_.

As outlined by @zyga in the original issue description, the problem is not that the log elements are useless or that there are a lot of them to look through. The problem is that right now, HawkBit simply cannot be used for range-request streamed updates at all, because too many activity items are generated for the job.

This makes HawkBit completely unusable for devices that rely on this technique to download an update for which they don't have the space or memory to cache it before installation.

```json
{
   "exceptionClass":"org.eclipse.hawkbit.repository.exception.AssignmentQuotaExceededException",
   "errorCode":"hawkbit.server.error.quota.tooManyEntries",
   "message":"Quota exceeded: Cannot assign 1 more ActionStatus entities to Action '26021'. The maximum is 1000."
}
```

zyga:
I'm not actively involved with hawkbit but I would love to see this fixed. It would unlock simpler installations to use rauc progressive updates.

avgustinmm:
@zyga, @cybrox,
thanks for your input. I agree this over detailed logging in case of range requests seems little bit useless and potentially bad.

I just extended this PR idea with skipping also action checks (could be too many, with no big sense) - #2261. This should cover this PR, so I'm closing it. If the #2261 doesn't solve the issue - please open an issue for hawkbit.

Thanks for your contribution in hawkBit.

PR #1332, opened 2023-03-14
Management API: Rollout collection to correctly expose totalTargetsPerStatus property 

stefbehl:
If the collection of rollouts is requested in representation mode "full" (e.g. by requesting "/rest/v1/actions?representation=full"), it needs to correctly expose the totalTargetsPerStatus property. Currently, all targets of the rollout are (always) returned with status "notstarted".


StefanKlt (approved):
thanks @stefbehl, looks good

PR #1339, opened 2023-03-28
Scan only latest image and library vulnerabilities

herdt-michael:


stefbehl (approved):
Looks good. Thanks @herdt-michael!

PR #1341, opened 2023-04-04
Add validation for deployment groups and rollouts

strailov:


hawkbit-rest/hawkbit-mgmt-resource/src/main/java/org/eclipse/hawkbit/mgmt/rest/resource/MgmtRolloutResource.java
```diff
@@ -263,6 +264,11 @@ public ResponseEntity<MgmtRolloutGroupResponseBody> getRolloutGroup(@PathVariabl
 
         final RolloutGroup rolloutGroup = rolloutGroupManagement.getWithDetailedStatus(groupId)
                 .orElseThrow(() -> new EntityNotFoundException(RolloutGroup.class, rolloutId));
+
+        if (!Objects.equals(rolloutId, rolloutGroup.getRollout().getId())) {
+            throw new ValidationException("The provided rolloutId does not match with the rolloutId in the deployment group.");
```
stefbehl, line 269: As this is a GET request which addresses the group directly, I think we should rather return a 404 Not Found (instead of a 400 Bad Request).

stefbehl (changes requested):
Thanks @strailov! Please see my review finding.

hawkbit-rest/hawkbit-mgmt-resource/src/main/java/org/eclipse/hawkbit/mgmt/rest/resource/MgmtRolloutResource.java
strailov: Applied some changes regarding this.

hawkbit-rest/hawkbit-mgmt-resource/src/main/java/org/eclipse/hawkbit/mgmt/rest/resource/MgmtRolloutResource.java
```diff
@@ -263,6 +264,11 @@ public ResponseEntity<MgmtRolloutGroupResponseBody> getRolloutGroup(@PathVariabl
 
         final RolloutGroup rolloutGroup = rolloutGroupManagement.getWithDetailedStatus(groupId)
                 .orElseThrow(() -> new EntityNotFoundException(RolloutGroup.class, rolloutId));
+
+        if (!Objects.equals(rolloutId, rolloutGroup.getRollout().getId())) {
+            throw new EntityNotFoundException("The provided rolloutId does not match with the rolloutId in the deployment group.");
```
stefbehl, line 269: I would not expose this information that the group does exist and that it belongs to a different rollout. I'd recommend to return a plain 404 with a corresponding message ("Group XYZ not found").

PR #1345, opened 2023-04-06
Management API: Full representation of rollout group must expose totalTargetsPerStatus property

stefbehl:
If the collection of rollout groups of a rollout is requested for representation mode 'full' (`/rest/v1/rollouts/<id>/deploygroups?representation=full`), the JSON representation of each group needs to expose the `totalTargetsPerStatus` property.


StefanKlt (approved):
Thanks @stefbehl, looks good

PR #1346, opened 2023-04-10
Fix CANCEL_REJECTED dmf behaviour

strailov:


stefbehl (approved):
Looks good. Thanks @strailov!

PR #1350, opened 2023-04-19
Remove dependency net._01001111:jlorem

stefbehl:


herdt-michael (approved):
Looks good, thank you @stefbehl !

PR #1351, opened 2023-04-20
Fix Spring config of RSQLUtilityTest

stefbehl:
In some test environments (e.g. the Eclipse CI infrastructure), the test case RSQLUtilityTest#correctRsqlWithOverdueMacro fails with a NullPointerException. Strangely enough the test used to pass in the scope of our CircleCI builds.

This PR delivers a fix which makes the test reliable wrt its Spring configuration (auto-wiring of beans etc.).


StefanKlt (approved):
looks good, thanks @stefbehl

PR #1357, opened 2023-04-25
Fix Dockerfile for 0.3.0M8-mysql

stefbehl:
Fixes the Dockerfile for the MySql setup of 0.3.0M8. Declares the MariaDB Java Client as a module dependency (as required for JRE 11).


floruschbaschan (approved):
LGTM, thanks @stefbehl

PR #1359, opened 2023-04-26
Set rollout error state for failed creation, starting or deletion process

herdt-michael:
Set rollout into error state in case of too many tries of starting, creating or deleting a rollout.

### DIFF
--- a/hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/AbstractRolloutExecutor.java
+++ b/hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/AbstractRolloutExecutor.java
@@ -1,0 +1,236 @@
+/**
+ * Copyright (c) 2023 Bosch.IO GmbH and others.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package org.eclipse.hawkbit.repository.jpa;
+
+import java.util.List;
+import java.util.Optional;
+import java.util.function.BiConsumer;
+import java.util.function.Consumer;
+import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
+
+import javax.persistence.EntityManager;
+
+import org.eclipse.hawkbit.repository.RepositoryProperties;
+import org.eclipse.hawkbit.repository.RolloutExecutor;
+import org.eclipse.hawkbit.repository.event.remote.entity.RolloutUpdatedEvent;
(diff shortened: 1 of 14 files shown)


hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/AbstractRolloutExecutor.java
```diff
@@ -0,0 +1,236 @@
...
+        setRolloutErrorState(rollout);
+    }
+
+    private void setRolloutErrorState(final JpaRollout rollout) {
+        if (!cleanupScheduledActions(rollout)) {
+            // could not clean up scheduled actions
+            LOGGER.warn("Could not clean up scheduled actions for rollout with id {}", rollout.getId());
+        }
+        setErrorStateForRolloutGroups(rollout);
+        rollout.setStatus(Rollout.RolloutStatus.ERROR_STARTING);
```
bogdan-bondar, line 152: method is used in all three cases: creating, starting and deleting, however status is always "Error_Starting", please adapt accordingly

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/AbstractRolloutExecutor.java
```diff
@@ -0,0 +1,236 @@
...
+    protected void deleteScheduledActions(final JpaRollout rollout) {
+        final Slice<JpaAction> scheduledActions = findScheduledActionsByRollout(rollout);
+        final boolean hasScheduledActions = scheduledActions.getNumberOfElements() > 0;
+
+        if (hasScheduledActions) {
+            try {
+                final Iterable<JpaAction> iterable = scheduledActions::iterator;
+                final List<Long> actionIds = StreamSupport.stream(iterable.spliterator(), false).map(Action::getId)
+                        .collect(Collectors.toList());
+                actionRepository.deleteByIdIn(actionIds);
```
bogdan-bondar, line 180: please consider splitting action ids in the batches of 999(max entries in IN statement), because the TRANSACTION_ACTIONS is 5000

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/AbstractRolloutExecutor.java
herdt-michael: done

avgustinmm:
@herdt-michael,
is this PR something that is to be processed? There was some time without a progress, and there are unresolved reviews.
Anyway, what is the motivation of this feature? An rollout start/delete could fail many times because of temporary DB failures/overload or whatever. Do we need to put it in error state? What are the cases when rollout isn't recoverable?
Could you provide a little bit motivation for the PR.

herdt-michael:
Not required anymore.

PR #1361, opened 2023-05-04
Fix platform-specific new line in confirmation status

avgustinmm:
In the JpaAutoConfirmationStatus#constructActionMessage format uses platform-specific new line. DeploymentManagementTest#assignmentWithAutoConfirmationWillBeHandledCorrectly test however expected "\n" - Linux new line. Thus it fails, for instance, on Windows.

This commit fixes the new line to "\n" in order to:
* have predictability and platform independence
* test to pass on all systems

The alternative is to use platform-specific new line in the test - but this would make the status a little bit indeterministic (platform dependent).


strailov (approved):
Looks good.

PR #1362, opened 2023-05-04
Enforce decoding in authentication check

strailov:
+Enforce decoding of controllerId when authentication is performed for the reverse proxy use case


avgustinmm (approved):
LGTM

PR #1370, opened 2023-06-02
Added support for cdn download url for mgmt API and tests

ramannas:
Added support for the Mgmt REST API GET /rest/v1/softwaremodules/{softwareModuleId}/artifacts/{artifactId} to support the CDN type download URL by introducing a request parameter


hawkbit-rest/hawkbit-mgmt-resource/src/main/java/org/eclipse/hawkbit/mgmt/rest/resource/MgmtSoftwareModuleResource.java
```diff
@@ -125,15 +135,21 @@ public ResponseEntity<List<MgmtArtifact>> getArtifacts(
...
-            @PathVariable("artifactId") final Long artifactId) {
+            @PathVariable("artifactId") final Long artifactId,
+            @RequestParam(value = MgmtRestConstants.ARTIFACT_DOWNLOAD_URL_TYPE, required = false) final String artifactDownloadUrlType) {
         final SoftwareModule module = findSoftwareModuleWithExceptionIfNotFound(softwareModuleId, artifactId);
 
-        final MgmtArtifact reponse = MgmtSoftwareModuleMapper.toResponse(module.getArtifact(artifactId).get());
+        final MgmtArtifact response = MgmtSoftwareModuleMapper.toResponse(module.getArtifact(artifactId).get());
         if (!module.isDeleted()) {
-            MgmtSoftwareModuleMapper.addLinks(module.getArtifact(artifactId).get(), reponse);
+            if(artifactDownloadUrlType == null || artifactDownloadUrlType == "default") {
```
tsvetalint, line 144: No matter if we have "?downloadurltype=cdn" or "?downloadurltype=cdnnnnnnnnn", the result will be the same. Why we need  "?downloadurltype=default" option. Wouldn't be enough if we have default behaviour without parameter value(parameter is null)

hawkbit-rest/hawkbit-mgmt-resource/src/test/java/org/eclipse/hawkbit/mgmt/rest/resource/MgmtSoftwareModuleResourceTest.java
```diff
@@ -592,6 +592,37 @@ void getArtifact() throws Exception {
                         "http://localhost/rest/v1/softwaremodules/" + sm.getId() + "/artifacts/" + artifact.getId())));
     }
 
+    @Test
```
tsvetalint, line 595: it will be good if we have negative test case.

hawkbit-rest/hawkbit-mgmt-resource/src/main/java/org/eclipse/hawkbit/mgmt/rest/resource/MgmtSoftwareModuleResource.java
sebastian-firsching: I agree, maybe it's not the best solution to have a String parameter here and we should use a boolean instead. Something like useCdn = true|false

hawkbit-rest/hawkbit-mgmt-resource/src/main/java/org/eclipse/hawkbit/mgmt/rest/resource/MgmtSoftwareModuleMapper.java
```diff
@@ -155,6 +161,17 @@ static void addLinks(final Artifact artifact, final MgmtArtifact response) {
...
 
+    static void addLinks(final Artifact artifact, final MgmtArtifact response,
+            final ArtifactUrlHandler artifactUrlHandler, final SystemManagement systemManagement) {
+
+        List<ArtifactUrl> urls = artifactUrlHandler.getUrls(
+                new URLPlaceholder(systemManagement.getTenantMetadata().getTenant(),
+                        systemManagement.getTenantMetadata().getId(), null, null,
+                        new URLPlaceholder.SoftwareData(artifact.getSoftwareModule().getId(), artifact.getFilename(),
+                                artifact.getId(), artifact.getSha1Hash())), ApiType.MGMT, null);
+        response.add(Link.of(urls.get(0).getRef()).withRel("download").expand());
```
sebastian-firsching, line 172: I think we shouldn't hardcode to use the first entry of the list with a fixed rel value. Instead we should do something like:
```suggestion
        urls.forEach(entry -> response.add(Link.of(entry.getRef()).withRel(entry.getRel()).expand()));
```

hawkbit-rest/hawkbit-mgmt-api/src/main/java/org/eclipse/hawkbit/mgmt/rest/api/MgmtRestConstants.java
```diff
@@ -271,6 +271,12 @@ public final class MgmtRestConstants {
      */
     public static final String AUTH_V1_REQUEST_MAPPING = BASE_V1_REQUEST_MAPPING + "/userinfo";
 
+    /**
+     * The artifact download URL type
+     */
+    public static final String ARTIFACT_DOWNLOAD_URL_TYPE = "downloadurltype";
```
stefbehl, line 277: rename to "REQUEST_PARAMETER_ARTIFACT_DOWNLOAD_URL_TYPE"

hawkbit-rest/hawkbit-rest-docs/src/test/java/org/eclipse/hawkbit/rest/documentation/MgmtApiModelProperties.java
```diff
@@ -55,10 +55,9 @@ public final class MgmtApiModelProperties {
     public static final String ARTIFACT_HASHES_SHA1 = "SHA1 hash of the artifact.";
     public static final String ARTIFACT_HASHES_MD5 = "MD5 hash of the artifact.";
     public static final String ARTIFACT_HASHES_SHA256 = "SHA256 hash of the artifact.";
-
-    public static final String ARTIFACT_DOWNLOAD_LINK = "Download link of the artifact.";
-
+    public static final String ARTIFACT_DOWNLOAD_LINK = "Download link of the artifact based on the chosen download url type parameter.";
     public static final String ARTIFACT_LIST = "List of artifacts of given software module.";
+    public static final String ARTIFACT_DOWNLOAD_URL_TYPE = "Type of the artifact download url. (can be default or cdn)";
```
stefbehl, line 60: "default" is not a good name for the value representing the default. Maybe we can call it "direct" or "mgmt"? Then we can specify it as:
public static final String ARTIFACT_DOWNLOAD_URL_TYPE = "Type of the artifact download url (can be \"direct\" (default) or \"cdn\")";

hawkbit-rest/hawkbit-rest-docs/src/test/java/org/eclipse/hawkbit/rest/mgmt/documentation/SoftwaremodulesDocumentationTest.java
```diff
@@ -368,7 +368,7 @@ public void getArtifact() throws Exception {
 
         mockMvc.perform(
                 get(MgmtRestConstants.SOFTWAREMODULE_V1_REQUEST_MAPPING + "/{softwareModuleId}/artifacts/{artifactId}",
-                        sm.getId(), artifact.getId()))
+                        sm.getId(), artifact.getId()).param("downloadurltype", "default"))
```
stefbehl, line 371: see above

stefbehl:
> @ramannas what is the idea behind this feature? Would this allow HawkBit to offer CDN-backed URLs for artifacts? Is that something that would be exposed over the DDI API or just the MGMT API (apologies if this is unclear from the code).

Yes, it allows a Mgmt API client to obtain the CDN URLs for an artifact (via Mgmt API). DDI and DMF expose CDN links by default.

zyga:
Is this coupled with a way to offload artifacts to a CDN without having to funnel all the traffic through hawkbit?

hawkbit-rest/hawkbit-mgmt-resource/src/main/java/org/eclipse/hawkbit/mgmt/rest/resource/MgmtSoftwareModuleResource.java
stefbehl: if we stick to strings, please use "mystring".equals(...)

stefbehl:
> Is this coupled with a way to offload artifacts to a CDN without having to funnel all the traffic through hawkbit?

Not directly. I think the name of the PR is a bit misleading. Actually it just introduces a URL parameter for "/rest/v1/softwaremodules/id/artifact/id" which acts as a switch to control whether or not Hawkbit's ArtifactUrlHandler plugpoint shall be invoked internally for creating the download URLs. A custom ArtifactUrlHandler bean implementation could generate links that point to an external CDN (e.g. to hold the download traffic off from the Hawkbit update server).

hawkbit-rest/hawkbit-mgmt-resource/src/main/java/org/eclipse/hawkbit/mgmt/rest/resource/MgmtSoftwareModuleResource.java
stefbehl: The difference is that, with the parameter being set to "cdn", the ArtifactUrlHandler is invoked internally to create the download URLs. So from a technical perspective, it would make sense to call the parameter "useArtifactUrlHandler=true|false". From an API usage perspective, however, sth with "CDN" might be more comprehensible...

tsvetalint (approved):

PR #1375, opened 2023-06-21
Enrich the Rollouts field with status and distribution set

denislavprinov:


hawkbit-core/src/main/java/org/eclipse/hawkbit/repository/RolloutFields.java
```diff
@@ -25,12 +29,35 @@ public enum RolloutFields implements FieldNameProvider {
...
+    }
+
+    private RolloutFields(final String fieldName, final String... subEntityAttributes) {
+        this.fieldName = fieldName;
+        this.subEntityAttributes = Arrays.asList(subEntityAttributes);
+    }
+
+    @Override
+    public List<String> getSubEntityAttributes() {
+        return subEntityAttributes;
```
avgustinmm, line 60: This way the returned list elements could be modified which modifies the enum which shall be immutable.
Better return Collections.unmodifiableList(subEntityAttributes)

hawkbit-core/src/main/java/org/eclipse/hawkbit/repository/RolloutFields.java
```diff
@@ -25,12 +29,35 @@ public enum RolloutFields implements FieldNameProvider {
...
+    /**
+     * The Distribution set field.
+     */
+    DISTRIBUTIONSET("distributionSet", DistributionSetFields.ID.getFieldName(),
+            DistributionSetFields.NAME.getFieldName(), DistributionSetFields.VERSION.getFieldName(),
+            DistributionSetFields.TYPE.getFieldName());
 
     private final String fieldName;
 
+    private List<String> subEntityAttributes;
```
avgustinmm, line 46: could be made final

avgustinmm (changes requested):

PR #1392, opened 2023-07-13
Do not retry AMQP messages which violates a quota

memoos:
We had a case where a remote device sent an action state update, that violated the quota defined in `hawkbit.server.security.dos.maxStatusEntriesPerAction`. This caused an `AssignmentQuotaExceededException`to be thrown, so AMQP message consumption was blocked (and thereby the Device Management Federation API not functional anymore) until a manual cleanup in the database.
This changes the behavior so messages which violate a quota are ignored so Device Management Federation API remains functional.


avgustinmm (approved):
LGTM

PR #1402, opened 2023-07-27
Batch system config update

denislavprinov:


hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaTenantConfigurationManagement.java
```diff
@@ -133,45 +143,65 @@ public <T> T getGlobalConfigurationValue(final String configurationKeyName, fina
     }
 
     @Override
-    @CacheEvict(value = "tenantConfiguration", key = "#configurationKeyName")
-    @Transactional
-    @Retryable(include = {
-            ConcurrencyFailureException.class }, maxAttempts = Constants.TX_RT_MAX, backoff = @Backoff(delay = Constants.TX_RT_DELAY))
     public <T extends Serializable> TenantConfigurationValue<T> addOrUpdateConfiguration(
```
avgustinmm, line 146: Without annotation of this method this wont work as expected. The spring proxy mechanism won't apply annotations of internally called method. See [Understanding the correct use of Declarative Transaction](https://www.logicbig.com/tutorials/spring-framework/spring-data-access-with-jdbc/correct-use-of-declarative-transaction.html) 
You should have _@Transactional_ and _@Retriable_ on both methods.
Maybe the clear approach would be to have third private method that will do the work and both methods to be annotated and directly to call the private method.

hawkbit-rest/hawkbit-mgmt-api/src/main/java/org/eclipse/hawkbit/mgmt/rest/api/MgmtTenantManagementRestApi.java
```diff
@@ -83,4 +86,20 @@ ResponseEntity<MgmtSystemTenantConfigurationValue> getTenantConfigurationValue(
...
+     *           a Map of name - value pairs for the configurations
+     *
+     * @return if the given configurations values exists and could be get HTTP OK.
+     *         In any failure the JsonResponseExceptionHandler is handling the
+     *         response.
+     */
+    @PutMapping(value = MgmtRestConstants.SYSTEM_V1_REQUEST_MAPPING + "/configs", consumes = {
+            MediaTypes.HAL_JSON_VALUE, MediaType.APPLICATION_JSON_VALUE }, produces = { MediaTypes.HAL_JSON_VALUE,
+            MediaType.APPLICATION_JSON_VALUE })
+    ResponseEntity<List<MgmtSystemTenantConfigurationValue>> updateTenantConfiguration(
```
avgustinmm, line 102: @produces / returns-> Map<String, Object>, not list

hawkbit-rest/hawkbit-mgmt-resource/src/main/java/org/eclipse/hawkbit/mgmt/rest/resource/MgmtTenantManagementResource.java
```diff
@@ -76,4 +78,22 @@ public ResponseEntity<MgmtSystemTenantConfigurationValue> updateTenantConfigurat
         return ResponseEntity.ok(MgmtTenantManagementMapper.toResponse(keyName, updatedValue));
     }
 
+    @Override
+    public ResponseEntity<List<MgmtSystemTenantConfigurationValue>> updateTenantConfiguration(
```
avgustinmm, line 82: @produces / returns-> Map<String, Object>, not list

avgustinmm (approved):

PR #1403, opened 2023-07-28
Fixing missing API doc

avgustinmm:
fixed by adding
```
[markup]
	[markup.goldmark]
        [markup.goldmark.renderer]
            unsafe = true
```

otherwise hugo removes raw html (as the iframes) 
+ links points to concrete html files 
+ www removed from baseurl (otherwise some cors issues occur)

### DIFF
--- a/docs/config.toml
+++ b/docs/config.toml
@@ -7,7 +7,7 @@
 # http://www.eclipse.org/legal/epl-v10.html
 #
 
-baseurl = "https://www.eclipse.dev/hawkbit/"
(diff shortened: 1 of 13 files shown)


laverman:
LGTM

zyga:
Has this been released to the documentation website? For example looking at https://eclipse.dev/hawkbit/apis/ddi_api/ I don't see the API links that used to be there before.

avgustinmm:
Hi @zyga, 
we have pushed website to the hawkbit-website project but it's not yet updated on eclipse website. Working on it ...

PR #1407, opened 2023-08-02
Add `mvn install` command for docs under Windows.

rojarsmith:
Support the `mvn install` command use Msys or Git Bash to execute the compilation behavior of docs at windows.


avgustinmm:
Hi @rojarsmith, 
thanks for contribution.
Would you check [Fix doc builds on windows PR](https://github.com/eclipse/hawkbit/pull/1409) I just created if it works for you. 
It not only changes executables but also uses translated install and clean bat files on windows - hence no need to have shell script interpreters on windows. It, also, minimises duplications in both profiles.

PR #1435, opened 2023-09-26
hawkbit-core: fix leftover after last typo

zyga:
Apologies for missing this @avgustinmm, it was staged but not committed.


avgustinmm:
Thanks, @zyga.
I've already fixed that. If there is still any problem please let me know.

PR #1441, opened 2023-09-29
Add example values for request/response examples in swagger

strailov:


avgustinmm (approved):

strailov:
Contributor Agreement seems to be valid, not sure why the check is still pending ... Ignoring it and merging.

Issue #1083, opened 2021-02-15
Document attribute length constraint


SijmenHuizenga:
https://github.com/eclipse/hawkbit/blob/541a0129ef77281b8df5cad0a5de38cca359df1b/hawkbit-repository/hawkbit-repository-api/src/main/java/org/eclipse/hawkbit/repository/model/Target.java#L38-L46

This shows length constraints on controller attribute key and value. I couldn't find any reference to this in the documentation. I propose we add these maximum values to the api docs.

AzazelSensei:
Opened https://github.com/eclipse-hawkbit/hawkbit/pull/3341 with the 128-character key and value limits on the DDI API page.

PR #1443, opened 2023-09-30
 Added Document attribute length constraint #1083

Fragnite-hub:
Updated the documentation , added the maximum constraint in key and value size #1083

### DIFF
--- a/hawkbit-repository/hawkbit-repository-api/src/main/java/org/eclipse/hawkbit/repository/model/Target.java
+++ b/hawkbit-repository/hawkbit-repository-api/src/main/java/org/eclipse/hawkbit/repository/model/Target.java
@@ -37,12 +37,14 @@ public interface Target extends NamedEntity {
     int ADDRESS_MAX_SIZE = 512;
 
     /**
-     * Maximum length of key of controller attribute
+     * Maximum length of key of controller attribute.
+     * The maximum controller attribute key size is 128.
      */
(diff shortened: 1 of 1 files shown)


Fragnite-hub:
could you tell me the issue as in why the test cases are not passed since this is my first contribution .

Fragnite-hub:
hi admins can you look at my PR.

avgustinmm:
Thanks for the PR. 
However, I don't see the value this commit brings.
 
The actual value of the constants is part of the documentation anyway - the value of the CONTROLLER_ATTRIBUTE_VALUE_SIZE is documented at constant-values.html#org.eclipse.hawkbit.repository.model.Target.CONTROLLER_ATTRIBUTE_VALUE_SIZE. There is a fine way for checking the constant's value in javadoc out of the box.

I see few problems with this PR:
* document something that already is in the documentation - no need for PR
* if the constant value is changed someone shall take care to modify comment also - break's single point of truth 
* in the same class some constants has value mentioned in the field description, some not - lack of consistency

Fragnite-hub:
So what we have to do to resolve #1083 . We can't just leave an issue open

avgustinmm:
I suppose the proposal of #1083 was to add these constraints in the hawkbit website documentation. Where, the DDI API is documented - not in the javadoc.
Otherwise, if the proposal is for adding just javadoc comments I would close it with the same argumentation as for this PR

PR #1445, opened 2023-10-02
Fix handling of artifact range requests

zyga:
Artifact range requests in HawkBit were impossible to do efficiently, as the underlying mechanism had no choice but to read the start byte and discard content both before and after.

To address this situation, we need to introduce a new method to DbArtifact: getFileInputStream(long, long), which maps to the semantics of a range request.

The next step is to adjust FileStreamingUtil to use it for both simple range and multi-part requests.

### DIFF
--- a/hawkbit-artifact-repository-filesystem/src/main/java/org/eclipse/hawkbit/artifact/repository/ArtifactFilesystem.java
+++ b/hawkbit-artifact-repository-filesystem/src/main/java/org/eclipse/hawkbit/artifact/repository/ArtifactFilesystem.java
@@ -11,12 +11,17 @@ package org.eclipse.hawkbit.artifact.repository;
 
 import java.io.BufferedInputStream;
 import java.io.File;
+import java.io.RandomAccessFile;
+import java.nio.channels.Channels;
 import java.io.FileInputStream;
 import java.io.FileNotFoundException;
+import java.io.IOException;
 import java.io.InputStream;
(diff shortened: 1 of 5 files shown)


hawkbit-artifact-repository-filesystem/src/main/java/org/eclipse/hawkbit/artifact/repository/ArtifactFilesystem.java
```diff
@@ -47,4 +52,19 @@ public InputStream getFileInputStream() {
...
     }
+
+    @Override
+    // suppress warning, this InputStream needs to be closed by the caller, this
+    // cannot be closed in this method
+    @SuppressWarnings("squid:S2095")
+    public InputStream getFileInputStream(long start, long end) {
+        try {
+            var f = new RandomAccessFile(file, "r");
+            var ch = f.getChannel();
```
avgustinmm, line 63: I'd rather not add an additional dependency on com.google.common
wouldn't 
```
f.seek(start);
return f;
```
do the same?

avgustinmm (changes requested):
Hi @zyga,
thanks for your contribution. Sorry, for the delayed review.
I did the review and do have some comments.

avgustinmm:
Closing as outdated - no communication and including conflicts. If still interested in PR I could reopen it or new one could be created

PR #1448, opened 2023-10-05
Introducing an entity access control management

herdt-michael:
**This feature is still under development!**

-------

This pull request introduces a new feature to restrict access to entities. It is solved by an access control management that is used on the JpaManagement layer of each "controlled" entity. It allows to extend the JPA specifications in order to additionally filter the request against the database. A standard implementation that does not yet restrict access exists and is initialized in the application context. However, it provides the opportunity to introduce further access control implementations to limit the results based on different criteria.


avgustinmm:
Replaced by https://github.com/eclipse/hawkbit/pull/1456 which lays over these commits

PR #1449, opened 2023-10-06
Add portRequest url handler property

flobz:
The option `server.forward-headers-strategy=NATIVE` allow to dynamicly generate link based on X-Forwarded-Proto header.
Unfortunately Hawkbit artifact use a custom url handler. This PR allow to serve artifacts behind a reverse proxy on http and https in the same time.


docs/content/concepts/authentication.md
```diff
@@ -100,14 +100,10 @@ Nginx doesn't support obtaining the issuer hash without addons. Therefore, in th
...
-   You need to add the given download settings to your hawkBit configurations so that hawkBit can generate the URLs according to the https that the client will use to download. If you're deploying hawkBit as a Docker container, add these configurations as environmental values in the docker-compose.yml file. Don't forget to replace "hostname" with your actual hostname.
+   You need to add the given setting to your hawkBit configurations so that hawkBit can generate the URLs according to the https that the client will use to download. If you're deploying hawkBit as a Docker container, add these configurations as environmental values in the docker-compose.yml file.
 
 ```
-hawkbit.artifact.url.protocols.download-http.rel=download-http
-hawkbit.artifact.url.protocols.download-http.hostname=hawkbit.dev.example
-hawkbit.artifact.url.protocols.download-http.protocol=https
-hawkbit.artifact.url.protocols.download-http.supports=DMF,DDI
-hawkbit.artifact.url.protocols.download-http.ref={protocol}://{hostnameRequest}/{tenant}/controller/v1/{controllerId}/softwaremodules/{softwareModuleId}/artifacts/{artifactFileName}
+server.forward-headers-strategy=NATIVE
```
avgustinmm, line 106: you actually don't need that config for download url - right?
so, this section could be removed or be updated that this configuration is needed in order for the hawkBit, in general be able to us forward headers?

avgustinmm (changes requested):

docs/content/concepts/authentication.md
avgustinmm: Hi @flobz, 
sorry for the delay. Thanks for the contribution.
I did a review and this is the only think I'm not so sure about.
Could you revision the documentation fragment.
Otherwise, it looks good to me and I could merge it.

flobz: Hi @avgustinmm ,
Sorry I didn't understand, you suggest that "server.forward-headers-strategy=NATIVE"  isn't needed ?

avgustinmm: I mean that this property is not actually needed while hawkBit resolve the download url - right? So, it is not fitting under:
```
You need to add the given setting to your hawkBit configurations so that hawkBit can generate the URLs according to the https that the client will use to download.
```
Is it really needed, anyway? What differs if it is not set? If so, could be cited with different description. 
I think that that configuration that you removed is not needed anymore - it was used to change http to https which is now done with support of protocolRequest. But is this new property needed and with what purpose?

flobz: If server.forward-headers-strategy=NATIVE is not set request to  https://localhost:32805/DEFAULT/controller/v1/controller_id return href with "http" instead of "https" because spring ignore the  X-Forwarded-Proto header.
```json
  {
   "config" : {
     "polling" : {
       "sleep" : "00:00:30"
     }
   },
   "_links" : {
     "deploymentBase" : {
       "href" : 
 "http://localhost:32805/DEFAULT/controller/v1/test-121439.276369783/deploymentBase/1?c=411599879"
     },
     "configData" : {
       "href" : 
 "http://localhost:32805/DEFAULT/controller/v1/test-121439.276369783/configData"
     }
   }
 }
```

avgustinmm (approved):

docs/content/concepts/authentication.md
avgustinmm: ok. seems like a needed property to get correct requestUri.

avgustinmm:
lgtm
@flobz, thanks for the contribution

PR #1452, opened 2023-10-12
ContextAware & AccessController features

avgustinmm:
Adds new concepts to Eclipse hawkBit
* _ContextAware_ (extending TenantAware) - a central entity used to get information about the current user as well as to switch contexts. Access contexts for auto assignments and rollouts are kept so the exact/fine-grained context to be applied for the background operations)
* _AccessController_ - a mechanism that provides means to plug in additional resource access limitations. E.g. the user could have TARGET_READ permissions but the _AccessController_ could, based on some rules, to restrict the access to some targets.


avgustinmm:
closed in favour or another approach

PR #1456, opened 2023-10-20
Feature/ctx aware and access controller2

avgustinmm:
This PR provides the following new features:
* **ContextAware** - extends _TenantAware_ providing means for propagating and storing custom / fine-grained contexts (not only tenant, tenant & user). Integrated with Auto Assignments and Rollouts background jobs.
* **AccessControl(Management)** , <font color="red">experimental</font>, plug-in endpoint, at JPA layer, that provides option to be defined entity based permission checks and restrictions. 
* <font color="red">**Experimental**</font> Implementation for the AccessControl support for the some of the repositories/entity types - artifacts, actions, software modules, distribution sets, targets and target filters. 
Notes: 
  - access to an entity, grants transitively READ access to the sub entities. E.e. if you get Target instance then you will be able to get its installed or assigned DistributionSet nevertheless you have READ access to them or not. Note, that in future we could change that (or just add as an option) to apply READ restrictions to the transitively linked objects.
  - metadata access is restricted by the main entities also. I.e. to modify software module metadata you have to be able to update the software module the metadata is attached to


hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/repository/BaseEntityRepositoryACM.java
```diff
@@ -0,0 +1,350 @@
...
+            throw new InsufficientPermissionException();
+        }
+        repository.deleteById(id);
+    }
+
+    @Override
+    public void deleteAllById(@NonNull final Iterable<? extends Long> ids) {
+        final List<Long> idList = toList(ids);
+        if (count(AccessController.Operation.DELETE, byIdsSpec(idList)) != count(AccessController.Operation.READ, byIdsSpec(idList))) {
+            throw new InsufficientPermissionException("Has at least one id that is not allowed for deletion!");
```
herdt-michael, line 143: Should we compare the counting results against the amount of provided id's as well?

avgustinmm: Actually maybe the best approach would be to compare only with idList.size instead of readable count
At the moment this check provides check that all that could be read  are with delete rights
However there could be inaccessible that would be deleted. 
What do you think to compare to idList.size?

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/OnlineDsAssignmentStrategy.java
```diff
@@ -122,9 +127,26 @@ public void closeActiveActions(final List<List<Long>> targetIds) {
...
-                set, System.currentTimeMillis(), currentUser, tIds));
-
+        final long now = System.currentTimeMillis();
+        targetIds.forEach(targetIdsChunk -> {
+            if (targetRepository.count(AccessController.Operation.UPDATE, targetRepository.byIdsSpec(targetIdsChunk)) != targetIdsChunk.size()) {
+                throw new InsufficientPermissionException("No update access to all targets!");
+            }
+            targetRepository.setAssignedDistributionSetAndUpdateStatus(TargetUpdateStatus.PENDING,
+                    set, now, currentUser, targetIdsChunk);
+            // TODO AC - current problem with this approach is that the caller detach the targets and seems doesn't save them
```
herdt-michael, line 137: Then let's stick to the old approach and perform the access control check before?

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/repository/BaseEntityRepositoryACM.java
herdt-michael: Yes, let's change it to compare the sizes DELETABLE.size == idList.size. However, let's map the `Iterable<? extends Long> ids` into a set instead of a `List<Long>` to get rid of potential duplicates.

herdt-michael (commented):
I haven't finished the complete review yet. However, I would like to push an intermediate status.

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaRolloutHandler.java
```diff
@@ -92,19 +94,29 @@ private static String createRolloutLockKey(final String tenant) {
...
         DeploymentHelper.runInNewTransaction(txManager, handlerId + "-" + rolloutId, status -> {
             rolloutManagement.get(rolloutId).ifPresentOrElse(
-                    rollout -> runInUserContext(rollout, () -> rolloutExecutor.execute(rollout)),
+                    rollout -> {
+                        rollout.getAccessControlContext().ifPresentOrElse(
+                            context -> // has stored context - executes it with it
+                                contextAware.runInContext(
+                                    context,
+                                    () -> rolloutExecutor.execute(rollout)),
+                            () -> // has no stored context - executes it in the tenant & user scope
```
herdt-michael, line 108: At this place I remembered a consequence of restoring the context again.

Let's assume the users permissions changed in the meanwhile, we will anyway set the permission present during rollout creation. I think this behaviour is fine, however it needs to be documented.

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaDistributionSetManagement.java
```diff
@@ -494,51 +549,62 @@ private Specification<JpaDistributionSetMetadata> byDsIdSpec(final long dsId) {
...
+                    }
+                    return distributionSetRepository
+                            .findOne(DistributionSetSpecification.byActionId(actionId))
+                            .orElseThrow(() ->
+                                    new InsufficientPermissionException("DistributionSet not accessible (or not found)!"));
+                })
+                .map(DistributionSet.class::cast)
+                .or(() -> {
+                    throw new EntityNotFoundException(Action.class, actionId);
+                });
```
herdt-michael, line 595: In generell I would not throw an InsufficientPermissionException in case of missing read rights. Instead, I would filter the stream (is target found) and provide an empty Optional in case the associate target is not accessible.

herdt-michael: Example:

```suggestion
        return actionRepository.findById(actionId)
                .filter(action -> targetRepository.exists(TargetSpecifications.hasId(action.getTarget().getId())))
                .map(action -> distributionSetRepository.findOne(DistributionSetSpecification.byActionId(actionId))
                        .orElseThrow(() -> new InsufficientPermissionException(
                                "DistributionSet not accessible (or not found)!")))
                .map(DistributionSet.class::cast).or(() -> {
                    throw new EntityNotFoundException(Action.class, actionId);
                });
```

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaDistributionSetManagement.java
```diff
@@ -757,10 +787,27 @@ public DistributionSet getValid(final long id) {
 
     @Override
     @Transactional
-    public void invalidate(final DistributionSet set) {
-        final JpaDistributionSet jpaSet = (JpaDistributionSet) set;
+    public void invalidate(final DistributionSet distributionSet) {
```
herdt-michael, line 790: We should extend the DistributionSet interface by the #invalidate function.

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaDistributionSetTagManagement.java
```diff
@@ -95,33 +100,34 @@ public Optional<DistributionSetTag> getByName(final String name) {
...
         final Collection<JpaTagCreate> creates = (Collection) dst;
 
-        return Collections.unmodifiableList(
-                creates.stream().map(create -> distributionSetTagRepository.save(create.buildDistributionSetTag()))
-                        .collect(Collectors.toList()));
+        return creates.stream()
+                .map(create -> distributionSetTagRepository
+                        .save(AccessController.Operation.CREATE, create.buildDistributionSetTag()))
+                .map(DistributionSetTag.class::cast)
+                .toList();
```
herdt-michael, line 118: I would suggest performing a bulk insert here:

    public List<DistributionSetTag> create(final Collection<TagCreate> dst) {
        final List<JpaDistributionSetTag> toCreate = dst.stream().map(JpaTagCreate.class::cast)
                .map(JpaTagCreate::buildDistributionSetTag).toList();
        return Collections
                .unmodifiableList(distributionSetTagRepository.saveAll(AccessController.Operation.CREATE, toCreate));
    }

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaSoftwareModuleManagement.java
```diff
@@ -202,19 +205,23 @@ public Optional<SoftwareModule> get(final long id) {
...
+                        SoftwareModuleSpecification.fetchType()))
+                .map(SoftwareModule.class::cast);
     }
 
     private void deleteGridFsArtifacts(final JpaSoftwareModule swModule) {
+        // clearArtifactBinary validates unconditionally the module access
         for (final Artifact localArtifact : swModule.getArtifacts()) {
-            artifactManagement.clearArtifactBinary(localArtifact.getSha1Hash(), swModule.getId());
+            ((JpaArtifactManagement)artifactManagement)
+                    .clearArtifactBinary(localArtifact.getSha1Hash(), swModule.getId());
```
herdt-michael, line 224: Lets extend the ArtifactManagement interface by the `clearArtifactBinary` function to avoid the cast here.

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaSoftwareModuleManagement.java
```diff
@@ -236,7 +242,9 @@ public void delete(final Collection<Long> ids) {
             // delete binary data of artifacts
             deleteGridFsArtifacts(swModule);
 
-            if (isUnassigned(swModule.getId())) {
+            // execute this count operation without access limitations since we have to
+            // ensure it's not assigned when deleting it.
+            if (distributionSetRepository.countByModulesId(swModule.getId()) <= 0) {
```
herdt-michael, line 247: Since this call will be executed against the repository, the access restriction is not applied, correct? Wondering if we should wrapt those kind of executions into an "system context" execution.

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaSoftwareModuleManagement.java
```diff
@@ -358,25 +377,33 @@ public Slice<AssignedSoftwareModule> findAllOrderBySetAssignmentAndModuleNameAsc
...
+        smWithAssignedFlagList.forEach(smWithAssignedFlag -> {
+            final JpaSoftwareModule softwareModule = smWithAssignedFlag.get("sm", JpaSoftwareModule.class);
+            try {
+                softwareModuleRepository.findById(softwareModule.getId()); // check read access
+                resultList
+                        .add(new AssignedSoftwareModule(softwareModule,
+                                smWithAssignedFlag.get("assigned", Number.class).longValue() == 1));
+            } catch (final InsufficientPermissionException e) {
+                // skip the entry
+            }
```
herdt-michael, line 389: What about perform an exists query using a specification by ID like:

```suggestion
            // check accessibility
            final JpaSoftwareModule softwareModule = smWithAssignedFlag.get("sm", JpaSoftwareModule.class);
            if (softwareModuleRepository.exists(SoftwareModuleSpecification.byId(softwareModule.getId()))) {
                resultList.add(new AssignedSoftwareModule(softwareModule,
                        smWithAssignedFlag.get("assigned", Number.class).longValue() == 1));
            }
```

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaTargetManagement.java
```diff
@@ -542,49 +539,45 @@ public TargetTypeAssignmentResult assignType(final Collection<String> controller
...
         allTargets.forEach(target -> target.setTargetType(null));
 
-        return new TargetTypeAssignmentResult(0, Collections.emptyList(), Collections
-                .unmodifiableList(allTargets.stream().map(targetRepository::save).collect(Collectors.toList())), null);
+        return new TargetTypeAssignmentResult(0, Collections.emptyList(), targetRepository.saveAll(allTargets), null);
     }
 
     private List<JpaTarget> findTargetsByInSpecification(final Collection<String> controllerIds,
             final Specification<JpaTarget> specification) {
         return Lists.partition(new ArrayList<>(controllerIds), Constants.MAX_ENTRIES_IN_STATEMENT).stream()
```
herdt-michael, line 558: What about using a parallel stream here?

```suggestion
        return Lists.partition(new ArrayList<>(controllerIds), Constants.MAX_ENTRIES_IN_STATEMENT).parallelStream()
```

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/repository/BaseEntityRepositoryACM.java
```diff
@@ -0,0 +1,350 @@
+/**
+ * Copyright (c) 2015 Bosch Software Innovations GmbH and others
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
```
herdt-michael, line 9: Fix copyright header

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/repository/DistributionSetTypeRepository.java
```diff
@@ -76,30 +54,18 @@ public interface DistributionSetTypeRepository
...
-     */
-    @Override
-    @Query("SELECT d FROM JpaDistributionSetType d WHERE d.id IN ?1")
-    List<JpaDistributionSetType> findAllById(Iterable<Long> ids);
-
     /**
      * Counts the {@link SoftwareModuleType}s which are associated with the
      * addressed {@link DistributionSetType}.
+     * <p/>
+     * No access control applied
```
herdt-michael, line 68: I like this kind of markers you added on repository level, nice! 👍

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/repository/HawkBitBaseRepository.java
```diff
@@ -0,0 +1,134 @@
+/**
+ * Copyright (c) 2021 Bosch.IO GmbH and others
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.hawkbit.repository.jpa.repository;
```
herdt-michael, line 10: Fix copyright header

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/rollout/window/layouts/AddRolloutWindowLayout.java
```diff
@@ -134,6 +134,7 @@ private void updateTotalTargets() {
     }
 
     private Long getTotalTargets(final String filterQuery, final Long distSetTypeId) {
+        // TODO AC -  Check for updatable targets only
```
herdt-michael, line 137: Open todo here

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/management/targettag/targettype/AddTargetTypeWindowController.java
```diff
@@ -81,6 +86,9 @@ protected Class<? extends ProxyIdentifiableEntity> getParentEntityClass() {
 
     @Override
     protected boolean isEntityValid(final ProxyTargetType entity) {
-        return validator.isEntityValid(entity, () -> targetTypeManagement.getByName(entity.getName()).isPresent());
+        return validator.isEntityValid(entity,
+                () -> contextAware.runAsTenant(
+                        contextAware.getCurrentTenant(),
+                        () -> targetTypeManagement.getByName(entity.getName()).isPresent()));
```
herdt-michael, line 92: contextAware#runAsTenant based on contextAware#getCurrentTenant. I assume this will temporary deactivate the access restrictions. However this looks confusing for me, since the affect is not recognizable

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/management/targettable/UpdateTargetWindowController.java
```diff
@@ -98,10 +105,14 @@ protected Class<ProxyTarget> getEntityClass() {
     @Override
     protected boolean isEntityValid(final ProxyTarget entity) {
         final String controllerId = entity.getControllerId();
-        return proxyTargetValidator.isEntityValid(entity, () -> hasControllerIdChanged(controllerId)
-                && targetManagement.getByControllerID(controllerId).isPresent());
+        return proxyTargetValidator.isEntityValid(entity,
+                () -> contextAware.runAsTenant(
+                        contextAware.getCurrentTenant(),
+                        () -> hasControllerIdChanged(controllerId)
+                                && targetManagement.getByControllerID(controllerId).isPresent()));
```
herdt-michael, line 112: contextAware#runAsTenant based on contextAware#getCurrentTenant. I assume this will temporary deactivate the access restrictions. However this looks confusing for me, since the affect is not recognizable

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/management/targettag/targettype/AddTargetTypeWindowController.java
herdt-michael: Maybe we should wrap protected `boolean isEntityValid(final ProxyTargetType entity)` per default in an unrestricted execution

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/repository/BaseEntityRepositoryACM.java
```diff
@@ -0,0 +1,350 @@
...
+                        try {
+                            // TODO - cache mapping so to speed things
+                            final Method delegateMethod =
+                                    BaseEntityRepositoryACM.class.getDeclaredMethod(
+                                            method.getName(), method.getParameterTypes());
+                            return delegateMethod.invoke(repositoryACM, args);
+                        } catch (final NoSuchMethodException e) {
+                            // call to repository itself
+                        }
+                        if (method.getName().startsWith("find") || method.getName().startsWith("get")) {
```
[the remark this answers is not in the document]
avgustinmm: yes, probably would be better. However, maybe this could be part of consequent PR

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/repository/BaseEntityRepositoryACM.java
```diff
@@ -0,0 +1,350 @@
...
+                                return ((Optional<T>)result).map(t -> isOperationAllowed(AccessController.Operation.READ, t, accessController));
+                            } else if (repository.getDomainClass().isAssignableFrom(method.getReturnType())) {
+                                accessController.assertOperationAllowed(AccessController.Operation.READ, (T)result);
+                            }
+                            return result;
+                        } else if ("toString".equals(method.getName()) && method.getParameterCount() == 0) {
+                            return BaseEntityRepositoryACM.class.getSimpleName() +
+                                    "(repository: " + repository + ", accessController: " +  accessController + ")";
+                        } else {
+                            return method.invoke(repository, args);
```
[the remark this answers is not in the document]
avgustinmm: I think that all not covered here methods (e.g. count) shall be carefully checked IF shall be filtered agains some access control rules - these, e.g. counts, could need DELETE or UPDATE pr READ permissions - so - not easy to make something automatically. 
In general the best approach would be to remove all declarative repository methods with direct calls with specification and correct access to be checked

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaDeploymentManagement.java
```diff
@@ -883,66 +915,44 @@ public long countMessagesByActionStatusId(final long actionStatusId) {
...
-    }
-
-    @Override
-    public Slice<Action> findActionsByDistributionSet(final Pageable pageable, final long dsId) {
-        throwExceptionIfDistributionSetDoesNotExist(dsId);
-        return actionRepository.findByDistributionSetId(pageable, dsId);
-    }
-
+    // TODO - return via Mgmt API all actions (event for targets the use has no access - check if should and could
+    // be limited
```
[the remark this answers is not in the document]
avgustinmm: I'm not sure how effective an action AccessController could be implemented.
Anyway these methods will just return actions and its implicitly loaded JPA entities. 
however max read access to probably not sensitive information will be achieved.
Maybe at some later point we could add restriction on getters level - so only action will be returned but Action#getTarget will fail.
For now, I think this could stay like it is. What do you think?

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaSoftwareModuleManagement.java
```diff
@@ -202,19 +205,23 @@ public Optional<SoftwareModule> get(final long id) {
...
 
-        throwExceptionIfSoftwareModuleTypeDoesNotExist(typeId);
-
-        return softwareModuleRepository.findOneByNameAndVersionAndTypeId(name, version, typeId);
-    }
-
-    private boolean isUnassigned(final Long moduleId) {
-        return distributionSetRepository.countByModulesId(moduleId) <= 0;
+        // TODO AC : this method is used to validate input. Is it fine to restrict the
+        // access?
```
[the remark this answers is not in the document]
avgustinmm: Ok - access check will stay here. This could have problems with Vaadin UI when access control is enabled.
However, probably Vaadin UI will be removed in few months (non from community supports its migration to new Vaadin versions), probably no problem. If, Vaadin UI stay for more time we could add correspondent permissions elevation for UI if needed

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaSoftwareModuleManagement.java
```diff
@@ -248,55 +256,62 @@ public void delete(final Collection<Long> ids) {
             if (auditorProvider != null) {
                 currentUser = auditorProvider.getCurrentAuditor().orElse(null);
             }
-            softwareModuleRepository.deleteSoftwareModule(System.currentTimeMillis(), currentUser,
-                    assignedModuleIds.toArray(new Long[0]));
+
+            /*
+               TODO AC - could use single update query as before via entity manager directly (considers tenant!):
+               "UPDATE JpaSoftwareModule b SET b.deleted = 1, b.lastModifiedAt = :lastModifiedAt, b.lastModifiedBy = :lastModifiedBy WHERE b.tenant = :tenant AND b.id IN :ids"
+             */
```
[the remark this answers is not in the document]
avgustinmm: The problem is that I don't know how to do delete with single query applying access control.
There is no delete method with Specification. When migrate to Spring Boot 3, if I'm not wrong, there will be delete by spec and we could implement this as single query.

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/repository/BaseEntityRepositoryACM.java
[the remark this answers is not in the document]
herdt-michael: Fine for me.

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaRolloutHandler.java
avgustinmm: This depends on the implementation of the SecurityContextSerializater that is used.
Implementation could, let say, does not store the user permissions (as Java serializer) but just the user id, and then to get entity base permissions from a dynamic database - so the permissions of stored context could change over time.
However, they could also be static (as java serializer). So, I think no explicit need to document.
Whoever implements  access control shall chose its serializer and have in account its features

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaDistributionSetManagement.java
avgustinmm: maybe in other PR

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaDistributionSetTagManagement.java
avgustinmm: Accepted and will apply the proposal. However, I don't think that bulk import works with hawkBit anyway - since the id generation strategy. Do you think it works?

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaDistributionSetManagement.java
avgustinmm: First of all, I think that returning optional methods shall not throw not found. So, if not found shall return empty,
However, this would change the behaviour. So ..
Regarding the insufficient permissions - currently, since there is no action access control I assume that the access control for actions is provided by targets. Thus, if no access to target I assume no access to action. That's why I think here shall be thrown insufficient permissions instead of not found

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaSoftwareModuleManagement.java
avgustinmm: As wrote above I made the method even with package visibility. It not check access anymore so better be as much invisible as possible. 
Anyway no need to be public.

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaSoftwareModuleManagement.java
avgustinmm: Assuming that such "countBy" are not covered by auth acm apply they are considered to be executed in system scope. If, at some point, we add such auto for them - they usage shall be reconsidered if to be executed in system scope

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaSoftwareModuleManagement.java
avgustinmm: AFAIU the you point is skip - not to throw exception if any inaccessible. You are right - it should be like so. I change it to (skipping, possibly, db call):
```java
// check read access
if (softwareModuleRepository.getAccessController().map(accessController -> {
    try {
        accessController.assertOperationAllowed(AccessController.Operation.READ, softwareModule);
        return true;
    } catch (final InsufficientPermissionException e) {
        return false;
    }
}).orElse(true)) {
    resultList.add(new AssignedSoftwareModule(softwareModule,
            smWithAssignedFlag.get("assigned", Number.class).longValue() == 1));
}
```

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaTargetManagement.java
avgustinmm: I'm not sure. This could occupy more threads. And anyway is not part of this PR. Maybe could evaluate such changes later.

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/repository/BaseEntityRepositoryACM.java
```diff
@@ -0,0 +1,350 @@
...
+    }
+
+    @Override
+    @NonNull
+    public boolean exists(@Nullable AccessController.Operation operation, Specification<T> spec) {
+        if (operation == null) {
+            return repository.exists(spec);
+        } else {
+            return repository.exists(
+                    Objects.requireNonNull(accessController.appendAccessRules(operation, spec)));
```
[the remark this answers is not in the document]
avgustinmm: I added it just to exists since there is no @Nullable annotation. The other JpaSpecificationExecutor methods accepts nullable specification.

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/rollout/window/layouts/AddRolloutWindowLayout.java
avgustinmm: Again, since it is a part of UI which will probably be dropped for next release we could not bother to add explicit support.
We could declare that access control management won't work fluently with Vaadin UI

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/management/targettable/UpdateTargetWindowController.java
avgustinmm: Added comment. I think modified just 4 places where you already elevated access. Maybe we could add on all palaces. However, not sure we need to full integration with UI?

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaArtifactManagement.java
```diff
@@ -142,64 +162,89 @@ private InputStream wrapInEncryptionStream(final long smId, final InputStream st
...
+     * {@link SoftwareModule#getId()} or {@link SoftwareModule}'s that are
+     * marked as deleted.
+     * <p/>
+     * Software module related UPDATE permission shall be checked by the callers!
+     *
+     * @param sha1Hash no longer needed
+     * @param softwareModuleId the garbage collection call is made for
+     */
+    @PreAuthorize(SpPermission.SpringEvalExpressions.HAS_AUTH_DELETE_REPOSITORY)
+    void clearArtifactBinary(final String sha1Hash, final long softwareModuleId) {
```
herdt-michael, line 195: Provided so `softwareModuleId` is unsued

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaArtifactManagement.java
```diff
@@ -19,19 +19,28 @@
...
+import org.eclipse.hawkbit.repository.jpa.EncryptionAwareDbArtifact;
+import org.eclipse.hawkbit.repository.jpa.JpaManagementHelper;
+import org.eclipse.hawkbit.repository.jpa.acm.AccessController;
 import org.eclipse.hawkbit.repository.jpa.configuration.Constants;
 import org.eclipse.hawkbit.repository.jpa.model.JpaArtifact;
 import org.eclipse.hawkbit.repository.jpa.model.JpaSoftwareModule;
+import org.eclipse.hawkbit.repository.jpa.repository.LocalArtifactRepository;
+import org.eclipse.hawkbit.repository.jpa.repository.SoftwareModuleRepository;
+import org.eclipse.hawkbit.repository.jpa.specifications.ArtifactSpecifications;
+import org.eclipse.hawkbit.repository.jpa.specifications.SoftwareModuleSpecification;
```
herdt-michael, line 43: Unused import

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaSoftwareModuleManagement.java
herdt-michael: I don't fully understand the reasoning. Method X checks whether the binaries can be deleted (if they are not used anywhere else). This is a valid check that does not take into account the access restrictions of the current user. In my opinion, nothing that needs to be hidden. An ArtifactManagement interface is intentionally autowired to allow multiple implementations. In case of a different implementation we have issues. In my opinion, the function required to delete artifacts should therefore also be part of the (public) interface.

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaArtifactManagement.java
```diff
@@ -142,64 +162,89 @@ private InputStream wrapInEncryptionStream(final long smId, final InputStream st
...
+        if (count <= 1) { // 1 artifact is the one being deleted!
+            // removes the real artifact ONLY AFTER the delete of artifact or software module
+            // in local history has passed successfully (caller has permission and no errors)
+            TransactionSynchronizationManager.registerSynchronization(new TransactionSynchronization() {
+                @Override
+                public void afterCommit() {
+                    try {
+                        LOG.debug("deleting artifact from repository {}", sha1Hash);
+                        // TODO - if the transaction fail the artifact will, anyway, be deleted,
+                        // TODO - could be thought about moving it after the root transaction
```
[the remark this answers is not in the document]
avgustinmm: Will remove it - it seems outdated - before I moved the code into afterCommit

herdt-michael (approved):
Looks good, thank you @avgustinmm!

Issue #1348, opened 2023-04-18
DataIntegrityViolationException: ; ERROR: null value in column "initiated_by" of relation "sp_action" violates not-null constraint


forsberg:
Seeing an exception when trying to drag-and-drop a distribution onto a target.

Hawkbit version: snapshot (1b342001a2722116e6161c415f6ec769deda4ec0)
Database: PostgreSQL

Authentication/Authorization is via OIDC, which may be relevant.

Traceback:

```
staging-ferro-hawkbit-85cddbd56-j96d6 ferro-hawkbit 2023-04-18 14:30:00.534 ERROR 325 --- [qtp571956559-74] o.e.h.ui.error.HawkbitUIErrorHandler     : Unexpected Ui error occured
staging-ferro-hawkbit-85cddbd56-j96d6 ferro-hawkbit 
staging-ferro-hawkbit-85cddbd56-j96d6 ferro-hawkbit com.vaadin.server.ServerRpcManager$RpcInvocationException: Unable to invoke method click in com.vaadin.shared.ui.button.ButtonServerRpc
staging-ferro-hawkbit-85cddbd56-j96d6 ferro-hawkbit 	at com.vaadin.server.ServerRpcManager.applyInvocation(ServerRpcManager.java:159)
staging-ferro-hawkbit-85cddbd56-j96d6 ferro-hawkbit 	at com.vaadin.server.ServerRpcManager.applyInvocation(ServerRpcManager.java:116)
staging-ferro-hawkbit-85cddbd56-j96d6 ferro-hawkbit 	at com.vaadin.server.communication.ServerRpcHandler.handleInvocation(ServerRpcHandler.java:442)
staging-ferro-hawkbit-85cddbd56-j96d6 ferro-hawkbit 	at com.vaadin.server.communication.ServerRpcHandler.handleInvocations(ServerRpcHandler.java:407)
staging-ferro-hawkbit-85cddbd56-j96d6 ferro-hawkbit 	at com.vaadin.server.communication.ServerRpcHandler.handleRpc(ServerRpcHandler.java:275)
staging-ferro-hawkbit-85cddbd56-j96d6 ferro-hawkbit 	at com.vaadin.server.communication.UidlRequestHandler.synchronizedHandleRequest(UidlRequestHandler.java:91)
[...]

PR #1122, opened 2021-05-25
bugfix missing OidcUser principal class

mdaur:
This PR will fix an issue I encountered using hawkBit with OIDC authentication schema (SpringSecurity oauth2). 
If SecurityContextTenantAware.getCurrentUsername return a null value, multiple error come up across the application:
-  e.g. if you are assigning a deployment which causes a sp_action
Some issues I encountered end up in the DB layer for the latest (e.g. Caused by: java.sql.SQLException: Column 'initiated_by' cannot be null) 
This fix at least ensures that there is a Username in case of OIDC usage. 
/Martin

PR #1458, opened 2023-10-22
Fixed DataIntegrityViolationException with OIDC 

kyberpunk:
Hi. This PR fixes the bug reported in #1348. Fix was originally pushed in #1122, but author could not comply to contributing guidelines.

In my case the issue was causing that it was not possible to trigger deployment in hawkBit UI when OIDC was configured.
Screenshot 2023-10-22 162437

After fixing I've manually tested that issue is not reproducible anymore and triggering deployments works as expected. There are no exceptions in log after fix.


avgustinmm:
@mdaur, 
is it ok if we merge this PR?

kyberpunk:
> @mdaur, is it ok if we merge this PR?

https://github.com/eclipse/hawkbit/pull/1122#issuecomment-1787953890

mdaur:
Sure, please go ahead

PR #1472, opened 2023-11-08
Enable specifying target type when created using DMF API

charvadzo:
Extension of DMF API with possibility of setting target type name when creating target. If a target type with the provided name is found (was created beforehand) then it is associated with the new target.

This change is backward compatible (target type parameter is not mandatory).

More info in docs/content/apis/dmf_api.md

### DIFF
--- a/hawkbit-dmf/hawkbit-dmf-amqp/src/main/java/org/eclipse/hawkbit/amqp/AmqpMessageHandlerService.java
+++ b/hawkbit-dmf/hawkbit-dmf-amqp/src/main/java/org/eclipse/hawkbit/amqp/AmqpMessageHandlerService.java
@@ -231,14 +231,18 @@ public class AmqpMessageHandlerService extends BaseAmqpService {
             final URI amqpUri = IpUtil.createAmqpUri(virtualHost, replyTo);
             final Target target;
             if (isOptionalMessageBodyEmpty(message)) {
+                LOG.debug("Received \"THING_CREATED\" AMQP message for thing \"{}\" without body.", thingId);
                 target = controllerManagement.findOrRegisterTargetIfItDoesNotExist(thingId, amqpUri);
             } else {
                 checkContentTypeJson(message);
                 final DmfCreateThing thingCreateBody = convertMessage(message, DmfCreateThing.class);
                 final DmfAttributeUpdate thingAttributeUpdateBody = thingCreateBody.getAttributeUpdate();
 
+                LOG.debug("Received \"THING_CREATED\" AMQP message for thing \"{}\" with target name \"{}\" and type " +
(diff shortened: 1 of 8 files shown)


avgustinmm:
It seems reasonable, though the gap between DMF and DDI become bigger. At the moment DMF allows on such auto registration to specify name, DDI - not. now type is also added. Since DMF is intended to for integration with another management system - isn't an option that system (that communicate with devices) to create devices via mgmt rest api?
Then it could set all possible information - no need to support even name.
Anyway if there is consent that it is good to add this feature (passing target type on create) then at least I'd like to do not expose TargetTypeManagement#getByName to controllers (by changing the access) but instead the callers to elevate the permissions when call getByName, e.g calling method in tenant's scope

charvadzo:
Hi @avgustinmm,
you are right that we could switch to MGMT API but this is not in line with our project plan at the moment.

I've changed the code according to your proposal so I'd be glad if you could have a look on it again if it can be done this way. 

I've also improved the newly added ControllerManagement integration tests a bit to verify if there are sufficient permissions to interact with the TargetTypeManagement.

avgustinmm:
I think it could be merged if you resolve the conflicts.

avgustinmm:
Conflicts are not resolved yet

charvadzo:
All fixed now. I'm glad to see that things are moving on Hawkbit now :-)

Just a note: I had issues with the updated RolloutManagementFlowTest. It did not give me consistent results when repeatedly run on my machine. It sometimes succeeded but sometimes failed on these assertions:

https://github.com/eclipse/hawkbit/blob/master/hawkbit-repository/hawkbit-repository-jpa/src/test/java/org/eclipse/hawkbit/repository/jpa/management/RolloutManagementFlowTest.java#L81

and/or

https://github.com/eclipse/hawkbit/blob/master/hawkbit-repository/hawkbit-repository-jpa/src/test/java/org/eclipse/hawkbit/repository/jpa/management/RolloutManagementFlowTest.java#L144

I assume the problem is not connected with this PR as it failed me on master as well so maybe you can try to run it yourself few times to see if it behaves ok for you?

```
[ERROR] Failures: 
[ERROR]   RolloutManagementFlowTest.dynamicRolloutFlow:144 
expected: FINISHED
 but was: RUNNING
[INFO] 
[ERROR] Tests run: 618, Failures: 1, Errors: 0, Skipped: 0
```

avgustinmm:
Yes, indeed, we have noted the sporadic failures in the test (I think on faster machines) and I plan to fix that. Thanks, for mentioning it.
I'll take a final look and if all is ok will merge.

avgustinmm (approved):

avgustinmm:
@charvadzo, thanks for the contribution!

Issue #1475, opened 2023-11-16
Internal Server Error when accessing the frontend; IPv6 address detection too simple? "host" variable also contains the port.


marfrde:
Hello,

I have a problem accessing a newly migrated (from 0.3.0M7-mysql to 0.3.0M9-mysql) docker instance of hawkbit that is running as an Azure App Service.

The (db-) migration as such worked without issues (did the log say). Thanks for that. The issues starts when I want to access the frontend. I only get an internal server error 

```
Whitelabel Error Page
This application has no explicit mapping for /error, so you are seeing this as a fallback.

Thu Nov 16 07:51:05 GMT 2023
There was an unexpected error (type=Internal Server Error, status=500).
```

and the following stack trace in the log

```
2023-11-16T07:51:05.506463464Z: [INFO]  2023-11-16 07:51:05.506 ERROR 1 --- [nio-8080-exec-6] o.a.c.c.C.[.[.[.[vaadin4SpringServlet]   : Servlet.service() for servlet [vaadin4SpringServlet] in context with path [] threw exception
2023-11-16T07:51:05.506506964Z: [INFO]  
2023-11-16T07:51:05.506512664Z: [INFO]  java.lang.IllegalArgumentException: Malformed IPv6 address at index 8: http://[1.2.3.4:48837] // IP Changed. It was the actual public IP of the client
2023-11-16T07:51:05.506543464Z: [INFO]  	at java.base/java.net.URI.create(Unknown Source)
2023-11-16T07:51:05.506547664Z: [INFO]  	at org.eclipse.hawkbit.util.IpUtil.createUri(IpUtil.java:130)
2023-11-16T07:51:05.506551164Z: [INFO]  	at org.eclipse.hawkbit.util.IpUtil.createHttpUri(IpUtil.java:160)
2023-11-16T07:51:05.506554564Z: [INFO]  	at org.eclipse.hawkbit.util.IpUtil.getClientIpFromRequest(IpUtil.java:96)
2023-11-16T07:51:05.506558164Z: [INFO]  	at org.eclipse.hawkbit.util.IpUtil.getClientIpFromRequest(IpUtil.java:79)
2023-11-16T07:51:05.506561664Z: [INFO]  	at org.eclipse.hawkbit.security.DosFilter.doFilterInternal(DosFilter.java:125)
[...]

PR #1483, opened 2023-11-21
#1475 fix IP resolution

avgustinmm:


strailov (approved):

hawkbit-security-core/src/test/java/org/eclipse/hawkbit/util/IpUtilTest.java
```diff
@@ -11,9 +11,7 @@
 
 import static com.google.common.net.HttpHeaders.X_FORWARDED_FOR;
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.mockito.Mockito.times;
-import static org.mockito.Mockito.verify;
-import static org.mockito.Mockito.when;
+import static org.mockito.Mockito.*;
```
strailov, line 14: I would escape wild card imports even in tests.

avgustinmm: will fix it

PR #1488, opened 2023-11-27
Improved Docker build and docker images

avgustinmm:
* Using upacked Spring layers for faster boot
* Added option to pass JAVA_OPT to modify the environment variables in both flavours
* Build support for different hawkbit apps - could be used in future if hawkBit is split in multiple micro-services
* Removed templates and replacing with parameterized (via build ARGs) docker builds


hawkbit-runtime/docker/dockerfiles/generate.sh
strailov, line 1: why not leave the shell script ?

avgustinmm: Because it is used to replace some values in templates
no templates are completely drop out. Versions are to be directly passed using build args or replaced in docker files (as defaults)

strailov (approved):
Seems to work as expected. 
Looks good.

PR #1504, opened 2023-12-07
Added an option to show a footer notification

denislavprinov:
To show a footer notification and stylise it, add this application property: 
hawkbit.server.ui.notification.text=...

**Example:**

hawkbit.server.ui.notification.text=`<div style='background-color: yellow; color: #282828; font-weight: bold;'>Notification !</div>`


avgustinmm (approved):
LGTM

PR #1666, opened 2024-02-23
Improve hawkBit user management

avgustinmm:
1. Definded with properties users (static) are configured using property map (no need of indexes)
2. AuthenticationProvider that authenticates them is always registered (if not needed - don't configure them)
3. UserDetailService (in case of missing - won't be registered)
4. Spring security user (spring.security.username) will be registered together with other users (if any). If any - it will be system-wide, otherwise tenant-scoped.
5. UserPrincipal renamed to TenantAwareUser in order to match its purpose.
6. Some if its fields are removes as not needed - to be closer to spring security user


hawkbit-autoconfigure/src/main/java/org/eclipse/hawkbit/autoconfigure/security/TenantAwareUserProperties.java
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * Configuration for hawwkBit static users.
+ */
+@Data
+@ToString
+@ConfigurationProperties("hawkbit.security.user")
+public class TenantAwareUserProperties {
+
+    private Map<String, User> users = new HashMap<>();
```
vasilchev, line 29: what is the key of this map supposed to be ? tenant or ?
in some places of code it is taken as username rather then tenant

hawkbit-autoconfigure/src/main/java/org/eclipse/hawkbit/autoconfigure/security/InMemoryUserManagementAutoConfiguration.java
```diff
@@ -11,129 +11,155 @@
...
-
-            final UserPrincipal userPrincipal = new UserPrincipal(user.getUsername(), user.getPassword(),
-                    user.getFirstname(), user.getLastname(), user.getUsername(), user.getEmail(), DEFAULT_TENANT,
-                    authorityList);
+    private static UserDetailsService userDetailsService(
+            final SecurityProperties securityProperties,
+            final TenantAwareUserProperties tenantAwareUserProperties,
+            final PasswordEncoder passwordEncoder) {
+        final List<User> userPrincipals = new ArrayList<>();
+        tenantAwareUserProperties.getUsers().forEach((username, user) -> {
```
vasilchev, line 78: here tenantAwareUserProperties#getUsers is supposed to be a map<String, Users[]> where key is tenant and values are colleciton of users, right :?
but here key of the map is taken as username, rather than tenant

hawkbit-autoconfigure/src/main/java/org/eclipse/hawkbit/autoconfigure/security/TenantAwareUserProperties.java
vasilchev: if that is the case maybe better naming of this field would be more clear - something like 'tenantUsers'

hawkbit-autoconfigure/src/main/java/org/eclipse/hawkbit/autoconfigure/security/InMemoryUserManagementAutoConfiguration.java
```diff
@@ -11,129 +11,155 @@
...
-            final List<GrantedAuthority> authorityList = roles.isEmpty() ? PermissionUtils.createAllAuthorityList()
-                    : createAuthoritiesFromList(roles);
             userPrincipals
-                    .add(new UserPrincipal(name, password, name, name, name, null, DEFAULT_TENANT, authorityList));
+                    .add(new TenantAwareUser(
+                            securityProperties.getUser().getName(),
+                            password(securityProperties.getUser().getPassword(), passwordEncoder),
+                            createAuthorities(
+                                    securityProperties.getUser().getRoles(), PermissionUtils::createAllAuthorityList),
+                            DEFAULT_TENANT));
```
vasilchev, line 95: here we always put DEFAULT_TENANT ?

hawkbit-autoconfigure/src/main/java/org/eclipse/hawkbit/autoconfigure/security/InMemoryUserManagementAutoConfiguration.java
```diff
@@ -11,129 +11,155 @@
...
+    private static UserDetailsService userDetailsService(
+            final SecurityProperties securityProperties,
+            final TenantAwareUserProperties tenantAwareUserProperties,
+            final PasswordEncoder passwordEncoder) {
+        final List<User> userPrincipals = new ArrayList<>();
+        tenantAwareUserProperties.getUsers().forEach((username, user) -> {
+            final TenantAwareUser userPrincipal = new TenantAwareUser(
+                    username, password(user.getPassword(), passwordEncoder),
+                    createAuthorities(user.getRoles(), Collections::emptyList),
+                    DEFAULT_TENANT);
```
vasilchev, line 82: using DEFAULT_TENANT always :?
havent seen tenantAwareUserProperties#getUsers()#keys usage - shouldn't we take tenant from there ?

hawkbit-autoconfigure/src/main/java/org/eclipse/hawkbit/autoconfigure/security/InMemoryUserManagementAutoConfiguration.java
avgustinmm: yes. in case there are "tenant users" we treat spring user as "global / non-tenant user" - such are already defined
if there are no "tenant users" - we that it as default tenant user for default tenant DEFAULT

hawkbit-autoconfigure/src/main/java/org/eclipse/hawkbit/autoconfigure/security/TenantAwareUserProperties.java
vasilchev: aah, i see, ok

hawkbit-autoconfigure/src/main/java/org/eclipse/hawkbit/autoconfigure/security/InMemoryUserManagementAutoConfiguration.java
avgustinmm: good catch. indeed this is a bug. It's forgotten to add tenant (as intended) as user property and to set here.
fixed

hawkbit-autoconfigure/src/main/java/org/eclipse/hawkbit/autoconfigure/security/InMemoryUserManagementAutoConfiguration.java
vasilchev: have we tested this with different usernames ? maybe some symbols are not allowed here, or are not properly parsed from spring properties ?

avgustinmm: since it is default "spring" approach (e.g. spring.security.user.<username>) all usernames that are accepted as spring users will be accepted here - which shall be enough

hawkbit-security-core/src/main/java/org/eclipse/hawkbit/im/authentication/TenantAwareUser.java
```diff
@@ -0,0 +1,78 @@
...
+        this.tenant = tenant;
+    }
+
+    /**
+     * Create user without password and any credentials. For test purposes only.
+     *
+     * @param username the username of the user
+     * @param tenant the tenant of the user
+     */
+    public TenantAwareUser(final String username, String tenant) {
```
vasilchev, line 55: this constructor is not used anywhere ?

avgustinmm: SecurityContextSwitch:166

vasilchev (approved):

PR #1669, opened 2024-02-29
Feature/make amqp handler conditional

strailov:


avgustinmm:
no need of explicitly pointing out the class

PR #1721, opened 2024-04-24
prevent npe during multipart exception handling

mwiesenbauer:


mwiesenbauer:
You are right. Changing the behaviour was not my intention.

PR #1766, opened 2024-07-11
Improve building of SQL from an RSQL query

avgustinmm:
* ignore case behavior could be disabled
* like is used only when needed


strailov (approved):
Looks good to me.

vasilchev (approved):

denislavprinov (approved):

PR #1771, opened 2024-07-16
Remove target type to target reference

avgustinmm:
it is not used and could lead to extensive memory usage if JPA provider load targets

### DIFF
--- a/hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/model/JpaTargetType.java
+++ b/hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/model/JpaTargetType.java
@@ -9,6 +9,8 @@
  */
 package org.eclipse.hawkbit.repository.jpa.model;
 
+import lombok.NoArgsConstructor;
+import lombok.ToString;
 import org.eclipse.hawkbit.repository.event.remote.TargetTypeDeletedEvent;
 import org.eclipse.hawkbit.repository.event.remote.entity.TargetTypeCreatedEvent;
 import org.eclipse.hawkbit.repository.event.remote.entity.TargetTypeUpdatedEvent;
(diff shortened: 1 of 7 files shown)


hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/model/JpaTargetType.java
```diff
@@ -56,96 +59,53 @@ public class JpaTargetType extends AbstractJpaTypeEntity implements TargetType,
...
     /**
-     * @param dsSetType
-     *          Distribution set type
+     * Adds a compatible distribution set type to this target type.
+     *
+     * @param dsSetType Distribution set type to add
      * @return Target type
      */
-    public void addCompatibleDistributionSetType(final DistributionSetType dsSetType) {
+    public JpaTargetType addCompatibleDistributionSetType(final DistributionSetType dsSetType) {
```
vasilchev, line 72: not sure why this was changed :?
return type is not used

vasilchev (approved):

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/model/JpaTargetType.java
avgustinmm: why not? makes API fluent. It provides option to add fixed set of types fluently, e.g. in tests.
also make it in sync which removeDistributionSetType - also returns 'this' without being actually used.
Both could be made to return void in future

vasilchev: i don't mind it, whenever i see return this i think of it as a builder/chain pattern. just noticed that with the change it was never used/referenced

PR #1796, opened 2024-07-31
Replace http code 200 by 201 for POST and PUT

diorcety:
openapi-generator-cli generator (for python for example) uses the responses code for mapping reply from request.
For POST and PUT request the "default" http code is 201 and not 200, producing incorrect parsing (getting None instead of the returned object(s))

### DIFF
--- a/hawkbit-rest/hawkbit-mgmt-api/src/main/java/org/eclipse/hawkbit/mgmt/rest/api/MgmtDistributionSetRestApi.java
+++ b/hawkbit-rest/hawkbit-mgmt-api/src/main/java/org/eclipse/hawkbit/mgmt/rest/api/MgmtDistributionSetRestApi.java
@@ -164,7 +164,7 @@ public interface MgmtDistributionSetRestApi {
             "distribution sets within Hawkbit. The request body must always be a list of sets. " +
             "Required permission: CREATE_REPOSITORY")
     @ApiResponses(value = {
-        @ApiResponse(responseCode = "200", description = "Successfully retrieved"),
+        @ApiResponse(responseCode = "201", description = "Successfully created"),
         @ApiResponse(responseCode = "400", description = "Bad Request - e.g. invalid parameters", 
                 content = @Content(mediaType = "application/json", schema = @Schema(implementation = ExceptionInfo.class))),
         @ApiResponse(responseCode = "401", description = "The request requires user authentication.", 
@@ -711,7 +711,7 @@ public interface MgmtDistributionSetRestApi {
     @Operation(summary = "Create a list of meta data for a specific distribution set", description = "Create a list " +
             "of meta data entries Required permissions: READ_REPOSITORY and UPDATE_TARGET")
     @ApiResponses(value = {
-        @ApiResponse(responseCode = "200", description = "Successfully retrieved"),
+        @ApiResponse(responseCode = "201", description = "Successfully created"),
         @ApiResponse(responseCode = "400", description = "Bad Request - e.g. invalid parameters", 
                 content = @Content(mediaType = "application/json", schema = @Schema(implementation = ExceptionInfo.class))),
         @ApiResponse(responseCode = "401", description = "The request requires user authentication.", 
--- a/hawkbit-rest/hawkbit-mgmt-api/src/main/java/org/eclipse/hawkbit/mgmt/rest/api/MgmtDistributionSetTagRestApi.java
+++ b/hawkbit-rest/hawkbit-mgmt-api/src/main/java/org/eclipse/hawkbit/mgmt/rest/api/MgmtDistributionSetTagRestApi.java
@@ -153,7 +153,7 @@ public interface MgmtDistributionSetTagRestApi {
     @Operation(summary = "Creates new Distribution Set Tags", description = "Handles the POST request of creating " +
             "new distribution set tag. The request body must always be a list of distribution set tags.")
     @ApiResponses(value = {
-        @ApiResponse(responseCode = "200", description = "Successfully retrieved"),
+        @ApiResponse(responseCode = "201", description = "Successfully created"),
         @ApiResponse(responseCode = "400", description = "Bad Request - e.g. invalid parameters", 
                 content = @Content(mediaType = "application/json", schema = @Schema(implementation = ExceptionInfo.class))),
         @ApiResponse(responseCode = "401", description = "The request requires user authentication.", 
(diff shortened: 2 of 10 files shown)


strailov:
Hello @diorcety ! 
Thanks for this PR! Can you please signe the Eclipse Contributor Agreement so we can proceed with it and merge it ? 
Thanks!

hawkbit-rest/hawkbit-ddi-api/src/main/java/org/eclipse/hawkbit/ddi/rest/api/DdiRootControllerRestApi.java
```diff
@@ -373,7 +373,7 @@ ResponseEntity<Void> postDeploymentBaseActionFeedback(@Valid final DdiActionFeed
         information that will allow the server to identify the device on a hardware level (e.g. hardware revision,
         mac address, serial number etc.).""")
     @ApiResponses(value = {
-        @ApiResponse(responseCode = "200", description = "Successfully retrieved"),
+        @ApiResponse(responseCode = "201", description = "Successfully updated"),
```
strailov, line 376: this PUT method does not create a new content - it just updates current configuration data. So I would stick with 200 here.

hawkbit-rest/hawkbit-ddi-api/src/main/java/org/eclipse/hawkbit/ddi/rest/api/DdiRootControllerRestApi.java
```diff
@@ -755,7 +755,7 @@ ResponseEntity<Void> postConfirmationActionFeedback(@Valid final DdiConfirmation
         be automatically confirmed, as long as auto-confirmation is active.
         """)
     @ApiResponses(value = {
-        @ApiResponse(responseCode = "200", description = "Successfully retrieved"),
+        @ApiResponse(responseCode = "201", description = "Successfully created"),
```
strailov, line 758: If you check the implementation here activateAutoConfirmation returns 200 : 

return new ResponseEntity<>(HttpStatus.OK);

hawkbit-rest/hawkbit-ddi-api/src/main/java/org/eclipse/hawkbit/ddi/rest/api/DdiRootControllerRestApi.java
```diff
@@ -798,7 +798,7 @@ ResponseEntity<Void> activateAutoConfirmation(@PathVariable("tenant") final Stri
         while all future actions need to be confirmed, before processing with the deployment.
         """)
     @ApiResponses(value = {
-        @ApiResponse(responseCode = "200", description = "Successfully retrieved"),
+        @ApiResponse(responseCode = "201", description = "Successfully created"),
```
strailov, line 801: Same for deactivateAutoConfirmation

hawkbit-rest/hawkbit-mgmt-api/src/main/java/org/eclipse/hawkbit/mgmt/rest/api/MgmtDistributionSetRestApi.java
```diff
@@ -237,7 +237,7 @@ Query fields based on the Feed Item Query Language (FIQL). See Entity Definition
     @Operation(summary = "Update Distribution Set", description = "Handles the UPDATE request for a single " +
             "Distribution Set. Required permission: UPDATE_REPOSITORY")
     @ApiResponses(value = {
-        @ApiResponse(responseCode = "200", description = "Successfully retrieved"),
+        @ApiResponse(responseCode = "201", description = "Successfully updated"),
```
strailov, line 240: update distribution set does not create a new entity.

hawkbit-rest/hawkbit-mgmt-api/src/main/java/org/eclipse/hawkbit/mgmt/rest/api/MgmtDistributionSetRestApi.java
```diff
@@ -635,7 +635,7 @@ ResponseEntity<MgmtMetadata> getMetadataValue(@PathVariable("distributionSetId")
     @Operation(summary = "Update single meta data value of a distribution set", description = "Update a single meta " +
             "data value for speficic key. Required permission: UPDATE_REPOSITORY")
     @ApiResponses(value = {
-        @ApiResponse(responseCode = "200", description = "Successfully retrieved"),
+        @ApiResponse(responseCode = "201", description = "Successfully updated"),
```
strailov, line 638: Update metadata does not create new entity.

hawkbit-rest/hawkbit-mgmt-api/src/main/java/org/eclipse/hawkbit/mgmt/rest/api/MgmtDistributionSetRestApi.java
```diff
@@ -757,7 +757,7 @@ ResponseEntity<List<MgmtMetadata>> createMetadata(@PathVariable("distributionSet
         always be a list of software module IDs. Required permissions: READ_REPOSITORY and UPDATE_REPOSITORY
         """)
     @ApiResponses(value = {
-        @ApiResponse(responseCode = "200", description = "Successfully retrieved"),
+        @ApiResponse(responseCode = "201", description = "Successfully created"),
```
strailov, line 760: Assigning a SM to a DS also returns status 200 in the implementation.

strailov: I think you got my point here - the other changes also should be double checked with implementation logic.

diorcety:
Yeah sorry, it was a long day yesterday, I didn't think correctly about some changes, I will double check each PUT/POST to check the correct return code in each case

diorcety:
@strailov Should we update Api description according to the current method returned code, and/or change the method returned code.
For example, currently POST methods create some entity but the method returns a 200, shouldn't be a 201? (example postConfirmationActionFeedback)

strailov:
postConfirmationActionFeedback indeed may create some entities but they are related more to the backend - method return type is ResponseEntity<Void> and it is more like a feedback channel

diorcety:
@strailov okay, i'm not really aware of all the behaviour of hawkbit. I modified my PR, changing only ApiResponse to 201 according to the current implementation uing HttpStatus.CREATED

strailov (approved):

strailov:
Looks good! Thank you @diorcety .

PR #1821, opened 2024-08-15
Propose SDK Refactor

vasilchev:


hawkbit-sdk/hawkbit-sdk-device/src/main/java/org/eclipse/hawkbit/sdk/device/DdiTenant.java
```diff
@@ -40,17 +41,26 @@ public void destroy() {
...
-    public DdiController create(final Controller controller, final UpdateHandler updateHandler) {
+    public DdiController createController(final Controller controller, final UpdateHandler updateHandler) {
         final DdiController ddiController = new DdiController(tenant, controller, updateHandler, hawkbitClient);
         controllers.put(controller.getControllerId(), ddiController);
         return ddiController;
     }
 
-    public void remove(final String controllerId) {
-        Optional.ofNullable(controllers.remove(controllerId)).ifPresent(DdiController::stop);
+    public void deleteController(final String controllerId) {
```
avgustinmm, line 50: I'm not sure that deletion from the backend is a part of DdiTenant - this entity represents DDI devices and should not do whatever they don't/can't do

hawkbit-sdk/hawkbit-sdk-device/pom.xml
```diff
@@ -35,5 +35,11 @@
          <artifactId>hawkbit-ddi-api</artifactId>
          <version>${project.version}</version>
       </dependency>
+
```
avgustinmm, line 38: maybe should not be used for DDI which is simulating device side

hawkbit-sdk/hawkbit-sdk-demo/src/main/java/org/eclipse/hawkbit/sdk/demo/device/DeviceApp.java
```diff
@@ -54,44 +52,38 @@ HawkbitClient hawkbitClient(
...
         private final ScheduledExecutorService scheduler = Executors.newSingleThreadScheduledExecutor();
 
-        Shell(final Tenant tenant, final DdiController device, final HawkbitClient hawkbitClient) {
-            this.tenant = tenant;
-            this.device = device;
-            this.hawkbitClient = hawkbitClient;
+        Shell(final DdiTenant ddiTenant, final Optional<UpdateHandler> updateHandler) {
+            this.ddiTenant = ddiTenant;
+            String controllerId = System.getProperty("demo.controller.id");
+            String securityToken = System.getProperty("demo.controller.securityToken");
```
avgustinmm, line 72: better get from spring properties - so it could be configured via application properties and so on

hawkbit-sdk/hawkbit-sdk-device/src/main/java/org/eclipse/hawkbit/sdk/device/DdiTenant.java
```diff
@@ -40,17 +41,26 @@ public void destroy() {
...
+        Optional.ofNullable(controllers.remove(controllerId)).ifPresent(removedController -> hawkbitClient.mgmtService(MgmtTargetRestApi.class, tenant).deleteTarget(removedController.getControllerId()));
```
strailov, line 51: Do we need indeed to mix Mgmt API calls here ? 
Ddi Controller is supposed to connect, poll and handle updates eventually. Not sure if we need to add deletion behaviour here. Of course I'm open for discussions about it.

hawkbit-sdk/hawkbit-sdk-dmf/src/main/java/org/eclipse/hawkbit/sdk/dmf/amqp/DmfSender.java
```diff
@@ -51,8 +51,16 @@ public DmfSender(final RabbitTemplate rabbitTemplate, final AmqpProperties amqpP
     }
 
     public void createOrUpdateThing(final String tenant, final String controllerId) {
+        sendThingMessage(tenant, controllerId, MessageType.THING_CREATED.name());
+    }
+
+    public void removeThing(final String tenant, final String controllerId) {
```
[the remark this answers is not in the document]
vasilchev: unregisterThing suits as it sends a THING_REMOVED 'request' message to delete the provisioned target

hawkbit-sdk/hawkbit-sdk-device/src/main/java/org/eclipse/hawkbit/sdk/device/SetupHelper.java
```diff
@@ -7,7 +7,7 @@
  *
  * SPDX-License-Identifier: EPL-2.0
  */
-package org.eclipse.hawkbit.sdk.demo;
+package org.eclipse.hawkbit.sdk.device;
```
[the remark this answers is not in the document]
vasilchev: Idea was to have everything you need to setup/work with devices in one module - sdk device,
meaning anything that needs to implement and work with devices to have everything there (This particular SetupHelper used inside DDiTenant so that one can work directly with its DDiTenant in order to setup the devices)

hawkbit-sdk/hawkbit-sdk-device/src/main/java/org/eclipse/hawkbit/sdk/device/DdiTenant.java
vasilchev: my impression is that DdiTenant / DmfTenant are entities that represent Client side logic.
meaning anything you would like to do as a Client (regarding the devices) you can do through these entities.

it is same as removeControlelr (after review unregisterThing on DmfController)

if we it is the case that control of devices (unregister) should be outside these modules, i would remove them

hawkbit-sdk/hawkbit-sdk-device/pom.xml
vasilchev: yes.. this is needed for the Authentication setup (feign client to get the security tokens/gateway tokens).. so that sdk-device has the utilities to setup/integrate device.

We could separate it (this utility SetupHelper) in different module meaning - if one wants to work with the SDK and have that nice functionality to have the 2 dependencies
-module1 sdk-device
-module2 wherever we decide to put the SetupHelper

hawkbit-sdk/hawkbit-sdk-device/src/main/java/org/eclipse/hawkbit/sdk/device/DdiTenant.java
vasilchev: yeah.. maybe we should discuss the scope of DDITenant/DMFTenant and what to have as sdk 
i personally would like to have such functionalities within the sdk of devices (even though it does call some management apis.. but in the end it is related to the devices)

It also makes sense to separate it.. but as SDK to have only option to register without removing is somewhat half sdk to me..

hawkbit-sdk/hawkbit-sdk-device/src/main/java/org/eclipse/hawkbit/sdk/device/DdiTenant.java
vasilchev: agreed, moved mgmt related api calls in new module hawkbit-sdk-mgmt where necessary mgmt calls are placed, hawkbit-sdk-device remains only device related

strailov (approved):

PR #1846, opened 2024-09-20
Quota for ActionStatus push to be checked only for intermediate statu…

vasilchev:
…ses. Always allow terminate/final status

### DIFF
--- a/hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaActionManagement.java
+++ b/hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaActionManagement.java
@@ -9,11 +9,6 @@
  */
 package org.eclipse.hawkbit.repository.jpa.management;
 
-import java.util.ArrayList;
-import java.util.Comparator;
-import java.util.List;
-import java.util.stream.Collectors;
-
 import lombok.extern.slf4j.Slf4j;
 import org.eclipse.hawkbit.repository.QuotaManagement;
 import org.eclipse.hawkbit.repository.RepositoryProperties;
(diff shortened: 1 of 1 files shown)


avgustinmm:
I'm not sure if it is a good idea. This way device could bypass limit and create unlimited statuses.
Anyway, if the device has reached the limit, something shall be done to clean up the action statuses.
If we allow action to be closed - what next? We next update will start and device again will be unable to report ..

vasilchev:
> I'm not sure if it is a good idea. This way device could bypass limit and create unlimited statuses. Anyway, if the device has reached the limit, something shall be done to clean up the action statuses. If we allow action to be closed - what next? We next update will start and device again will be unable to report ..

There is an initial check whenever status comes if action is still active (meaning no final status yet reported)

Once reported final status - no further statuses intermediate/final could be reported

avgustinmm:
> > I'm not sure if it is a good idea. This way device could bypass limit and create unlimited statuses. Anyway, if the device has reached the limit, something shall be done to clean up the action statuses. If we allow action to be closed - what next? We next update will start and device again will be unable to report ..
> 
> There is an initial check whenever status comes if action is still active (meaning no final status yet reported)
> 
> Once reported final status - no further statuses intermediate/final could be reported.
> 
> Behavior of device is indeed unpredictable, but we should accept final status so that action can be reported as completed, otherwise we miss the final status and action is in pending state until manual close (even though action has completed, we just reject the final status)

yes, but this won't percent new action and device could user only final statuses. And if it use intermediate - next action will be broken and hang again. So, this won't fix the problem - device has too many action statuses - they shall be clean up

avgustinmm:
aha, if the quota is per action - then should be ok, if you add a check that the action statuses won't become more than quota + 1 or if it is carefully checked than in all flows someone has already checks the action is active (as you said).
Otherwise, again the device could create too many statuses (finished).
Note: even if you have checked that all flows check that action is finished - it's good to add quote + 1 check because in future new flows could be added or existing flows could be changed

avgustinmm (approved):

PR #2024, opened 2024-11-13
add flywaydb postgre dependency

samos667:
Since 0.6.1, hawkbit with postgre is not yet support if not specified it in dependency https://github.com/flyway/flyway/issues/3909#issuecomment-2188173863

We only have postgre databases in our infra and we will want to keep it that way.

Thanks in advance !

### DIFF
--- a/hawkbit-repository/hawkbit-repository-jpa/pom.xml
+++ b/hawkbit-repository/hawkbit-repository-jpa/pom.xml
@@ -62,6 +62,10 @@
             <artifactId>flyway-mysql</artifactId>
         </dependency>
         <dependency>
+             <groupId>org.flywaydb</groupId>
+             <artifactId>flyway-database-postgresql</artifactId>
+        </dependency>
+        <dependency>
             <groupId>cz.jirutka.rsql</groupId>
             <artifactId>rsql-parser</artifactId>
         </dependency>


avgustinmm:
Hi @samos667,

we could not accept your contribution because you haven't signed ECA, @see the failed check.

Then I wonder if we shall add flyway optional dependencies in into hawkbit-repository-jpa.
If you are using hawkbit as dependency in your project you could just add postgres support jar into your dependencies. 
If you are using hawkbit runtimes - as a jar or as dockers - then I suppose this dependency shall be added into monolith hawkbit runtime only.

Could you share more about how are you using hawkbit?

samos667:
Hi @avgustinmm 
thanks for you response.

Sorry I didn't see about ECA.

We use docker image for now. We started using it at 0.5.0 with postgre 16.4. But when trying to upgrade, when using hawkbit-server image 0.6.1, there is flyway error about postgre version but this a about postgre usage itself and seems related to flyway deps declaration. 

Is there a way to easily enable Postgre support for Docker image?

avgustinmm:
Hi,
the standard docker images contain H2, Postgresql, and MSSQL drivers. The one with MySQL has explicitly added MySQL driver.
So, having that they already have Postgresql and MSSQL drivers, I suppose, we should add there (maybe via monolith runtime jar) also flyway-database-postgresql and flyway-sqlserver (?).

For now, to make 0.6.1 work with PostgreSQL you could create a docker image with flyway-database-postgresql by yourself. There are two easy ways I see:
* create a docker image FROM the official hawkbit 0.6.1 by adding into /BOOT-INF/lib/ the flyway-database-postgresql 
* or to create a docker image from scratch. You could check how [Dockerfile-mysql](https://github.com/eclipse-hawkbit/hawkbit/blob/master/docker/docker_build/Dockerfile-mysql) adds MySQL driver to the hawkbit-update-server and add flyway-database-postgresql instead.

samos667:
Thanks for the info, I really appreciate it !

We'll build from the official one, that's the best option at the moment.

So I will close this PR as I think my change is not the way. And I'm not really sure (as a deploy manager) how to do it right.

samos667:
For others peoples who use postgre like us(on a kubernetes env):

We don't have yet a production container build chain.
So we have add a initContainer with the same image to our Deployment 

empyDir is mount on /tmp/lib/ on initContainer and do a copy of all /BOOT-INF/lib/ to /tmp/lib and wget https://repo.maven.apache.org/maven2/org/flywaydb/flyway-database-postgresql/10.10.0/flyway-database-postgresql-10.10.0.jar to /tmp/lib

After init, the "real" hawkbit container mount the emptyDir on /BOOT-INF/lib/ in readOnly

samos667:
@avgustinmm  do you thing this will be added on maven deps in a future release ?

avgustinmm:
Hi @samos667,
yes. Yesterday, I added the flyway PostgreSQL support to the monolith and mgmt servers (jar).  So, it will be available in the next release (including docker images). I also added example docker compose files for PostgreSQL in https://github.com/eclipse-hawkbit/hawkbit/tree/master/docker.

samos667:
I really appreciate it, thanks for that !

PR #2025, opened 2024-11-13
Move Flyway into separate artifact

avgustinmm:
Now it is not part of the ddi and dmf starters - only mgmt (where could be excluded on packaging)


vasilchev (approved):
lgtm

PR #2081, opened 2024-11-18
Switch from deprecated EnableGlobalMethodSecurity to EnableMethodSecu…

avgustinmm:
…rity

### DIFF
--- a/hawkbit-rest-core/src/main/java/org/eclipse/hawkbit/rest/SecurityManagedConfiguration.java
+++ b/hawkbit-rest-core/src/main/java/org/eclipse/hawkbit/rest/SecurityManagedConfiguration.java
@@ -20,13 +20,11 @@ import org.eclipse.hawkbit.rest.security.DosFilter;
 import org.eclipse.hawkbit.security.HawkbitSecurityProperties;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.boot.web.servlet.FilterRegistrationBean;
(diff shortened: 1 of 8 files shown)


hawkbit-rest-core/src/main/java/org/eclipse/hawkbit/rest/SecurityManagedConfiguration.java
```diff
@@ -39,7 +37,6 @@
 @Slf4j
 @Configuration
 @EnableWebSecurity
-@EnableGlobalMethodSecurity(prePostEnabled = true, mode = AdviceMode.ASPECTJ, proxyTargetClass = true, securedEnabled = true)
```
vasilchev, line 42: isnt here a change in behavior ? mode = AdviceMode.ASPECTJ wheres by default mode is PROXY ?

avgustinmm: @PreAuthenticate annotations are used only in repository and management API
that's why the EnableMethod security is moved only in repository and management API configurations

avgustinmm: and regarding removal of ASPECTJ - it seems this mode doesn't work - and it was the server app annotations with  PROXY mode that really makes things to work.

vasilchev (approved):
lgtm

PR #2177, opened 2025-01-08
Move deprecated repository and mgmt rest methods in separate module

avgustinmm:
Move deprecated repository and mgmt rest methods in separate module

Some already deprecated management REST methods are moved in separate module (together with used only for them repository api and impl) in order to have cleanly separate deprecatd REST API.
The new module is hawkbit-mgmt-resource-deprecated. It is inculded, by default, in hawkbit-mgmt-stater.
* when we decide to remove the deprecated REST API implementation completely - will be easily remved - just module and refs
* deprecated REST API could be excluded (by removing the module from runtime) even before that for the runtimes.
* after removal, for some time (untill the usad management and repository APIs are compatible) it will be possible to refer (and include) the deprecated method implementation together with the next hawkBit versions.

The deprecated methods are:
* POST /rest/v1/distributionsettags/{distributionsetTagId}/assigned/toggleTagAssignment
* POST /rest/v1/distributionsettags/{distributionsetTagId}/assigned
* POST /rest/v1/targettags/{targetTagId}/assigned/toggleTagAssignment
* POST /rest/v1/targettags/{targetTagId}/assigned


strailov (approved):
Looks good.

PR #2248, opened 2025-01-28
Introduce parallel rollout processing

denislavprinov:
- Introduce a ThreadPoolTaskExecutor for parallel rollout execution
- Introduce a BlockWhenFullPolicy RejectedExecutionHandler to block scheduler submissions if all threads (default : 2) are occupied.

### DIFF
--- a/hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/RepositoryApplicationConfiguration.java
+++ b/hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/RepositoryApplicationConfiguration.java
@@ -167,6 +167,7 @@ import org.eclipse.hawkbit.tenancy.configuration.TenantConfigurationProperties;
 import org.eclipse.hawkbit.utils.TenantConfigHelper;
 import org.springframework.beans.BeansException;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Value;
 import org.springframework.beans.factory.config.BeanPostProcessor;
 import org.springframework.boot.autoconfigure.AutoConfigureAfter;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
@@ -189,6 +190,7 @@ import org.springframework.integration.support.locks.LockRegistry;
 import org.springframework.lang.NonNull;
 import org.springframework.retry.annotation.EnableRetry;
 import org.springframework.scheduling.annotation.EnableScheduling;
+import org.springframework.scheduling.concurrent.ThreadPoolTaskExecutor;
 import org.springframework.transaction.PlatformTransactionManager;
 import org.springframework.transaction.annotation.EnableTransactionManagement;
 import org.springframework.validation.beanvalidation.MethodValidationPostProcessor;
@@ -262,6 +264,20 @@ public class RepositoryApplicationConfiguration {
                 return BeanPostProcessor.super.postProcessAfterInitialization(bean, beanName);
             }
         };
+    }
+
+    @Bean
+    public ThreadPoolTaskExecutor rolloutTaskExecutor(
+        @Value("${hawkbit.rollout.executor.thread-pool.size:2}") int poolSize) {
+        ThreadPoolTaskExecutor executor = new ThreadPoolTaskExecutor();
+        executor.setCorePoolSize(poolSize);
+        executor.setMaxPoolSize(poolSize);
+        executor.setQueueCapacity(0); // forces a Synchronous Queue
+        // This policy will block the submitter until a worker thread is free
+        executor.setRejectedExecutionHandler(new BlockWhenFullPolicy());
+        executor.setThreadNamePrefix("rollout-exec-");
+        executor.initialize();
+        return executor;
     }
 
     @Bean
@@ -992,8 +1008,8 @@ public class RepositoryApplicationConfiguration {
     @Profile("!test")
     @ConditionalOnProperty(prefix = "hawkbit.rollout.scheduler", name = "enabled", matchIfMissing = true)
     RolloutScheduler rolloutScheduler(final SystemManagement systemManagement,
-            final RolloutHandler rolloutHandler, final SystemSecurityContext systemSecurityContext) {
-        return new RolloutScheduler(rolloutHandler, systemManagement, systemSecurityContext);
+                                      final RolloutHandler rolloutHandler, final SystemSecurityContext systemSecurityContext, final ThreadPoolTaskExecutor rolloutTaskExecutor) {
+        return new RolloutScheduler(rolloutHandler, systemManagement, systemSecurityContext, rolloutTaskExecutor);
     }
 
     /**
--- a/hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/rollout/RolloutScheduler.java
+++ b/hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/rollout/RolloutScheduler.java
@@ -14,6 +14,7 @@ import org.eclipse.hawkbit.repository.RolloutHandler;
 import org.eclipse.hawkbit.repository.SystemManagement;
 import org.eclipse.hawkbit.security.SystemSecurityContext;
 import org.springframework.scheduling.annotation.Scheduled;
+import org.springframework.scheduling.concurrent.ThreadPoolTaskExecutor;
 
 /**
  * Scheduler to schedule the {@link RolloutHandler#handleAll()}. The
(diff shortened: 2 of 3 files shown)


hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/rollout/RolloutScheduler.java
```diff
@@ -44,12 +48,29 @@ public RolloutScheduler(
...
-            systemManagement.forEachTenant(tenant -> rolloutHandler.handleAll());
+            // workaround eclipselink that is currently not possible to
+            // execute a query without multi-tenancy if MultiTenant
+            // annotation is used.
+            // https://bugs.eclipse.org/bugs/show_bug.cgi?id=355458. So
+            // iterate through all tenants and execute the rollout check for
+            // each tenant seperately.
+
+            systemManagement.forEachTenant(tenant ->
+                rolloutTaskExecutor.submit(() ->
```
vasilchev, line 62: if there is no queue for this rolloutTaskExecutor and we always block - meaning only 2 tenants are processed - isnt't there are case where randomly only given Tenants are always processed but others are rarely 

Executor is ran every 2seconds-> 
Lets say we have 50 tenants

On first run of the runningRolloutScheduler() -> this will start the systemManagement.forEachTenant -> rolloutTaskExecutor.submint()

While this is running (while still processing tenants of current run) and 2 seconds have passed another runningRolloutScheduler() will be started -> meaning again systemManagement.forEachTenant -> rolloutTaskExecutor.submit() will be performed

But if we have only 2 threads, it is unknown which Tenant will actually be submitted - meaning that if we have

Tenant1, Tenant2, Tenant3, Tenant4, Tenant5, Tenant6, Tenant7 etc...

it could be that on first run all Tenant1-Tenant4 are already processed
on the second run again Tenant1 will be tried to be processed and submit will block

Once Tenant4 is fully processed and a Thread is free for work, it could be again Tenant1 to be processed (rather then Tenant5 from the first run) ?

denislavprinov: Actually if on the first run tenant 1 and tenant 4 are being processed, the next submission (lets say tenant3) will block and the scheduler will no exit thus will not tick again in the next 2 seconds - The same behaviour we have currently where the main thread processes the tenants. 

The 3rd (and subsequent) tenant tasks will block the calling thread (the scheduler method) until at least one of the first two tasks completes. The entire @Scheduled method remains busy until it finishes submitting all tenants’ tasks and only then is ready to tick again.

vasilchev: aah, if indeed @Scheduled is waiting for previous scheduled task to finish before new execution, then we are ok

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/RepositoryApplicationConfiguration.java
```diff
@@ -264,6 +266,20 @@ public Object postProcessAfterInitialization(@NonNull final Object bean, @NonNul
         };
     }
 
+    @Bean
+    public ThreadPoolTaskExecutor rolloutTaskExecutor(
```
avgustinmm, line 270: It's a common utility _ThreadPoolTaskExecutor_ 'interface' but too specific - intended to be used just for rollout scheduler. 
It could happen that, registering like that, it could be injected in other beans and used.

Better create this instance into RolloutsScheduler and doesn't expose it as a bean

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/rollout/RolloutScheduler.java
```diff
@@ -44,12 +48,29 @@ public RolloutScheduler(
...
-            // check for each tenant separately.
-            systemManagement.forEachTenant(tenant -> rolloutHandler.handleAll());
+            // workaround eclipselink that is currently not possible to
+            // execute a query without multi-tenancy if MultiTenant
+            // annotation is used.
+            // https://bugs.eclipse.org/bugs/show_bug.cgi?id=355458. So
+            // iterate through all tenants and execute the rollout check for
+            // each tenant seperately.
+
+            systemManagement.forEachTenant(tenant ->
```
avgustinmm, line 61: why not forEachTenant?
```java
systemManagement.forEachTenant(tenant -> rolloutTaskExecutor.submit(() -> { ... }));
```

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/rollout/RolloutScheduler.java
```diff
@@ -44,12 +48,29 @@ public RolloutScheduler(
...
+                rolloutTaskExecutor.submit(() ->
+                    systemSecurityContext.runAsSystemAsTenant(() -> {
+                        try {
+                            log.trace("Handling rollout for tenant: {}", tenant);
```
avgustinmm, line 65: handleAll actually logs db errors and caches (most of) the exceptions ..
but this logging could handle additionally locking exceptions.
However, for consistency - maybe it's good to put all the logging?
Not sure, though. It could be seen as different scheduler handling ...
Just for consideration

avgustinmm (changes requested):

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/rollout/RolloutScheduler.java
denislavprinov: It is being handled in a different thread, thus the tenant context should be passed again

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/rollout/RolloutScheduler.java
```diff
@@ -43,14 +48,56 @@ public RolloutScheduler(
...
+            handleParallel(tenant);
+        }
+    }
+
+    private void handleParallel(String tenant) {
+        rolloutTaskExecutor.submit(() -> systemSecurityContext.runAsSystemAsTenant(() -> {
+            try {
+                rolloutHandler.handleAll();
+            } catch (Exception e) {
+                log.error("Error processing rollout for tenant {}", tenant, e);
```
avgustinmm, line 80: this catch shall also be present in non-parallel case?
maybe it will be better to have handle method
```java
log.trace("Handling rollout for tenant: {}", tenant);
try {
    rolloutHandler.handleAll();
} catch (Exception e) {
    log.error("Error processing rollout for tenant {}", tenant, e);
}
```
which shall be called directly or in "rolloutTaskExecutor.submit(() -> systemSecurityContext.runAsSystemAsTenant()"

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/rollout/BlockWhenFullPolicy.java
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+package org.eclipse.hawkbit.repository.jpa.rollout;
+
+import java.util.concurrent.RejectedExecutionException;
+import java.util.concurrent.RejectedExecutionHandler;
+import java.util.concurrent.ThreadPoolExecutor;
+
+public class BlockWhenFullPolicy implements RejectedExecutionHandler {
+  @Override
```
avgustinmm, line 18: 4 spaces indents, please use hawkbit formatter

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/rollout/RolloutScheduler.java
```diff
@@ -28,12 +29,16 @@ public class RolloutScheduler {
...
+    private final ThreadPoolTaskExecutor rolloutTaskExecutor;
 
     public RolloutScheduler(
-            final RolloutHandler rolloutHandler, final SystemManagement systemManagement, final SystemSecurityContext systemSecurityContext) {
+        final RolloutHandler rolloutHandler, final SystemManagement systemManagement, final SystemSecurityContext systemSecurityContext,
+        final int threadPoolSize) {
         this.systemManagement = systemManagement;
         this.rolloutHandler = rolloutHandler;
         this.systemSecurityContext = systemSecurityContext;
+        this.rolloutTaskExecutor = threadPoolTaskExecutor(threadPoolSize);
```
avgustinmm, line 40: this. is not needed

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/rollout/RolloutScheduler.java
```diff
@@ -43,14 +48,58 @@ public RolloutScheduler(
...
+                    handleAll(tenant);
+                } else {
+                    handleAllAsync(tenant);
+                }
+            });
             return null;
         });
     }
+
+    private void handleAll(String tenant) {
```
avgustinmm, line 72: make it final

avgustinmm (approved):

PR #2271, opened 2025-02-12
Mgmt/actions confirm

vasilchev:
MGMT API For Actions confirm/deny when Confirmation Flow is Enabled.

Until now only Controllers (Devices) had option to confirm/deny given Action that is in status WAIT_FOR_CONFIRMATION.

Via MGMT API we could only activate/deactivate autoConfirm all Actions on given Target, but could not confirm/deny single Action.
Having that - the only option to confirm/deny singleAction is through DDI, but that overlaps with device specific calls - e.g. polling (overriding last poll query).

With this change we could separate calls triggered by Operators from End Devices (e.g. Operators to be able to use only MGMT API in order to confirm/deny given Actions)

### DIFF
--- a/hawkbit-mgmt/hawkbit-mgmt-resource/src/main/java/org/eclipse/hawkbit/mgmt/rest/resource/MgmtTargetResource.java
+++ b/hawkbit-mgmt/hawkbit-mgmt-resource/src/main/java/org/eclipse/hawkbit/mgmt/rest/resource/MgmtTargetResource.java
@@ -15,6 +15,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Optional;
 import java.util.Set;
 import java.util.function.Function;
 import java.util.stream.Collectors;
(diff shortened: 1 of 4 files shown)


hawkbit-mgmt/hawkbit-mgmt-resource/src/main/java/org/eclipse/hawkbit/mgmt/rest/resource/MgmtTargetResource.java
```diff
@@ -276,6 +284,46 @@ public ResponseEntity<MgmtAction> updateAction(final String targetId, final Long
...
+    @Override
+    public ResponseEntity<Void> updateActionConfirmation(final String targetId, final Long actionId,
+            final MgmtActionConfirmationRequestBodyPut actionConfirmation) {
+        log.debug("updateActionConfirmation with data [targetId={}, actionId={}]: {}", targetId, actionId, actionConfirmation);
+
+        return getValidatedAction(targetId, actionId).map(action -> {
+            try {
+                switch (actionConfirmation.getConfirmation()) {
+                    case CONFIRMED:
+                        log.info("Confirmed the action (actionId: {}, targetId: {}) as we got {} report",
```
avgustinmm, line 296: better be debug, could be many

avgustinmm (approved):

PR #2285, opened 2025-02-17
Remove anonymous controller support

avgustinmm:
It's not usable feature, and is error prone - someone could left anonymous enabled by mistake


maass-hamburg:
@avgustinmm this removed `hawkbit.server.ddi.security.authentication.anonymous.enabled` option right?

PR #2296, opened 2025-02-21
Remove unnecessary JsonProperty annotations

avgustinmm:
1. Removed @JsonProperty - no value
2. Mix mode with @JsonProperty and no annotation is fixed - only if required = true
3. DDI and AMQP messages are made immutable (via JsonCreator and final fields
4. All fields renamed to match the json properties - so to be consistent
5. getXXXId of the mgmt entities are renamed to getId, getId of the MgmtBaseEntity - removed - self link shall be still available


denislavprinov (approved):
LGTM

PR #2304, opened 2025-02-26
simple-ui: Also show target attributes in Target's view

andr2000:
It is useful to see target attributes in the Target's view: for instance SWUpdate can deliver values from the platform which identify some features of it. This can include an arbitrary data, e.g. platform HW revision, component's versions etc.
Thus, add a possibility to see those attributes in the dedicated view.

### DIFF
--- a/hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/view/TargetView.java
+++ b/hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/view/TargetView.java
@@ -14,6 +14,7 @@ import java.util.Date;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.stream.Stream;
@@ -51,6 +52,7 @@ import org.eclipse.hawkbit.ui.simple.view.util.SelectionGrid;
 import org.eclipse.hawkbit.ui.simple.view.util.TableView;
 import org.eclipse.hawkbit.ui.simple.view.util.Utils;
 import org.springframework.util.ObjectUtils;
+import java.util.stream.Collectors;
 
 @PageTitle("Targets")
 @Route(value = "targets", layout = MainLayout.class)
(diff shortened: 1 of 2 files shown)


hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/view/TargetView.java
```diff
@@ -239,6 +243,13 @@ private void setItem(final MgmtTarget target) {
             lastModifiedBy.setValue(target.getLastModifiedBy());
             lastModifiedAt.setValue(new Date(target.getLastModifiedAt()).toString());
             securityToken.setValue(target.getSecurityToken());
+            var response = hawkbitClient.getTargetRestApi().getAttributes(target.getControllerId());
+            if (response.getStatusCode().is2xxSuccessful()) {
+                targetAttributes.setValue(response.getBody().entrySet().stream().map(entry -> entry.getKey()+": "+
```
avgustinmm, line 248: If your target doesn't have any attributes the body is null and throws exception at this line:
```
java.lang.NullPointerException: Cannot invoke "org.eclipse.hawkbit.mgmt.json.model.target.MgmtTargetAttributes.entrySet()" because the return value of "org.springframework.http.ResponseEntity.getBody()" is null
	at org.eclipse.hawkbit.ui.simple.view.TargetView$TargetDetails.setItem(TargetView.java:248) ~[classes/:na]
```
please handle response.getBody() if null

avgustinmm (changes requested):
Hi, @andr2000,
thanks for your contribution! Attributes view is actually good feature for the Simple UI.
Please fix the NPE when no attributes and improve the code style.

hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/view/TargetView.java
andr2000: handled

hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/view/TargetView.java
```diff
@@ -239,6 +243,13 @@ private void setItem(final MgmtTarget target) {
...
+                entry.getValue()).collect(Collectors.joining("\n")));
+            }else{
```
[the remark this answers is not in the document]
andr2000: applied

andr2000:
@avgustinmm thank you for reviewing
I have updated the patch with all issues resolved (hope so)

avgustinmm (approved):
LGTM
Thanks for your contribution

PR #2327, opened 2025-03-24
Fix AMQP retries when attribute characters are invalid

avgustinmm:


hawkbit-dmf/hawkbit-dmf-amqp/src/main/java/org/eclipse/hawkbit/amqp/AmqpConfiguration.java
```diff
@@ -303,4 +289,41 @@ private static Map<String, Object> getTTLMaxArgsAuthenticationQueue() {
...
+            this.fatalByDefault = fatalByDefault;
+            this.excludeSqlStates = excludeSqlStates;
+            this.messagePatterns = messagePatterns;
+        }
+
+        @Override
+        public boolean isFatal(final Throwable t) {
+            if (t instanceof SQLException sqlException) {
+                if (excludeErrorCodes.contains(sqlException.getErrorCode())) {
+                    return !fatalByDefault;
```
vasilchev, line 314: it is a bit hard to read.
what does fatalByDefault should mean? and exclude* and messagePatterns configurations?

is it 
- fatalByDefault=true (then all should go as fatal errors and ignore exclude*)
- fatalByDefault=false (then all should go as fatal excluding the exclude* - meaning exclude are only taken into account whenever this is disabled)

I would rather use fatalByDefault field only in conditions and return explicitly the true/false uppon the conditions - e.g. return always false by default but if condition matched only then return true) or vise-versa 

Currently if excludeErrorCodes match some sql exception and fatalByDefault=false, this will in the end return isFatal true, that seems wrong

hawkbit-dmf/hawkbit-dmf-amqp/src/main/java/org/eclipse/hawkbit/amqp/AmqpProperties.java
```diff
@@ -13,15 +13,13 @@
...
- * Bean which holds the necessary properties for configuring the AMQP
- * connection.
+ * Bean which holds the necessary properties for configuring the AMQP connection.
  */
 @Data
 @ConfigurationProperties("hawkbit.dmf.rabbitmq")
 public class AmqpProperties {
 
     private static final int DEFAULT_QUEUE_DECLARATION_RETRIES = 50;
-    private static final long DEFAULT_REQUEUE_DELAY = 0;
```
vasilchev, line 24: how do we "control" the re-queue delay now and do we actually need delay at all (as it is now implemented)

hawkbit-dmf/hawkbit-dmf-amqp/src/main/java/org/eclipse/hawkbit/amqp/AmqpConfiguration.java
avgustinmm: The idea "all SQLException-s ARE/ARE NOT (fatalByDefault) fatal EXCEPT matching excluding".
That means:
* if I want to have SQLException in general fatal:
** fatalByDefault = true, excluding only some 
* if I want to have SQLException in general NOT fatal:
** fatalByDefault = false, excluding only some
i.e. the fatalByDefault is a MODE. I'll add some comments on that.

hawkbit-dmf/hawkbit-dmf-amqp/src/main/java/org/eclipse/hawkbit/amqp/AmqpProperties.java
avgustinmm: Do we really need delay? How this delay would work? It actually blocks a thread which is anyway something strange.
Since not really clear why and how this works - I'd propose to remove this support for now. And add it only if we decide why and how to do that - later.

hawkbit-dmf/hawkbit-dmf-amqp/src/main/java/org/eclipse/hawkbit/amqp/AmqpConfiguration.java
vasilchev: aah i see, so this means changing the contents of excludeError codes but also controlling the mode to actually mark it as fatal or not?


codeErrorType: enum fatal/non-fatal
codeErrorMatcher: [code1,code2]
defaultErrorType: enum fatal/non-fatal

result of isFinal returns defaultErrorPolicy,
if (codeMatcher.contains(sqlException.getErrorCode()) {
return codeErrorType;
}
return defaultErrorType;

it adds a new property and enum instead of the boolean, but is more intuitive for reading and adjusting later

vasilchev (approved):

PR #2337, opened 2025-04-01
Add CORS support for DDI API

avgustinmm:
For instance if used in remote swagger or web apps

### DIFF
--- a/hawkbit-ddi/hawkbit-ddi-starter/src/main/java/org/eclipse/hawkbit/autoconfigure/ddi/ControllerSecurityConfiguration.java
+++ b/hawkbit-ddi/hawkbit-ddi-starter/src/main/java/org/eclipse/hawkbit/autoconfigure/ddi/ControllerSecurityConfiguration.java
@@ -27,6 +27,7 @@ import org.eclipse.hawkbit.security.controller.SecurityHeaderAuthenticator;
 import org.eclipse.hawkbit.security.controller.SecurityTokenAuthenticator;
 import org.eclipse.hawkbit.tenancy.TenantAware;
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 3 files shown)


vasilchev (approved):

hawkbit-ddi/hawkbit-ddi-starter/src/main/java/org/eclipse/hawkbit/autoconfigure/ddi/ControllerSecurityConfiguration.java
```diff
@@ -88,17 +89,16 @@ protected FilterRegistrationBean<DosFilter> dosFilterDDI(final HawkbitSecurityPr
         return filterRegBean;
     }
 
+
     @Bean
     @Order(301)
-    protected SecurityFilterChain filterChainDDI(final HttpSecurity http) throws Exception {
+    protected SecurityFilterChain filterChainDDI(
+            final HttpSecurity http,
+            @Value("${hawkbit.server.security.cors.disableForDdiApi:false}") final boolean disableCorsForDdiApi) throws Exception {
```
vasilchev, line 97: prefer less inversions and simpler variables like 
hawkbit.server.security.cors.ddi.enabled=true/false

avgustinmm: this value is more like "something just in case, not expected to use", so complex seems fine + is more intuitive. What means "cors ddi enabled"? 
Regarding inversion - yes, it's more complex but from other point of view my expectation is that its almost always enabled - so, to be not enabled - "disable it" .. that is not state - it's action.
Leave it like now for the moment. Could be changed after.

PR #2349, opened 2025-04-09
README: add note to set JAVA_HOME

maass-hamburg:
add note to set JAVA_HOME before exetuting maven.


avgustinmm:
The project uses java and maven and their setup (as setting JAVA_HOME) is, IMO, out of the scope of documentation.
Maybe it just be clearly said that before building you shall have java (17+) and maven installed, though, for anyone who is up to build hawkbit I suppose it should be already clear.

PR #2365, opened 2025-04-17
Fix/db refactoring rename

avgustinmm:


strailov (approved):
Code part looks good. 
Played locally with h2 and mysql seems to work fine.
In short - LGTM.

strailov (approved):

PR #2370, opened 2025-04-24
Remove flywat from DDI & DMF servers

avgustinmm:


hawkbit-ddi/hawkbit-ddi-server/pom.xml
```diff
@@ -51,24 +51,26 @@
...
+
         <!-- h2 flyway is supported by the flyway-core dependency -->
         <dependency>
             <groupId>org.flywaydb</groupId>
             <artifactId>flyway-database-postgresql</artifactId>
+            <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.flywaydb</groupId>
             <artifactId>flyway-mysql</artifactId>
```
vasilchev, line 70: flyway-mysql and flyway-database-postgresql actually remain in ddi server ? but removed from dmf server ?

avgustinmm: just as test artifacts - not packed into the binary

PR #2375, opened 2025-04-25
Remove h2 legacy mode

avgustinmm:


avgustinmm:
Still not fixed in EclipseLink

PR #2386, opened 2025-04-30
OIDC Authentication/Authorization on Management API

ChristianBreitwieser-blue-zone:
As discussed in issue #2332 this pull request enables the possibility to login via a jwt token that can be verified by an oidc identity provider.  Hawkbit would act as an oidc resourceserver in this case.

* Added new configuration file [OidcProperties.java](https://github.com/blue-zone-at/hawkbit/blob/ad1d62a58df5e95d30ebbe672aadc31c944f1bae/hawkbit-mgmt/hawkbit-mgmt-starter/src/main/java/org/eclipse/hawkbit/oidc/OidcProperties.java), if there is a better location for this please tell me - i've not found specific configuration properties for the mgmt part.
  * "hawkbit.server.security.oauth2.resourceserver.default-enabled" which defaults to false. If this is false the current behaviour of integrating the hawkbitOAuth2ResourceServerCustomizer works as it does now. If this is true a new default oauth2 resource server configuration is loaded.
  * "hawkbit.server.security.oauth2.resourceserver.jwt.claim.roles" which defaults to "roles". This defines the claim within the jwt token that supplies the hawkbit permissions. For example when using keycloak this can be "resource_access.client-id.roles".
  * "hawkbit.server.security.oauth2.resourceserver.jwt.claim.username" which defaults to "preferred_username". This defines the claim within the jwt token that supplies the hawkbit username.
  * "hawkbit.server.security.oauth2.resourceserver.jwt.claim.tenant" which defaults to null. This defines the claim within the jwt token that supplies the hawkbit tenant. If it is default null the DEFAULT tenant is used for every user.
* Use the existing spring configuration for the resource server as described here: [Spring Documentation](https://docs.spring.io/spring-security/reference/servlet/oauth2/resource-server/jwt.html#oauth2resourceserver-jwt-authorization)
* Add a new Customizer class inside MgmtSecurityConfiguration.java which implements this default OAuth2ResourceServerCustomizer.

PR #2393, opened 2025-05-08
Fix followPathInJwtClaims and slight OIDC refactoring

avgustinmm:


ChristianBreitwieser-blue-zone:
@avgustinmm 
Tested the PR as discussed in #2386  - Login via OIDC works, the cases where claims are incorrectly configured do now also behave well.
LGTM

strailov (approved):

PR #2400, opened 2025-05-14
SimpleUI: add oauth2 support

zeapo:
With these changes one can now authenticated on the simple-ui using oauth2.

The rights checks have been mutualized with the basic auth mechanism. 

For example, using azure, one simply has to add the following env environment:
```
HAWKBIT_SERVER_SECURITY_OAUTH2_CLIENT_ENABLED=true

SPRING_SECURITY_OAUTH2_CLIENT_REGISTRATION_AZURE_CLIENT_NAME=Login With Azure
SPRING_SECURITY_OAUTH2_CLIENT_REGISTRATION_AZURE_PROVIDER=azure
SPRING_SECURITY_OAUTH2_CLIENT_REGISTRATION_AZURE_CLIENT_ID=....
SPRING_SECURITY_OAUTH2_CLIENT_REGISTRATION_AZURE_CLIENT_SECRET=...
SPRING_SECURITY_OAUTH2_CLIENT_REGISTRATION_AZURE_AUTHORIZATION_GRANT_TYPE=...
SPRING_SECURITY_OAUTH2_CLIENT_REGISTRATION_AZURE_REDIRECT_URI=...
SPRING_SECURITY_OAUTH2_CLIENT_REGISTRATION_AZURE_SCOPE=openid,profile,email
SPRING_SECURITY_OAUTH2_CLIENT_PROVIDER_AZURE_ISSUER_URI=....
SPRING_SECURITY_OAUTH2_CLIENT_PROVIDER_AZURE_JWK_SET_URI=....
```

# NOTE
~~An important drawback of the way the interceptor is made is that when the token expires there is no simple way to refresh it. As this is used too early and the user is not fully authenticated. Trying to rely on OAuth2AuthorizedClientManager leads nowhere as it wont find the client, hence will be unable to authenticate the request.~~

~~I'll probably open a future PR to update the way we obtain the authorities; this PR at least allows the user to work for 1h :)~~

Edit: it's fixed and is able to refresh the id token now

### DIFF
--- a/hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/security/OAuth2TokenManager.java
+++ b/hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/security/OAuth2TokenManager.java
@@ -1,0 +1,72 @@
+/**
+ * Copyright (c) 2025 Contributors to the Eclipse Foundation
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.hawkbit.ui.simple.security;
+
+import java.util.Optional;
+
+import org.springframework.security.oauth2.client.OAuth2AuthorizeRequest;
+import org.springframework.security.oauth2.client.OAuth2AuthorizedClient;
+import org.springframework.security.oauth2.client.OAuth2AuthorizedClientManager;
+import org.springframework.security.oauth2.client.OAuth2AuthorizedClientService;
+import org.springframework.security.oauth2.client.authentication.OAuth2AuthenticationToken;
+import org.springframework.security.oauth2.client.endpoint.OAuth2RefreshTokenGrantRequest;
+import org.springframework.security.oauth2.client.endpoint.RestClientRefreshTokenTokenResponseClient;
+import org.springframework.security.oauth2.client.registration.ClientRegistration;
+import org.springframework.security.oauth2.core.OAuth2AccessToken;
+import org.springframework.security.oauth2.core.OAuth2RefreshToken;
+import org.springframework.security.oauth2.core.endpoint.OAuth2AccessTokenResponse;
+import org.springframework.security.oauth2.core.oidc.user.DefaultOidcUser;
+import org.springframework.stereotype.Component;
+
+@Component
+public class OAuth2TokenManager {
+    private final OAuth2AuthorizedClientService clientService;
+    private final OAuth2AuthorizedClientManager clientManager;
+
+    OAuth2TokenManager(final OAuth2AuthorizedClientService clientService, final OAuth2AuthorizedClientManager clientManager) {
+        this.clientService = clientService;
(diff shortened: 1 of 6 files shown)


zeapo:
Switching to draft as I found a working solution that allows a refresh

hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/security/SecurityConfiguration.java
```diff
@@ -11,17 +11,26 @@
...
 import org.springframework.security.config.annotation.web.configuration.EnableWebSecurity;
+import org.springframework.security.config.annotation.web.configurers.oauth2.client.OAuth2LoginConfigurer;
 import org.springframework.security.crypto.bcrypt.BCryptPasswordEncoder;
 import org.springframework.security.crypto.password.PasswordEncoder;
 import org.springframework.security.web.util.matcher.AntPathRequestMatcher;
 
 @EnableWebSecurity
 @Configuration
 public class SecurityConfiguration extends VaadinWebSecurity {
+    private Customizer<OAuth2LoginConfigurer<HttpSecurity>> oAuth2LoginConfigurerCustomizer;
```
avgustinmm, line 28: Please use a empty line after the class declaration and the fields/methods as in other places of in hawkBit. Have you applied the hawkbit code format?

hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/security/OAuth2TokenManager.java
```diff
@@ -0,0 +1,72 @@
...
+        return Optional.ofNullable(authorizedToken(authentication)).orElse(
+                ((DefaultOidcUser) authentication.getPrincipal()).getIdToken().getTokenValue()
+        );
+    }
+
+    /**
+     * Tries to refresh the id token if it is expired and adds it to the request.
+     */
+    private String authorizedToken(OAuth2AuthenticationToken authentication) {
+        String registrationId = authentication.getAuthorizedClientRegistrationId();
```
strailov, line 48: Make variables final whenever possible.

hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/security/SecurityConfiguration.java
zeapo: I just noticed that it was not applied. sorry.

zeapo:
I added the same mechanims that was added to mgmt server;  the boolean `hawkbit.server.security.oauth2.client.enabled` makes it simpler to enable / disable rather than checking for the client registery

added some null checks; tested with above boolean enabled / disabled; with registration configured / non configured. 
Behavior is:
- if enabled and no config > spring will complain that the registration is missing
- if disabled > with and without config is the same, will use basic auth

zeapo:
Please let me know if any further changes are required. Knowing that once Spring Security 6.5 is out and we bump version, I'll revisit this to remove the manual token refresh.

strailov:
I'll try to check this out one more time today and if everything is fine I think we are ready to merge it.

strailov:
Looks good, thanks @zeapo !

PR #2401, opened 2025-05-14
SimpleUI: improve dialogs and add delete confirmation dialog

zeapo:
- Add a dialog to confirm delete before any remove action
dialog_delete_confirmation
- Improve the way the dialogs are displayed by resizing them and moving the action handlers to the footer
  Before
  dialog_layout_before
  After
  dialog_layout_after


strailov (approved):
Hello @zeapo ! 
Thanks for your contribution! Looking good to me.

PR #2404, opened 2025-05-15
SimpleUI: Add Poll Status and Update Status + improve target filtering

zeapo:
Before and after the poll status / update status. I used the same ideas from the previous ui, one minor difference is when no poll status is available I used a "?" rather than "!" which I reserved for overdue
add_status_column

Some changes to the Target Filter:
- Used the correct icons: loop for search, refresh for reset + better alignment
- Fixed discrepancy in the padding between the filters
- Added filter update mechanism and made it clear when we're saving vs duplicating a filter

output

### DIFF
--- a/hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/view/TargetView.java
+++ b/hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/view/TargetView.java
@@ -9,43 +9,7 @@
  */
 package org.eclipse.hawkbit.ui.simple.view;
 
-import java.time.ZoneOffset;
-import java.util.Collections;
-import java.util.Date;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Map;
-import java.util.Objects;
-import java.util.Optional;
-import java.util.Set;
-import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.CompletionStage;
-import java.util.function.Function;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
-
-import jakarta.annotation.security.RolesAllowed;
-
-import org.eclipse.hawkbit.mgmt.json.model.MgmtPollStatus;
-import org.eclipse.hawkbit.mgmt.json.model.distributionset.MgmtActionType;
-import org.eclipse.hawkbit.mgmt.json.model.distributionset.MgmtDistributionSet;
-import org.eclipse.hawkbit.mgmt.json.model.distributionset.MgmtTargetAssignmentRequestBody;
-import org.eclipse.hawkbit.mgmt.json.model.tag.MgmtTag;
-import org.eclipse.hawkbit.mgmt.json.model.target.MgmtTarget;
-import org.eclipse.hawkbit.mgmt.json.model.target.MgmtTargetRequestBody;
-import org.eclipse.hawkbit.mgmt.json.model.targetfilter.MgmtTargetFilterQuery;
-import org.eclipse.hawkbit.mgmt.json.model.targetfilter.MgmtTargetFilterQueryRequestBody;
-import org.eclipse.hawkbit.mgmt.json.model.targettype.MgmtTargetType;
-import org.eclipse.hawkbit.ui.simple.HawkbitMgmtClient;
-import org.eclipse.hawkbit.ui.simple.MainLayout;
-import org.eclipse.hawkbit.ui.simple.view.util.Filter;
-import org.eclipse.hawkbit.ui.simple.view.util.SelectionGrid;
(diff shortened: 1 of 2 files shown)


hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/view/TargetView.java
```diff
@@ -200,21 +211,66 @@ private RawFilter(final HawkbitMgmtClient hawkbitClient) {
...
                         add(name);
                         open();
                     }});
-            saveBtn.addClickShortcut(Key.ENTER);
+
+            updateBtn.addClickListener(e ->
+                    new Utils.BaseDialog<Void>("Update Filter") {{
+                        final MgmtTargetFilterQuery selected = savedFilters.getValue();
+                        if (selected == null) {
+                            close();
```
avgustinmm, line 226: shouldn't here you have "return"?
and then remove "assert selected != null"

strailov: Isn't this anonymous inner class initializer ? I believe you cannot invoke return statements there. 
If we do not want assert to be used, maybe the approach should be changed ?

hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/view/TargetView.java
```diff
@@ -200,21 +211,66 @@ private RawFilter(final HawkbitMgmtClient hawkbitClient) {
...
             textSaveLayout.setAlignItems(FlexComponent.Alignment.BASELINE);
             textSaveLayout.setWidthFull();
             layout.add(savedFilters, textSaveLayout);
+            layout.addClassNames(LumoUtility.Gap.SMALL);
+        }
+
+        private static List<MgmtTargetFilterQuery> listFilters(HawkbitMgmtClient hawkbitClient) {
+            return Optional.ofNullable(
+                    hawkbitClient.getTargetFilterQueryRestApi()
+                            .getFilters(0, 30, null, null, null)
```
strailov, line 271: Would't this limit target filters size list up to 30 ? 
Basically UI will list the first 30 target filters from hawkbit and nothing more ?

hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/view/TargetView.java
zeapo: yes, return is not usable; I'll see if there is a nicer approach

hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/view/TargetView.java
zeapo: You're right, just noticed that, I moved the line above here :D
we're missing a way to paginate over the filters in a select; I'll check if there are any other locations with the same issue and see if there is a simple way to fix that.

hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/view/TargetView.java
```diff
@@ -286,7 +342,8 @@ private RegisterDialog(final HawkbitMgmtClient hawkbitClient) {
             final Button register = Utils.tooltip(new Button("Register"), "Register (Enter)");
             type = new Select<>(
                     "Type",
-                    e -> {},
+                    e -> {
+                    },
                     hawkbitClient.getTargetTypeRestApi()
                             .getTargetTypes(0, 30, Constants.NAME_ASC, null)
```
zeapo, line 348: @strailov see, same here too. I guess this will need a separate PR to handle thses cases

hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/view/TargetView.java
strailov: Ahh, yes - seems initially it was this way...

hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/view/TargetView.java
zeapo: fixed by extracting the nullability check; and cleaned up by extracting the event listeners; 
the selected should never be null, but in case we mistakingly allow the button to be clicked it would not open a dialog

strailov (approved):
Looks good. Thanks @zeapo

PR #2409, opened 2025-05-19
SimpleUI: support target's tags edition, display installed & assigned distribution sets, and action history

zeapo:
Contains a lot of UI improvement and a fix to a flaky behavior;

* After multiple deployements to our test environment, we noticed that sometimes the OAuth2TokenManager is not initalized all the time; by making it conditioned on the same property as the customizer, it's always there and working fine.

* Details view transformed into tabs 
target_tab_details

* Display the installed and assigned distribution sets
target_tab_assigned

* Add tags listing; assigning, unassigning, and creation
target_tag_1
target_tag_2
target_tag_3

I didn't add tag deletion of tags, it will need a new modal to list & manage existing tags.

* Add Action History view
Screenshot 2025-05-21 at 13 42 44


zeapo:
switching to draft as I'm adding Actions history tab; this way this PR will contain all intended features rather than part of it :)

zeapo:
@strailov dunno if you want me to split this into two PR, one to fix the flaky OAuth2TokenManager and the rest as UI features?

maass-hamburg:
@zeapo great work you are doing with the SimpleUI lately.

strailov:
Hello @zeapo. Sorry I missed this. Will check this PR asap :)

hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/view/TargetView.java
```diff
@@ -92,13 +108,14 @@ protected void addColumns(final Grid<MgmtTarget> grid) {
                         grid.addColumn(MgmtTarget::getTargetTypeName).setHeader(Constants.TYPE).setAutoWidth(true);
 
                         grid.setItemDetailsRenderer(new ComponentRenderer<>(
-                                () -> new TargetDetails(hawkbitClient), TargetDetails::setItem));
+                                () -> new TargetDetailedView(hawkbitClient), TargetDetailedView::setItem));
```
strailov, line 111: Personally I would separate this from the grid totally. 
For example I would use the on click event and show the TargetDetailedView under the whole grid. 
However using the detailed grid seems quite efficient here and I am also fine with it.

hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/view/TargetView.java
```diff
@@ -335,14 +389,365 @@ private void setItem(final MgmtTarget target) {
...
+            private Boolean isCancelingOrCanceled() {
+                return action.getType().equals(MgmtAction.ACTION_CANCEL);
+            }
+
+            public Icon getActiveIcon() {
+                Icon icon;
+                if (isActive()) {
+                    icon = new SpinnerIcon();
+                } else {
+                    icon = VaadinIcon.CHECK_CIRCLE.create();
```
strailov: Would this also mean that if the action is in ERROR state still the CHECK_CIRCLE will be present? It is quite confusing this with the Status column in the grid. I would unify this Active column with the Status in the target grid maybe, what do you think ? Like when it is pending there is a yellow half full circle

zeapo:
Thanks, just le me know if you need further improvements. I'll be working on another subject and be back to this once Spring Boot 3.5 is merged.

hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/view/TargetView.java
```diff
@@ -335,14 +389,365 @@ private void setItem(final MgmtTarget target) {
...
+        private void setItem(final MgmtTarget target) {
+            this.target = target;
+        }
+
+        @Override
+        protected void onAttach(AttachEvent attachEvent) {
+            Optional.ofNullable(hawkbitClient.getTargetRestApi().getInstalledDistributionSet(target.getControllerId()))
+                    .map(ResponseEntity<MgmtDistributionSet>::getBody)
+                    .ifPresent(value -> {
+                        installed.setValue("Name : " + value.getName() + "\n" +
```
strailov: Maybe we should use System.lineSeparator() here to be compatible with all operating systems. And I would define it as constant in the whole class.

hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/view/TargetView.java
```diff
@@ -335,14 +389,365 @@ private void setItem(final MgmtTarget target) {
...
+                                "Version : " + value.getVersion() + "\n" +
+                                value.getModules().stream().map(module -> module.getTypeName() + ": " + module.getVersion())
+                                        .collect(Collectors.joining("\n")));
+                    });
+            Optional.ofNullable(hawkbitClient.getTargetRestApi().getAssignedDistributionSet(target.getControllerId()))
+                    .map(ResponseEntity<MgmtDistributionSet>::getBody)
+                    .ifPresent(value -> {
+                        assigned.setValue("Name : " + value.getName() + "\n" +
+                                "Version : " + value.getVersion() + "\n" +
```
strailov: Line separator here also

hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/view/TargetView.java
zeapo: That would work, I didn't want to break this behavior right now. The current way of showing the details makes the UI quite jumpy; having a split layout that would display at the bottom the details would work great.

hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/view/TargetView.java
zeapo: Absolutely, I struggled to get a clear separation of the status / type concepts; 
I wanted to keep a sort of similarity with the previous UI 
image

I'll try to merge them and see how it ends up 👍🏼

hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/view/TargetView.java
zeapo: updated both and merged them as they use the same code;
I switched to text blocks + a replace all as suggested by the section `Normalization Of Line Terminators` in the official doc

zeapo:
FYI, https://github.com/eclipse-hawkbit/hawkbit/pull/2409/commits/a4e4ea93fcae73b44dd906c1376375b64010f432 
I had to create a custom theme, as since I started working on this PR, something weird happened and the badges stopped working with new builds. Seems like we've got to explicitly import the badge into the theme, hence the custom theme :)

hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/view/TargetView.java
zeapo: Just updated it; here are the four cases, half_circle + orange is pending as target status; half_circle + red is canceling ; the others are the same
Screenshot 2025-06-03 at 19 15 42
Screenshot 2025-06-03 at 19 17 13
Screenshot 2025-06-03 at 19 16 50
Screenshot 2025-06-03 at 19 16 20

zeapo: PS: also extracted the force quit so that it's cleaner as it was confusing close to the other two actions

strailov (approved):
Looks good! Thanks @zeapo !

PR #2428, opened 2025-06-04
simple-ui: add support for side panel vs grid details

zeapo:
As discussed in previous PR; this allows to show the details in a side panel. After some testing, having the details on the a side panel on the right is a way nicer experience than having it at the bottom (the close / open button is close to the panel).

Behaviour:
- When an "eye" button is clicked we rely on a provided function to show the details; the button is then converted into a cross
- Clicking on another "eye' button, will replace the content without closing
- The cross closes the panel as expected

Bonus, fix a performance issue introduced in previous PR when showing details, the tags were fetched too early which slowed down the details display.

Screenshot 2025-06-04 at 12 50 22
Screenshot 2025-06-04 at 12 50 53


strailov (approved):
Looks good! Thanks @zeapo !

PR #2434, opened 2025-06-06
Fix dynamic rollouts when there are finished actions from previous ro…

avgustinmm:
…llouts


hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/specifications/TargetSpecifications.java
```diff
@@ -471,44 +427,39 @@ public static Specification<JpaTarget> hasTargetTypeNot(final Long typeId) {
...
-//                                    cb.gt(actionsJoin.get(JpaAction_.weight), weight))));
+                                    cb.and(
+                                            cb.lt(actionsJoin.get(JpaAction_.rollout).get(AbstractJpaBaseEntity_.id), rolloutId),
+                                            cb.gt(actionsJoin.get(JpaAction_.weight), weight)),
+                                    cb.or(
+                                            cb.equal(actionsJoin.get(JpaAction_.active), true),
+                                            cb.equal(actionsJoin.get(JpaAction_.status), Action.Status.SCHEDULED))),
+                            cb.and(
+                                    cb.ge(actionsJoin.get(JpaAction_.rollout).get(AbstractJpaBaseEntity_.id), rolloutId),
+                                    cb.ge(actionsJoin.get(JpaAction_.weight), weight))));
```
vasilchev, line 462: isn't there a case where these targets never execute its action if for some reason newer rollout is created with greater weight-> this way newer rollout will override and execute.
but what happens after this newer rollout finishes its action? 
This rollout (intermediate with less weight) will never be executed ?

PR #2435, opened 2025-06-06
Fix retrieve Rollout with details

vasilchev:


hawkbit-repository/hawkbit-repository-api/src/main/java/org/eclipse/hawkbit/repository/RolloutManagement.java
```diff
@@ -412,6 +412,6 @@ Slice<Rollout> findByFiltersWithDetailedStatus(@NotNull Pageable pageable, @NotE
      *
      * @param rollouts the rollouts to be enriched.
      */
-    @PreAuthorize(SpringEvalExpressions.HAS_AUTH_ROLLOUT_MANAGEMENT_UPDATE)
+    @PreAuthorize(SpringEvalExpressions.HAS_AUTH_ROLLOUT_MANAGEMENT_READ)
```
avgustinmm, line 415: Better rename the method to make it obvious that it doesn't change the rollouts into DB - i.e. it's not an update method - but adds info to the in memory objects - i.e. it's an read method.
For instance append/add/get/readRolloutStatusDetails. It would be also good to return the "enriched" slice

avgustinmm (changes requested):

PR #2449, opened 2025-06-12
Fix / improves management of the Rollout's total target counter

avgustinmm:


hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaRolloutExecutor.java
```diff
@@ -625,20 +596,18 @@ private JpaRolloutGroup fillRolloutGroupWithTargets(
...
                     targetsLeftToAdd -= assigned;
                 }
-            } while (targetsLeftToAdd > 0);
+            }
 
             group.setStatus(RolloutGroupStatus.READY);
-            group.setTotalTargets(DeploymentHelper.runInNewTransaction(txManager,
-                    "countRolloutTargetGroupByRolloutGroup",
-                            count -> rolloutTargetGroupRepository.countByRolloutGroup(group)).intValue());
+            group.setTotalTargets((int) (expectedInGroup - targetsLeftToAdd));
```
vasilchev, line 602: isnt targetsLeftToAdd always 0 in the end ? 

anyways, this expectedInGroup is calculated with percentage of all groups by the amount of targets that currently match the filter.. What happens if devices disapear and we have already processed some groups with higher amount of devices (because the total amount was bigger then)

maybe a corner case... but previous impl kind-of sets a number that is really assigned, here it seems we work with relative information that could change from execution to execution? or i am missing something ?

avgustinmm: yes, it's not expected to be < 0, but for sure we could ensure correctness even in case < 0 - e.g. bug, or the requested page is a return bigger.
expectedInGroup was calculated this way before. We could not do much here - well the count will differ - why not? 
and for real use-cases hundred / thousands the diff will be insignificant

vasilchev (approved):

PR #2503, opened 2025-06-27
Support cache eviction in microservice mode

blechschmidt-ldc:
When running in microservice mode, updating configuration properties via the management component does not automatically update caches in other services.

This change introduces support for distributed cache eviction using Spring Cloud Bus. It listens for remote tenant configuration events and evicts the relevant cache key upon receipt.

### DIFF
--- a/hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaTenantConfigurationManagement.java
+++ b/hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaTenantConfigurationManagement.java
@@ -28,12 +28,15 @@ import java.util.stream.Collectors;
 import lombok.extern.slf4j.Slf4j;
 import org.eclipse.hawkbit.im.authentication.SpPermission;
 import org.eclipse.hawkbit.repository.TenantConfigurationManagement;
+import org.eclipse.hawkbit.repository.event.remote.entity.RemoteEntityEvent;
 import org.eclipse.hawkbit.repository.exception.InsufficientPermissionException;
+import org.eclipse.hawkbit.repository.exception.InvalidTenantConfigurationKeyException;
 import org.eclipse.hawkbit.repository.exception.TenantConfigurationValidatorException;
 import org.eclipse.hawkbit.repository.exception.TenantConfigurationValueChangeNotAllowedException;
 import org.eclipse.hawkbit.repository.jpa.configuration.Constants;
 import org.eclipse.hawkbit.repository.jpa.executor.AfterTransactionCommitExecutor;
 import org.eclipse.hawkbit.repository.jpa.model.JpaTenantConfiguration;
+import org.eclipse.hawkbit.repository.jpa.model.helper.TenantAwareHolder;
 import org.eclipse.hawkbit.repository.jpa.repository.TenantConfigurationRepository;
 import org.eclipse.hawkbit.repository.model.PollStatus;
 import org.eclipse.hawkbit.repository.model.Target;
@@ -49,6 +52,7 @@ import org.springframework.cache.CacheManager;
 import org.springframework.cache.annotation.CacheEvict;
 import org.springframework.cache.annotation.Cacheable;
 import org.springframework.context.ApplicationContext;
+import org.springframework.context.ApplicationListener;
 import org.springframework.core.convert.support.ConfigurableConversionService;
 import org.springframework.core.convert.support.DefaultConversionService;
 import org.springframework.dao.ConcurrencyFailureException;
@@ -63,7 +67,8 @@ import org.springframework.validation.annotation.Validated;
 @Slf4j
 @Transactional(readOnly = true)
 @Validated
-public class JpaTenantConfigurationManagement implements TenantConfigurationManagement {
+public class JpaTenantConfigurationManagement
+        implements TenantConfigurationManagement, ApplicationListener<RemoteEntityEvent<TenantConfiguration>> {
 
     private static final ConfigurableConversionService CONVERSION_SERVICE = new DefaultConversionService();
 
@@ -324,5 +329,30 @@ public class JpaTenantConfigurationManagement implements TenantConfigurationMana
                 throw new TenantConfigurationValueChangeNotAllowedException();
             }
         }
+    }
+
+    /**
+     * Ensures that cache eviction takes place in microservice mode.
+     *
+     * @param event The event indicating that a configuration value has been updated.
+     * @throws InvalidTenantConfigurationKeyException if the configuration key cannot be found.
+     */
+    @Override
+    public void onApplicationEvent(RemoteEntityEvent<TenantConfiguration> event) {
+        if (!TenantAwareHolder.getInstance().getTenantAware().getCurrentTenant().equals(event.getTenant())) {
+            return;
+        }
+
+        final Long entityId = event.getEntityId();
+
+        tenantConfigurationRepository.findById(entityId).ifPresentOrElse(tenantConfiguration -> {
+            final Cache cache = cacheManager.getCache("tenantConfiguration");
+            if (cache != null) {
+                cache.evict(tenantConfiguration.getKey());
+            }
+        }, () -> {
+            throw new InvalidTenantConfigurationKeyException(String.format(
+                    "Cannot find tenant configuration key with ID %d.", entityId));
+        });
     }
 }


hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaTenantConfigurationManagement.java
```diff
@@ -63,7 +67,8 @@
 @Slf4j
 @Transactional(readOnly = true)
 @Validated
-public class JpaTenantConfigurationManagement implements TenantConfigurationManagement {
+public class JpaTenantConfigurationManagement
+        implements TenantConfigurationManagement, ApplicationListener<RemoteEntityEvent<TenantConfiguration>> {
```
avgustinmm: Maybe it's better to use event listener approach used in other places in hawkBit. E.g. `@EventListener(classes = RolloutDeletedEvent.class)` in RolloutStatusCache.
Also, maybe the eviction events of interests are not all TenantConfiguration events but only TenantUpdatedEvent?

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaTenantConfigurationManagement.java
```diff
@@ -325,4 +330,29 @@ private void assertBatchAssignmentValueChange(final String key, final JpaTenantC
...
+
+    /**
+     * Ensures that cache eviction takes place in microservice mode.
+     *
+     * @param event The event indicating that a configuration value has been updated.
+     * @throws InvalidTenantConfigurationKeyException if the configuration key cannot be found.
+     */
+    @Override
+    public void onApplicationEvent(RemoteEntityEvent<TenantConfiguration> event) {
+        if (!TenantAwareHolder.getInstance().getTenantAware().getCurrentTenant().equals(event.getTenant())) {
```
avgustinmm: Not 100% sure but I suppose this listener will be called without `current tenant` - i.e. the code will always return here?

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaTenantConfigurationManagement.java
```diff
@@ -325,4 +330,29 @@ private void assertBatchAssignmentValueChange(final String key, final JpaTenantC
...
+            return;
+        }
+
+        final Long entityId = event.getEntityId();
+
+        tenantConfigurationRepository.findById(entityId).ifPresentOrElse(tenantConfiguration -> {
```
avgustinmm: why call to db to check if there is a tenant? you could just evict the cache entry if exists.
So, just check if there is a cache entry and evict key (if doesn't exist - fine).
Also, what's the benefit of throwing an exception if not such tenant? Who will react on that?

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaTenantConfigurationManagement.java
avgustinmm: I see that there is TenantAwareApplicationEventPublisher that maybe will run the event into the tenant's contexts. But then there is not need to check it. So, I propose to skip that check. Do you think there is a reason to keep it?

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaTenantConfigurationManagement.java
blechschmidt-ldc: 227a928284efe7d43dfe9639e5df665d3b405a24 now makes use of `@EventListener`. Being quite new to the Spring Framework, I was under the wrong impression that this would not work.

In my opinion, it makes sense to keep `RemoteEntityEvent<TenantConfiguration>` to clear the cache on any change. `TenantConfigurationCreatedEvent` and `TenantConfigurationUpdatedEvent` are the classes extending `RemoteEntityEvent<TenantConfiguration>`. Reacting to the `TenantConfigurationCreatedEvent` might be useful in rare edge cases, for example in case the management server crashes after the database transaction without having dispatched the AMQP message. In any case, the cache should be empty then. Since the `TenantConfigurationDeletedEvent` does not extend `RemoteEntityEvent<TenantConfiguration>`, I have added a second listener.

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaTenantConfigurationManagement.java
blechschmidt-ldc: My initial understanding was that upon receiving a message, `TenantAwareHolder.getInstance().getTenantAware().getCurrentTenant()` would always return `DEFAULT` (in single-tenant mode). Having performed further tests, this is not the case. Replacing the `tenant` property in the AMQP message would cause it to return the value specified in the message. The idea was to filter out message not destined for the tenant which the `JpaTenantConfigurationManagement` class is responsible for in case multiple tenants use the same AMQP exchange. To be able to handle that 227a928284efe7d43dfe9639e5df665d3b405a24 now follows the pattern that was implemented here:
https://github.com/eclipse-hawkbit/hawkbit/blob/4a6e862d57d18c73d56dcc529e346b8658cdd642/hawkbit-ddi/hawkbit-ddi-resource/src/main/java/org/eclipse/hawkbit/ddi/rest/resource/DdiRootController.java#L138-L141

So in the multi-tenancy case a tenant-aware service matcher could be implemented.

However, that code was removed just two days ago in https://github.com/eclipse-hawkbit/hawkbit/commit/bdbc658ea9fb96ce9fe43bb6ca7dd44285740cbf#diff-437dcd9de50357ffbc373f806229a3a2a5ac0e25f42284f3d683fc64371497cdL138-L142. Alternatively, an approach could consist in extending the `tenantConfigurationManagement` method with a `@Nullable ServiceMatcher` here:

https://github.com/eclipse-hawkbit/hawkbit/blob/baab05f00962520326b8f820cef64d7c8a28c137/hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/RepositoryApplicationConfiguration.java#L602-L611

Do you think that would be the best approach?

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaTenantConfigurationManagement.java
blechschmidt-ldc: I had some possible invariants in mind (e.g. if an update event comes in, the configuration key should exist) that should result in user attention (e.g. through an Exception being logged). But you are right, they are probably not worth handling (since anyhow, the invariant is not necessarily true in case of an update and and deletion immediately being performed afterwards) and the cache is not really that critical.

blechschmidt-ldc:
Hi @avgustinmm, thanks a lot for your review. 227a928284efe7d43dfe9639e5df665d3b405a24 makes some changes to the original PR.

Another change I have not yet discussed is using `evictIfPresent` instead of using `evict`. IMO, there is no reason to use the latter, since `evictIfPresent` guarantees immediate eviction, according to the method's documentation.

It would be great if you could review those and provide feedback again.

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaTenantConfigurationManagement.java
avgustinmm: JpaTenantConfigurationManagement is not per tenant - it is a single instance/service. Every call to it, however, works in a tenant's context. As you have found out, in case of remote event, it is set to the context. 
So, the context you have is the one for the tenant. So you have which tenant is related here - if you need.
So, you have a system event where the system says - it's tenant X. When the system (other host) receive this event - it dispatch it into the tenant X scope - so it's correct tenant there. Then you clean cache for the key.
I don't see need to check tenant. 
Service matchers are more for testing if the event comes from this host or is a remote one

blechschmidt-ldc: Thanks a lot for the explanation. In the current state of the PR, the service matcher is indeed used to perform that kind of testing (local or remote event). Then for me, the only question left is whether the public setter for the service matcher would really be needed or whether it would make more sense to just have a `private @Autowired(required = false)` matcher. What do you think?

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaTenantConfigurationManagement.java
```diff
@@ -325,4 +338,45 @@ private void assertBatchAssignmentValueChange(final String key, final JpaTenantC
...
+
+    private void evictCacheEntryByKeyIfPresent(final String key) {
+        final Cache cache = cacheManager.getCache("tenantConfiguration");
+        if (cache != null) {
+            cache.evictIfPresent(key);
+        }
+    }
+
+    private boolean shouldProcessRemoteTenantAwareEvent(final RemoteTenantAwareEvent event) {
+        return serviceMatcher == null || !serviceMatcher.isFromSelf(event) && serviceMatcher.isForSelf(event);
```
avgustinmm, line 394: I'm not exactly sure how this isFrom/ForSelf works - would it be event true?
Why don't just clear cache unconditionally? or only if !isFromSelf (if the cache is cleaned locally anyway)?

blechschmidt-ldc: A transmitted event looks like this:

```json
{
  "type": "TenantConfigurationUpdatedEvent",
  "timestamp": 1752069948597,
  "originService": "mgmt-server:8080:9e9fbdb6e2582b8e1f1a87076940a291",
  "destinationService": "**",
  "id": "6f50cb6e-dfb4-4cdb-80ea-79e2fc6d2e0f",
  "tenant": "DEFAULT",
  "entityId": 1,
  "entityClass": "org.eclipse.hawkbit.repository.jpa.model.JpaTenantConfiguration",
  "interfaceClass": "org.eclipse.hawkbit.repository.model.TenantConfiguration"
}
```

By default, Spring Cloud Bus injects a `PathServiceMatcher`:

```
PathServiceMatcher@59
  configNames: String[0]@86
  id: "ddi-server:8081:154568bed1264effa90b151449aa2823"
  matcher: DefaultBusPathMatcher@88
    delagateMatcher: AntPathMatcher@174
      cachePatterns: null
      caseSensitive: true
      pathSeparator: ":"
      pathSeparatorPatternCache: AntPathMatcher$PathSeparatorPatternCache@177
      stringMatcherCache: ConcurrentHashMap@178 size=0
      tokenizedPatternCache: ConcurrentHashMap@179 size=2
      trimTokens: false
```


The `isForSelf` method is implemented here: https://github.com/spring-cloud/spring-cloud-bus/blob/2fa5448dab950f278ac7e73e698608d1a55e0d6e/spring-cloud-bus/src/main/java/org/springframework/cloud/bus/PathServiceMatcher.java#L58-L73). The referenced `PathMatcher matcher` is a `DefaultBusPathMatcher` (cf. its `match` method at https://github.com/spring-cloud/spring-cloud-bus/blob/2fa5448dab950f278ac7e73e698608d1a55e0d6e/spring-cloud-bus/src/main/java/org/springframework/cloud/bus/DefaultBusPathMatcher.java#L110) and the `delagateMatcher` (sic!) of that matcher in turn is an `AntPathMatcher`. According to the documentation of that class, `** matches zero or more directories in a path`: https://github.com/spring-projects/spring-framework/blob/47d9182f56bb512c7b1081c81a4188a6a52aa8dc/spring-core/src/main/java/org/springframework/util/AntPathMatcher.java#L39

As a result, `**` indicates that the message is a broadcast message. Given that a message contains a `destinationService` field, I think it makes sense to respect it. However, if you do not share that opinion, I am happy to remove that condition, which I have locally tested to be true.

avgustinmm: Ok, then isForSelf seems fine. isFromSelf will skip processing event on the same host - I suppose you have tested that the local cache is cleaned by other flows?
Regarding @Autowired - I'd prefer the way you have implemented it - in order to kept it hidden.
Please move the private methods at the end (in general in hawkbit we follow public -> private methods order), and I think the PR will be ready to be merged.
Thanks

blechschmidt-ldc: > I suppose you have tested that the local cache is cleaned by other flows?

I could indeed confirm that the local cache was cleaned through the `@CacheEvict` annotations.

https://github.com/eclipse-hawkbit/hawkbit/blob/21581c4ea4bfec27b0e9370ea9dc2ae6b950d981/hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaTenantConfigurationManagement.java#L94-L101

https://github.com/eclipse-hawkbit/hawkbit/blob/21581c4ea4bfec27b0e9370ea9dc2ae6b950d981/hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaTenantConfigurationManagement.java#L119-L126

These will call the `evict` method of the cache. (In practice, this is https://github.com/spring-projects/spring-framework/blob/440bf94cee10c00077eb81162e38b6a9443d97a0/spring-context-support/src/main/java/org/springframework/cache/caffeine/CaffeineCache.java#L180.)

I have rebased my PR. 562bb7929206216d393e7fcfb980e002fa7c1165 also changes the order of methods.

Thanks a lot.

avgustinmm (approved):
LGTM. Thanks for the contribution!

PR #2534, opened 2025-07-04
simple-ui: improve the oidc id token refresh

zeapo:
After the upgrade to Spring Boot 3.5.x, the authorized client manager will attempt a refresh of the oidc id token.

Sadly, I thought that it would be a simple clean replacement of the previous hackish code. However, it looks like vaadin does not like the SpringContextHolder.setContext() hacks (if you guys have a solution I'm all ears). Here's what happens when the token expires:

- We get to the interceptor, it calls getToken, which detecting that the token expired will call a refresh 
- During the refresh, spring security will try to reload the oidc user, we end up in the OidcUserService bean which is going to try to refresh the user
- In that exact bean, sadly, we tried to fetch the rights of the user , which in turn will call the interceptor, and we loop again!

Unfortunately, i'm unable to make the SpringContextHolder clear and use the temporary context in this situation, regardless of changing the strategy, the interceptor is always using the previous context.

To remediate this, I went with a bandaid, I fetch the previous user and re-use its grants; this works! But it is not the ideal solution :) 

Opening this in Draft, as we'll be testing this during the next few days with our QA.

### DIFF
--- a/hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/SimpleUIApp.java
+++ b/hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/SimpleUIApp.java
@@ -12,7 +12,6 @@ package org.eclipse.hawkbit.ui.simple;
 import com.vaadin.flow.component.page.AppShellConfigurator;
 import com.vaadin.flow.server.PWA;
 import com.vaadin.flow.theme.Theme;
-import com.vaadin.flow.theme.lumo.Lumo;
 import feign.Contract;
 import feign.RequestInterceptor;
 import feign.codec.Decoder;
@@ -42,6 +41,7 @@ import org.springframework.security.oauth2.client.userinfo.OAuth2UserService;
 import org.springframework.security.oauth2.core.oidc.user.DefaultOidcUser;
 import org.springframework.security.oauth2.core.oidc.user.OidcUser;
 
+import java.util.ArrayList;
 import java.util.Base64;
 import java.util.LinkedList;
 import java.util.List;
@@ -59,9 +59,9 @@ public class SimpleUIApp implements AppShellConfigurator {
 
     private static final Function<OAuth2TokenManager, RequestInterceptor> AUTHORIZATION = (oAuth2TokenManager) -> requestTemplate -> {
         final Authentication authentication = SecurityContextHolder.getContext().getAuthentication();
-        if (oAuth2TokenManager != null && authentication instanceof OAuth2AuthenticationToken oAuth2AuthenticationToken) {
-            String bearerToken = oAuth2TokenManager.getToken(oAuth2AuthenticationToken);
-            requestTemplate.header("Authorization", "Bearer " + bearerToken);
+        if (authentication.getPrincipal() instanceof OidcUser) {
+            var token = oAuth2TokenManager.getToken((OAuth2AuthenticationToken) authentication);
+            requestTemplate.header("Authorization", "Bearer " + token);
         } else {
             requestTemplate.header(
                     "Authorization", "Basic " + Base64.getEncoder().encodeToString(
@@ -110,15 +110,29 @@ public class SimpleUIApp implements AppShellConfigurator {
     OAuth2UserService<OidcUserRequest, OidcUser> oidcUserService(final HawkbitMgmtClient hawkbitClient) {
         final OidcUserService delegate = new OidcUserService();
         return (userRequest) -> {
-            OidcUser oidcUser = delegate.loadUser(userRequest);
+            // fetch the previous oidc user to re-use the same granted authorities and avoid too many requests
+            // when the token refreshes (and potential looping, as the refresh will most certainly go through the oidc user service)
+            Authentication authentication = SecurityContextHolder.getContext().getAuthentication();
+            OidcUser existingUser = authentication != null && authentication.getPrincipal() instanceof OidcUser existing
+                    ? existing
+                    : null;
 
-            final OAuth2AuthenticationToken tempToken = new OAuth2AuthenticationToken(
-                    oidcUser,
-                    emptyList(),
-                    userRequest.getClientRegistration().getRegistrationId()
-            );
-            final List<SimpleGrantedAuthority> grantedAuthorities =
-                    getGrantedAuthorities(hawkbitClient, tempToken);
+            OidcUser oidcUser = delegate.loadUser(userRequest);
+            final List<SimpleGrantedAuthority> grantedAuthorities = new ArrayList<>();
+            if (existingUser == null) {
+                final OAuth2AuthenticationToken tempToken = new OAuth2AuthenticationToken(
+                        oidcUser,
+                        emptyList(),
+                        userRequest.getClientRegistration().getRegistrationId()
+                );
+                grantedAuthorities.addAll(getGrantedAuthorities(hawkbitClient, tempToken));
+            } else {
+                grantedAuthorities.addAll(existingUser.getAuthorities()
+                        .stream()
+                        .map(authority -> new SimpleGrantedAuthority(authority.getAuthority()))
+                        .toList()
+                );
+            }
             return new DefaultOidcUser(
                     grantedAuthorities,
                     oidcUser.getIdToken(),
--- a/hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/security/OAuth2TokenManager.java
+++ b/hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/security/OAuth2TokenManager.java
@@ -15,17 +15,10 @@ import org.springframework.security.oauth2.client.OAuth2AuthorizedClient;
 import org.springframework.security.oauth2.client.OAuth2AuthorizedClientManager;
 import org.springframework.security.oauth2.client.OAuth2AuthorizedClientService;
 import org.springframework.security.oauth2.client.authentication.OAuth2AuthenticationToken;
-import org.springframework.security.oauth2.client.endpoint.OAuth2AccessTokenResponseClient;
-import org.springframework.security.oauth2.client.endpoint.OAuth2RefreshTokenGrantRequest;
-import org.springframework.security.oauth2.client.endpoint.RestClientRefreshTokenTokenResponseClient;
-import org.springframework.security.oauth2.client.registration.ClientRegistration;
-import org.springframework.security.oauth2.core.OAuth2AccessToken;
-import org.springframework.security.oauth2.core.OAuth2RefreshToken;
-import org.springframework.security.oauth2.core.endpoint.OAuth2AccessTokenResponse;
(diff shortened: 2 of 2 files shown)


avgustinmm:
@zeapo ,
is this PR finalized? Working, for review and for eventual merge?

zeapo:
@avgustinmm hey, it's done. I've been testing it for a while on our deployment.

 I'm off for the next two weeks, won't be able to test changes on my work environment.

zeapo:
Please let me know if anything is missing :)

hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/security/OAuth2TokenManager.java
```diff
@@ -10,73 +10,48 @@
...
-        // A small trick to refresh the token if it is expired; the current spring version does not refresh the ID Token when the Access Token is refreshed
-        // This won't be necessary after Spring Security 6.5; cf. https://github.com/spring-projects/spring-security/pull/16589
-        OAuth2AccessToken accessToken = refreshClient.getAccessToken();
-        OAuth2RefreshToken refreshToken = refreshClient.getRefreshToken();
-        ClientRegistration clientRegistration = refreshClient.getClientRegistration();
-        // if this is null, please request it via the scopes
-        if (refreshToken == null) return null;
+        OAuth2AuthorizeRequest request = OAuth2AuthorizeRequest.withClientRegistrationId(registrationId).principal(authentication).build();
+        // since Spring Security 6.5 this will trigger a refresh of the id token
+        authorizedClient = clientManager.authorize(request);
```
avgustinmm, line 50: Does it refresh token only if expired / expiring or always?

hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/SimpleUIApp.java
```diff
@@ -59,9 +59,9 @@ public class SimpleUIApp implements AppShellConfigurator {
 
     private static final Function<OAuth2TokenManager, RequestInterceptor> AUTHORIZATION = (oAuth2TokenManager) -> requestTemplate -> {
         final Authentication authentication = SecurityContextHolder.getContext().getAuthentication();
-        if (oAuth2TokenManager != null && authentication instanceof OAuth2AuthenticationToken oAuth2AuthenticationToken) {
-            String bearerToken = oAuth2TokenManager.getToken(oAuth2AuthenticationToken);
-            requestTemplate.header("Authorization", "Bearer " + bearerToken);
+        if (authentication.getPrincipal() instanceof OidcUser) {
+            var token = oAuth2TokenManager.getToken((OAuth2AuthenticationToken) authentication);
```
avgustinmm: are we sure the auth is OAuth2AuthenticationToken?
shouldn't we check if "authentication instanceof OAuth2AuthenticationToken"?
why check "authentication.getPrincipal() instanceof OidcUser" at all?

hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/SimpleUIApp.java
```diff
@@ -110,15 +110,29 @@ HawkbitMgmtClient hawkbitMgmtClient(final Tenant tenant, final HawkbitClient haw
...
+            final List<SimpleGrantedAuthority> grantedAuthorities = new ArrayList<>();
+            if (existingUser == null) {
+                final OAuth2AuthenticationToken tempToken = new OAuth2AuthenticationToken(
+                        oidcUser,
+                        emptyList(),
+                        userRequest.getClientRegistration().getRegistrationId()
+                );
+                grantedAuthorities.addAll(getGrantedAuthorities(hawkbitClient, tempToken));
+            } else {
+                grantedAuthorities.addAll(existingUser.getAuthorities()
```
avgustinmm: where the existing user comes from? Are we sure existing user is the oidcUser? should we check they are the same before granting new user the existingUser's permissions?

hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/security/OAuth2TokenManager.java
zeapo: Only if it expired, or expiring given the skew; 
> Attempt to authorize or re-authorize (if required) the client identified by the provided clientRegistrationId.

previous version, before 6.5, didn't refresh the id token when needed

hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/SimpleUIApp.java
zeapo: unless I'm mistaken; if the principal is an OidcUser, the authentication is an OAuth2AuthenticationToken;

we need an OidcUser to have an idToken; without the idToken we can't interact with the MgmtService; as it needs the id token to fill in the user's information

avgustinmm: anyway, wince you just need the case to OAuth2AuthenticationToken IMO it's better to do something like:
```java
        if (authentication instanceof OAuth2AuthenticationToken oauth2AuthToken) {
            var token = oAuth2TokenManager.getToken(oauth2AuthToken);
```
this will ensure not class cast exception will occur

hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/SimpleUIApp.java
zeapo: As we're in the oidc user service (OAuth2UserService<OidcUserRequest, OidcUser>), we're sure it's an OidcUser

the existingUser, is the OidcUser in the current SecurityContext; 
the SecurityContext relies on the currently connected user, either using the session or, in this case, the token thye provide during their browsing

hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/SimpleUIApp.java
zeapo: wouldn't that remove the guarantee that the principal is an OidcUser? cause the OAuth2AuthenticationToken can contain a simple DefaultOAuth2User (https://docs.spring.io/spring-security/site/docs/current/api/org/springframework/security/oauth2/core/user/OAuth2User.html) which wont contain the id token, 

regardless, just updated the code :)

hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/SimpleUIApp.java
avgustinmm: I meant are we sure that oidcUser == existingUser. I mean:
* oidsUser is the user we are authorizing (it has not yet any granted authorities).
* existingUser is the context user/entity
 
how do we know "who is" existingUser and if it is the same user (oidsUser)? Could it be a leftover from another session? Where it comes from. If it is a user with all necessary properties - why we do here any authorization?

I'm not deeply in that and the flow is a little bit unclear here.

zeapo: First Login:
* when it's the first time the user is authenticated, the exisstingUser is null, as the context does not contain the current user
* in this case we have to hijack the OidcUser service to be able to grant him authorities (they have to go through feing to find the rights)
* once we grant him the authorities we save the context and we move on

On token refresh:
* we'll end up in the OidcUser Service as this is the spring security flow for it to provide a new oidc user 
* However, as at this location, the token in the authentication is not up to date (it will be after this method is done); trying to reload the granted authorities by calling feing will try to use the old token; hence try to refresh the token again, hence calling the same method again, and hence an infinite loop
* Therefore, to avoid this, we fetch the granted authorities from the previously saved context

Now, how can we be sure that the existing user is the newly authenticated user; we're sure cause' we rely on the token the user is providing us through the calls they're making to the server. It can't be someone else, unless they stole their token (which is a different story)

I hope this answers your questions.

desislava-marinova: This code will be used not only when you need a refresh token, but also when you do several logins.
The `existing` user and the `oidcUser` can be different if you log in one tab with user1 and in another tab with user2. In this case, user2 will be assigned the permissions of user1 in the UI view. This will lead to errors if the permissions are not the same.

zeapo: would comparing the subjects be good for you here? Indeed, I didn't take into account someone using multiple accounts on same browser.

desislava-marinova: In most cases, comparing the subjects will work. It will not work only if the granted permissions for the same user are different between the two logins (different consent is given), but that should be a rare case.

zeapo: great news, since the upgrade to latest version of spring boot, which happened after this PR was created, the infinite loop was fixed.
no need to hack things together, now it gets the granted authorities every-time, which solves both issues

hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/security/OAuth2TokenManager.java
```diff
@@ -10,73 +10,48 @@
...
-     */
-    private String authorizedToken(final OAuth2AuthenticationToken authentication) {
+        final String currentToken = ((DefaultOidcUser) authentication.getPrincipal()).getIdToken().getTokenValue();
         String registrationId = authentication.getAuthorizedClientRegistrationId();
-        OAuth2AuthorizeRequest request = OAuth2AuthorizeRequest.withClientRegistrationId(registrationId).principal(authentication).build();
 
         // This ensures that there is a client already, otherwise we won't be able to call the manager for authorization
         OAuth2AuthorizedClient authorizedClient = clientService.loadAuthorizedClient(registrationId, authentication.getName());
-        if (authorizedClient == null) return null;
+        if (authorizedClient == null) return currentToken;
```
flobz, line 44: If client isn't found why return a token ?

zeapo: Hehe, this is the tricky parts of how simple-ui rights are working. It hijacks the authentication process, hence the initial step where it fetches the rights to create the authorities, the authorizedClient does not exist. 

That's why this OIDC process is much more complex than it should've been.

image
Screenshot 2025-09-01 at 14 31 10

desislava-marinova (approved):
LGTM

PR #2535, opened 2025-07-07
Fine-grained permissions

avgustinmm:
Adds support for permissions of type <permission>(/<rsql filter scope>)


hawkbit-ddi/hawkbit-ddi-resource/src/test/java/org/eclipse/hawkbit/ddi/rest/resource/DdiRootControllerTest.java
```diff
@@ -12,6 +12,7 @@
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.eclipse.hawkbit.im.authentication.SpPermission.SpringEvalExpressions.CONTROLLER_ROLE_ANONYMOUS;
 import static org.eclipse.hawkbit.im.authentication.SpPermission.TENANT_CONFIGURATION;
+import static org.eclipse.hawkbit.repository.test.util.SecurityContextSwitch.*;
```
strailov: Don't get use of wildcard imports

hawkbit-monolith/hawkbit-update-server/src/test/java/org/eclipse/hawkbit/app/PreAuthorizeEnabledTest.java
```diff
@@ -47,6 +47,26 @@ void successIfHasRole() throws Exception {
...
+        mvc.perform(get("/rest/v1/distributionsets")).andExpect(result ->
+                assertThat(result.getResponse().getStatus()).isEqualTo(HttpStatus.OK.value()));
+    }
+
+    /**
+     * Tests whether request succeed if a role with scope is granted for the user
+     */
+    @Test
+    @WithUser(authorities = { SpPermission.READ_REPOSITORY + "/type==2" }, autoCreateTenant = false)
+    void failIfHasNoForbiddingScope() throws Exception {
```
strailov, line 69: didn't quite get the intention here 
test name indicates that something should fail, however we expect HttpStatus.ok ?

hawkbit-repository/hawkbit-repository-jpa/src/test/java/org/eclipse/hawkbit/repository/jpa/acm/controller/DistributionSetAccessControllerTest.java
```diff
@@ -1,327 +0,0 @@
-/**
```
strailov, line 1: Why do we remove the old Access Control tests, they should be still valid, yes ?

hawkbit-repository/hawkbit-repository-core/src/main/java/org/eclipse/hawkbit/repository/RepositoryDefaultConfiguration.java
```diff
@@ -23,4 +41,85 @@
...
 @PropertySource("classpath:/hawkbit-repository-defaults.properties")
-public class RepositoryDefaultConfiguration {}
\ No newline at end of file
+public class RepositoryDefaultConfiguration {
+
+    @Bean
+    @Primary
+    MethodSecurityExpressionHandler methodSecurityExpressionHandler(
+            final Optional<RoleHierarchy> roleHierarchy, final Optional<ApplicationContext> applicationContext) {
+        final DefaultMethodSecurityExpressionHandler methodSecurityExpressionHandler = new DefaultMethodSecurityExpressionHandler() {
```
strailov: In SecurityAutoConfiguration there is already bean defined MethodSecurityExpressionHandler, should this one be removed since seems like the new one covers the old csase also or ?

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/acm/DefaultAccessController.java
```diff
@@ -0,0 +1,167 @@
...
+        }
+    }
+
+    @Autowired
+    void setContextAware(final ContextAware contextAware) {
+        this.contextAware = contextAware;
+    }
+
+    @Autowired(required = false)
+    void setContextAware(final RoleHierarchy roleHierarchy) {
```
strailov: maybe rename this setter to setRoleHierarchy ?

strailov (approved):

hawkbit-monolith/hawkbit-update-server/src/test/java/org/eclipse/hawkbit/app/PreAuthorizeEnabledTest.java
avgustinmm: Good catch. It doesn't test anything - will fix it

hawkbit-repository/hawkbit-repository-jpa/src/test/java/org/eclipse/hawkbit/repository/jpa/acm/controller/DistributionSetAccessControllerTest.java
avgustinmm: Yes, but why to have same tests for same things?

hawkbit-repository/hawkbit-repository-core/src/main/java/org/eclipse/hawkbit/repository/RepositoryDefaultConfiguration.java
avgustinmm: yes, you are right. Will remove from SecurityAutoConfig - only repository uses mathod security so better be there

PR #2538, opened 2025-07-08
Introduce target grouping

strailov:
In order to be provide option to group devices in different groups introducing new field/attribute to the targets (sp_target / JpaTarget). Though at the moment the targets have target type and metadata / attributes which also allows for grouping, this has drawbacks:

-  target type is related with a concrete device discrimination based on things like hardware type. It is related to the distribution sets that could be installed. Groups will provide different way of separation
- to filter against metadata / attributes - joins of target table shall be made which does thins much more heavy.

Current PR introduces target_group field in JpaTarget where : 

1. The paths support hierarchy - e.g. x/y/z, and the search should provide all devices in sub groups
2. Proper indexing so the searches work fine
3. Group is part of sp_target - so the query is fast
4. Initial Simple UI support 

Search could be made with direct equals of provided group, but also with filter based group : 
If we have groups like x/y/z we are able to search for targets matching x/y/*.

### DIFF
--- a/hawkbit-core/src/main/java/org/eclipse/hawkbit/repository/TargetFields.java
+++ b/hawkbit-core/src/main/java/org/eclipse/hawkbit/repository/TargetFields.java
@@ -29,6 +29,7 @@ public enum TargetFields implements RsqlQueryField {
     UPDATESTATUS("updateStatus"),
     IPADDRESS("address"),
     ATTRIBUTE("controllerAttributes"),
+    TARGETGROUP("targetGroup"),
     ASSIGNEDDS(
             "assignedDistributionSet",
             DistributionSetFields.NAME.getJpaEntityFieldName(), DistributionSetFields.VERSION.getJpaEntityFieldName()),
--- a/hawkbit-mgmt/hawkbit-mgmt-api/src/main/java/org/eclipse/hawkbit/mgmt/json/model/target/MgmtTarget.java
+++ b/hawkbit-mgmt/hawkbit-mgmt-api/src/main/java/org/eclipse/hawkbit/mgmt/json/model/target/MgmtTarget.java
@@ -127,4 +127,7 @@ public class MgmtTarget extends MgmtNamedEntity {
 
     @Schema(description = "Present if user consent flow active. Indicates if auto-confirm is active", example = "false")
     private Boolean autoConfirmActive;
+
+    @Schema(description = "Target group", example = "Europe/East")
+    private String group;
 }
--- a/hawkbit-mgmt/hawkbit-mgmt-api/src/main/java/org/eclipse/hawkbit/mgmt/rest/api/MgmtTargetGroupRestApi.java
+++ b/hawkbit-mgmt/hawkbit-mgmt-api/src/main/java/org/eclipse/hawkbit/mgmt/rest/api/MgmtTargetGroupRestApi.java
@@ -1,0 +1,337 @@
+/**
+ * Copyright (c) 2025 Contributors to the Eclipse Foundation
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.hawkbit.mgmt.rest.api;
+
+import io.swagger.v3.oas.annotations.Operation;
+import io.swagger.v3.oas.annotations.extensions.Extension;
+import io.swagger.v3.oas.annotations.extensions.ExtensionProperty;
+import io.swagger.v3.oas.annotations.media.Content;
+import io.swagger.v3.oas.annotations.media.Schema;
+import io.swagger.v3.oas.annotations.responses.ApiResponse;
+import io.swagger.v3.oas.annotations.responses.ApiResponses;
+import io.swagger.v3.oas.annotations.tags.Tag;
+import org.eclipse.hawkbit.mgmt.json.model.PagedList;
+import org.eclipse.hawkbit.mgmt.json.model.target.MgmtTarget;
+import org.eclipse.hawkbit.rest.OpenApi;
+import org.eclipse.hawkbit.rest.json.model.ExceptionInfo;
+import org.springframework.hateoas.MediaTypes;
+import org.springframework.http.MediaType;
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.DeleteMapping;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.PathVariable;
+import org.springframework.web.bind.annotation.PutMapping;
+import org.springframework.web.bind.annotation.RequestBody;
+import org.springframework.web.bind.annotation.RequestParam;
+
+import java.util.List;
+
+import static org.eclipse.hawkbit.mgmt.rest.api.MgmtRestConstants.TARGET_GROUP_ORDER;
+
+@Tag(
+        name = "Target Groups", description = "REST API for Target Groups operations.",
+        extensions = @Extension(name = OpenApi.X_HAWKBIT, properties = @ExtensionProperty(name = "order", value = TARGET_GROUP_ORDER)))
+public interface MgmtTargetGroupRestApi {
+
+
+    /**
+     * Handles the GET request of retrieving a list of assigned targets for a specific group. Complex grouping (subgroups) not supported here.
+     * For complex grouping use the analogical resource with query parameter for target group.
+     *
+     * @param targetGroup - target group
+     * @param pagingOffsetParam the offset of list of targets for pagination, might not be present in the rest request then default value will
+     *         be applied
+     * @param pagingLimitParam the limit of the paged request, might not be present in the rest request then default value will be applied
+     * @param sortParam the sorting parameter in the request URL, syntax {@code field:direction, field:direction}
+     * @return a list of targets matching the provided group for a defined or default page request with status OK. The response is always paged. In any failure the
+     *         JsonResponseExceptionHandler is handling the response.
+     */
+    @Operation(summary = "Return assigned targets for group",
+            description = "Handles the GET request of retrieving a list of assigned targets for a specific group. Complex grouping (subgroups) not supported here." +
+                    "For complex grouping use the analogical resource with query parameter for target group.")
+    @ApiResponses(value = {
+            @ApiResponse(responseCode = "200", description = "Successfully retrieved"),
+            @ApiResponse(responseCode = "400", description = "Bad Request - e.g. invalid parameters",
+                    content = @Content(mediaType = "application/json", schema = @Schema(implementation = ExceptionInfo.class))),
+            @ApiResponse(responseCode = "401", description = "The request requires user authentication."),
+            @ApiResponse(responseCode = "403", description = "Insufficient permissions, entity is not allowed to be " +
+                    "changed (i.e. read-only) or data volume restriction applies."),
+            @ApiResponse(responseCode = "405", description = "The http request method is not allowed on the resource."),
+            @ApiResponse(responseCode = "406", description = "In case accept header is specified and not application/json."),
+            @ApiResponse(responseCode = "429", description = "Too many requests. The server will refuse further attempts " +
+                    "and the client has to wait another second.")
+    })
+    @GetMapping(value = MgmtRestConstants.TARGET_GROUP_V1_REQUEST_MAPPING + MgmtRestConstants.TARGET_GROUP_TARGETS_REQUEST_MAPPING,
+            produces = { MediaTypes.HAL_JSON_VALUE, MediaType.APPLICATION_JSON_VALUE })
+    ResponseEntity<PagedList<MgmtTarget>> getAssignedTargets(
+            @PathVariable
+            @Schema(description = "The target group of the targets")
+            String targetGroup,
+            @RequestParam(
+                    value = MgmtRestConstants.REQUEST_PARAMETER_PAGING_OFFSET,
+                    defaultValue = MgmtRestConstants.REQUEST_PARAMETER_PAGING_DEFAULT_OFFSET)
+            @Schema(description = "The paging offset (default is 0)")
+            int pagingOffsetParam,
+            @RequestParam(
+                    value = MgmtRestConstants.REQUEST_PARAMETER_PAGING_LIMIT,
+                    defaultValue = MgmtRestConstants.REQUEST_PARAMETER_PAGING_DEFAULT_LIMIT)
+            @Schema(description = "The maximum number of entries in a page (default is 50)")
+            int pagingLimitParam,
+            @RequestParam(value = MgmtRestConstants.REQUEST_PARAMETER_SORTING, required = false)
+            @Schema(description = """
+                    The query parameter sort allows to define the sort order for the result of a query. A sort criteria
+                    consists of the name of a field and the sort direction (ASC for ascending and DESC descending).
+                    The sequence of the sort criteria (multiple can be used) defines the sort order of the entities
+                    in the result.""")
+            String sortParam
+    );
+
+    /**
+     * Handles the GET request of retrieving a list of assigned targets for a specific group. Complex grouping (subgroups) is supported here.
+     * Group filter could be proper complex group - e.g. Parent/Child or also could be filtered with wildcard for subgroups - Parent/*
+     *
+     * @param groupFilter - An Actual group or group filter - Parent/Child or Parent/*
+     * @param subgroups - If set to {@code true} enables searches with wildcard, proper equals will be used otherwise
+     * @param pagingOffsetParam the offset of list of targets for pagination, might not be present in the rest request then default value will
+     *         be applied
+     * @param pagingLimitParam the limit of the paged request, might not be present in the rest request then default value will be applied
+     * @param sortParam the sorting parameter in the request URL, syntax {@code field:direction, field:direction}
+     * @return a list of targets matching the provided group for a defined or default page request with status OK. The response is always paged. In any failure the
+     *         JsonResponseExceptionHandler is handling the response.
+     */
+    @Operation(summary = "Return assigned targets for group",
+            description = "Handles the GET request of retrieving a list of assigned targets for a specific group. Complex grouping (subgroups) is supported here." +
+                    "Group filter could be proper complex group - e.g. Parent/Child or also could be filtered with wildcard for subgroups - Parent/*")
+    @ApiResponses(value = {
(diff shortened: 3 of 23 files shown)


hawkbit-core/src/main/java/org/eclipse/hawkbit/repository/TargetFields.java
```diff
@@ -29,6 +29,7 @@ public enum TargetFields implements RsqlQueryField {
     UPDATESTATUS("updateStatus"),
     IPADDRESS("address"),
     ATTRIBUTE("controllerAttributes"),
+    TARGETGROUP("targetGroup"),
```
avgustinmm: since the column in db is anyway target_group better rename this search key to GROUP and the jpa attribute to "group" and all params and so on

hawkbit-mgmt/hawkbit-mgmt-api/src/main/java/org/eclipse/hawkbit/mgmt/json/model/target/MgmtTarget.java
```diff
@@ -127,4 +127,7 @@ public class MgmtTarget extends MgmtNamedEntity {
 
     @Schema(description = "Present if user consent flow active. Indicates if auto-confirm is active", example = "false")
     private Boolean autoConfirmActive;
+
+    @Schema(description = "Target group", example = "Europe/East")
+    private String group;
```
avgustinmm: maybe move it just after controllerId parameter

hawkbit-mgmt/hawkbit-mgmt-resource/src/main/java/org/eclipse/hawkbit/mgmt/rest/resource/MgmtTargetGroupResource.java
```diff
@@ -0,0 +1,116 @@
...
+import static org.eclipse.hawkbit.mgmt.rest.resource.util.PagingUtility.sanitizeTargetSortParam;
+
+@Slf4j
+@RestController
+public class MgmtTargetGroupResource implements MgmtTargetGroupRestApi {
+
+    private final TargetManagement targetManagement;
+    private final TenantConfigHelper tenantConfigHelper;
+
+    public MgmtTargetGroupResource(final TargetManagement targetManagement, final SystemSecurityContext systemSecurityContext,
```
avgustinmm, line 39: organize somehow methods, e.g. 
Create - assign
Read - get
Delete - unassign
for every method create / delete you should have 6 methods - single controller (optional), controller list, with rsql x with and without subgroup
for read - 2 methods with and without subgroup

hawkbit-mgmt/hawkbit-mgmt-resource/src/main/java/org/eclipse/hawkbit/mgmt/rest/resource/MgmtTargetGroupResource.java
```diff
@@ -0,0 +1,116 @@
...
+        return ResponseEntity.ok(new PagedList<>(rest, targets.getTotalElements()));
+    }
+
+    @Override
+    public ResponseEntity<PagedList<MgmtTarget>> getAssignedTargetsWithSubgroups(String groupFilter, boolean subgroups, int pagingOffsetParam, int pagingLimitParam, String sortParam) {
+        final Pageable pageable = PagingUtility.toPageable(pagingOffsetParam, pagingLimitParam, sanitizeTargetSortParam(sortParam));
+
+        Page<Target> targets;
+        if (subgroups) {
+            String groupFilterRegex = "^[^*%]*\\*?$";
```
avgustinmm: This is an SQL logic - better hide it in the JPA layer - i.e. 
* accept here simple greal group names - not a filter - e.g. x/y/z
* have a single findTargetsByGroup(group name, withSubGroups)
which will do eq or like by name

hawkbit-repository/hawkbit-repository-api/src/main/java/org/eclipse/hawkbit/repository/TargetManagement.java
```diff
@@ -601,6 +601,61 @@ boolean isTargetMatchingQueryAndDSNotAssignedAndCompatibleAndUpdatable(
...
+    List<String> findGroups();
+
+    /**
+     * Update the target group of the targets matching the provided rsql filter.
+     *
+     * @param group target group parameter
+     * @param rsql rsql filter for {@link Target}
+     */
+    @PreAuthorize(SpringEvalExpressions.HAS_AUTH_UPDATE_TARGET)
+    void updateTargetGroupsWithRsql(@NotEmpty String group, @NotNull String rsql);
```
avgustinmm: better rename to assignTargetGroup - to match "assigned" in rest api
also at first look I'm getting impression that method updates group and it will now have exactly these

hawkbit-mgmt/hawkbit-mgmt-resource/src/main/java/org/eclipse/hawkbit/mgmt/rest/resource/MgmtTargetGroupResource.java
strailov: Here for the delete methods we do not need to provide group/subgroup so I see it as 
1.delete with controllerId
2.delete with List of controllerIds
3.delete by rsql 

For the Create (currently Put mapping) we have now : 
1. PUT rest/v1/targetgroups/{group}/assigned with body list of controllerIds 
2. PUT rest/v1/targetgroups/assigned?group=<?> with body list of controllerIds - here we support group & subgroups
3. PUT rest/v1/targetgroups/{controllerId}?group=<?> - here we also support group & subgroups
4. PUT rest/v1/targetgroups?group=<group>&q=<rsqlExpression> 

As far as I understood you need to split 2 & 3 for the different cases - group and subgroup is that right? 
According to the delete methods I don't think there is need more than the above 3

avgustinmm: Summarized like that PUT rest/v1/targetgroups/{controllerId}?group=<?> looks strange to me - after targetgroups I'd expect {group}
Maybe replace it with PUT rest/v1/targetgroups/{groupId}?q=
Single controller would be then assigned with list of single element

strailov: We do not have groupId. If we pass group as a path param we won't support subgroups here. And where to pass the controllerId then ?

avgustinmm: what about:

    1. PUT rest/v1/targetgroups/{group}/assigned with body list of controllerIds
    2. PUT rest/v1/targetgroups/assigned?group=<?> with body list of controllerIds - here we support group & subgroups
    3. PUT rest/v1/targetgroups/{groupId}?q= with rsql
    4. PUT rest/v1/targetgroups?group=&q= with group (supporting subgroups) + rsql

strailov: 1,2,4 are supported in the current implementation
Will add logic for case 3 in the next commit. 
Meanwhile implemented the missing delete logic - and combined Delete cases 2 & 3 in single resource because of conflict for resource... You can check it out how it turned out

strailov: But If we have 3 we cannot have : 
PUT rest/v1/targetgroups/{controllerId}?group=x/y/z ?

avgustinmm: yes, don't want that - anyway this could be done with 1 with list of single controller id

strailov: Removed it and replaced it with PUT rest/v1/targetgroups/{groupId}?q= with rsql.

hawkbit-mgmt/hawkbit-mgmt-api/src/main/java/org/eclipse/hawkbit/mgmt/rest/api/MgmtTargetGroupRestApi.java
```diff
@@ -0,0 +1,374 @@
...
+                    content = @Content(mediaType = "application/json", schema = @Schema(implementation = ExceptionInfo.class))),
+            @ApiResponse(responseCode = "401", description = "The request requires user authentication."),
+            @ApiResponse(responseCode = "403", description = "Insufficient permissions, entity is not allowed to be " +
+                    "changed (i.e. read-only) or data volume restriction applies."),
+            @ApiResponse(responseCode = "405", description = "The http request method is not allowed on the resource."),
+            @ApiResponse(responseCode = "406", description = "In case accept header is specified and not application/json."),
+            @ApiResponse(responseCode = "429", description = "Too many requests. The server will refuse further attempts " +
+                    "and the client has to wait another second.")
+    })
+    @DeleteMapping(value = MgmtRestConstants.TARGET_GROUP_V1_REQUEST_MAPPING + "/assigned")
```
avgustinmm, line 279: when you assign you use schema:
* .../assigned - when assign List of controller ids
* ... without /assigned - when assign with RSQL
So, another option (instead of having such one or another method) by this schema when this delete method to have 
* .../assinged delete with controller ids(with group in path)
* .. delete with rsql - again with group

hawkbit-mgmt/hawkbit-mgmt-api/src/main/java/org/eclipse/hawkbit/mgmt/rest/api/MgmtTargetGroupRestApi.java
```diff
@@ -0,0 +1,374 @@
...
+            @ApiResponse(responseCode = "403", description = "Insufficient permissions, entity is not allowed to be " +
+                    "changed (i.e. read-only) or data volume restriction applies."),
+            @ApiResponse(responseCode = "404", description = "Target not found.",
+                    content = @Content(mediaType = "application/json", schema = @Schema(hidden = true))),
+            @ApiResponse(responseCode = "405", description = "The http request method is not allowed on the resource."),
+            @ApiResponse(responseCode = "406", description = "In case accept header is specified and not application/json."),
+            @ApiResponse(responseCode = "429", description = "Too many requests. The server will refuse further attempts " +
+                    "and the client has to wait another second.")
+    })
+    @DeleteMapping(value = MgmtRestConstants.TARGET_GROUP_V1_REQUEST_MAPPING + "/{controllerId}/assigned")
```
avgustinmm: when you have delete by list of controller ids (upper method) this one is not needed
And it is confusing to have controller id in path when at all other places only group is used in path
PS: and typo - unnasign

hawkbit-mgmt/hawkbit-mgmt-resource/src/main/java/org/eclipse/hawkbit/mgmt/rest/resource/MgmtTargetGroupResource.java
```diff
@@ -0,0 +1,117 @@
...
+
+    public MgmtTargetGroupResource(final TargetManagement targetManagement, final SystemSecurityContext systemSecurityContext,
+            final TenantConfigurationManagement tenantConfigurationManagement) {
+        this.targetManagement = targetManagement;
+        this.tenantConfigHelper = TenantConfigHelper.usingContext(systemSecurityContext, tenantConfigurationManagement);
+    }
+
+    @Override
+    public ResponseEntity<PagedList<MgmtTarget>> getAssignedTargets(String group, int pagingOffsetParam, int pagingLimitParam, String sortParam) {
+        final Pageable pageable = PagingUtility.toPageable(pagingOffsetParam, pagingLimitParam, sanitizeTargetSortParam(sortParam));
```
avgustinmm, line 47: same code as getAssignedTargetsWithSubgroups
just call getAssignedTargetsWithSubgroups with subgroups false
or extract method that both will call

hawkbit-mgmt/hawkbit-mgmt-api/src/main/java/org/eclipse/hawkbit/mgmt/rest/api/MgmtTargetGroupRestApi.java
```diff
@@ -0,0 +1,374 @@
...
+                    content = @Content(mediaType = "application/json", schema = @Schema(implementation = ExceptionInfo.class))),
+            @ApiResponse(responseCode = "401", description = "The request requires user authentication."),
+            @ApiResponse(responseCode = "403", description = "Insufficient permissions, entity is not allowed to be " +
+                    "changed (i.e. read-only) or data volume restriction applies."),
+            @ApiResponse(responseCode = "405", description = "The http request method is not allowed on the resource."),
+            @ApiResponse(responseCode = "406", description = "In case accept header is specified and not application/json."),
+            @ApiResponse(responseCode = "429", description = "Too many requests. The server will refuse further attempts " +
+                    "and the client has to wait another second.")
+    })
+    @GetMapping(value = MgmtRestConstants.TARGET_GROUP_V1_REQUEST_MAPPING + MgmtRestConstants.TARGET_GROUP_TARGETS_REQUEST_MAPPING,
```
avgustinmm, line 71: gouldn't be better if we have boolean subgroups param here too?

hawkbit-repository/hawkbit-repository-jpa-flyway/src/main/resources/db/migration/H2/V1_12_33__add_group_to_target__H2.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE sp_target ADD COLUMN target_group VARCHAR(256);
+CREATE INDEX sp_idx_target_group_01 ON sp_target (tenant, target_group);
```
avgustinmm: No need of _01 if you have just one target and group index

hawkbit-mgmt/hawkbit-mgmt-api/src/main/java/org/eclipse/hawkbit/mgmt/rest/api/MgmtTargetGroupRestApi.java
strailov: Here the group is as path param - how will be able to search for subgroups ?

hawkbit-repository/hawkbit-repository-jpa-flyway/src/main/resources/db/migration/H2/V1_12_33__add_group_to_target__H2.sql
strailov: What if we would like to add more indexes for target group in future ?

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/specifications/TargetSpecifications.java
```diff
@@ -192,6 +192,22 @@ public static Specification<JpaTarget> likeControllerIdOrName(final String searc
...
+            return criteriaBuilder.equal(criteriaBuilder.lower(targetRoot.get(JpaTarget_.group)), groupTextToLower);
+        };
+    }
+
+    public static Specification<JpaTarget> likeTargetGroup(final String targetGroupSearch) {
+        return (targetRoot, query, criteriaBuilder) ->  {
+            final String searchTextToLower = targetGroupSearch.toLowerCase();
+            return criteriaBuilder.or(
+                    criteriaBuilder.equal(criteriaBuilder.lower(targetRoot.get(JpaTarget_.group)), searchTextToLower),
+                    criteriaBuilder.like(criteriaBuilder.lower(targetRoot.get(JpaTarget_.group)), searchTextToLower.concat("%")));
```
avgustinmm: ! Here you should concatenate with "/%" otherwise a search for "xy" group and subgroups will give you "xyz"

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/specifications/TargetSpecifications.java
```diff
@@ -192,6 +192,22 @@ public static Specification<JpaTarget> likeControllerIdOrName(final String searc
...
     }
 
+    public  static Specification<JpaTarget> eqTargetGroup(final String targetGroup) {
+        return (targetRoot, query, criteriaBuilder) -> {
+            final String groupTextToLower = targetGroup.toLowerCase();
+            return criteriaBuilder.equal(criteriaBuilder.lower(targetRoot.get(JpaTarget_.group)), groupTextToLower);
+        };
+    }
+
+    public static Specification<JpaTarget> likeTargetGroup(final String targetGroupSearch) {
```
avgustinmm: maybe eqOrSubTargetGroup will match more what the method does

hawkbit-repository/hawkbit-repository-api/src/main/java/org/eclipse/hawkbit/repository/TargetManagement.java
```diff
@@ -601,6 +601,51 @@ boolean isTargetMatchingQueryAndDSNotAssignedAndCompatibleAndUpdatable(
...
 
+    /**
+     * Assigns the target group to a given {@link Target}.
+     *
+     * @param controllerId to be updated
+     * @param group - to be assigned to target
+     * @return updated target
+     */
+    @PreAuthorize(SpringEvalExpressions.HAS_AUTH_UPDATE_TARGET)
+    void assignTargetGroup(@NotEmpty String controllerId, String group);
```
avgustinmm: ! Not used?

avgustinmm (changes requested):

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/specifications/TargetSpecifications.java
strailov: good catch

avgustinmm (approved):

PR #2546, opened 2025-07-14
Allow to filter actions by createdAt and lastModifiedAt timestamp

strailov:


hawkbit-core/src/main/java/org/eclipse/hawkbit/repository/ActionStatusFields.java
```diff
@@ -18,7 +18,10 @@
 public enum ActionStatusFields implements RsqlQueryField {
 
     ID("id"),
-    REPORTEDAT("createdAt");
+    REPORTEDAT("createdAt"),
```
avgustinmm, line 22: In order to keep compatibility I'd propose to add 
CREATEDAT("createdAt"),
CREATEDBY("createdBy"),
Then we will be able to search her with createdat/by as the rest + reportedby/at as more meaningful for status, or (?) - if the status is NEVER modified - we could remove LASTMODIFIED (they would be meaningless anyway) and have here different fields as the entity is different

avgustinmm (changes requested):

hawkbit-core/src/main/java/org/eclipse/hawkbit/repository/ActionStatusFields.java
strailov: Most of the times status is never modified after creation, at least I could not think of such cases ... So I will directly remove it ...

PR #2554, opened 2025-07-16
Improve Simple UI

flobz:
This PR improve and fix some issue of the Simeple-UI. 

Major improvement:
- for each action show step aka MgmtActionStatus
-  add DS metadata
- add sort for all view
- access to a specific ds by query url
- add link between target version and DS view
- local date format

I can split this PR in smaller one if you want.

### DIFF
--- a/docker/build/build_dev.sh
+++ b/docker/build/build_dev.sh
@@ -9,7 +9,7 @@
 # SPDX-License-Identifier: EPL-2.0
 #
 
-#set -xe
+set -xe
 
 # Usage: builds all docker images. Use:
 # -v <version> to pass version
@@ -20,6 +20,8 @@
 VERSION=0-SNAPSHOT
 MVN_REPO=~/.m2/repository
 TAG=latest
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+cd "${SCRIPT_DIR}"
 
 while getopts v:r:t: option
 do
--- a/hawkbit-simple-ui/pom.xml
+++ b/hawkbit-simple-ui/pom.xml
@@ -65,12 +65,6 @@
         <dependency>
             <groupId>com.vaadin</groupId>
             <artifactId>vaadin-core</artifactId>
-            <exclusions>
-                <exclusion>
-                    <groupId>com.vaadin</groupId>
-                    <artifactId>vaadin-dev</artifactId>
-                </exclusion>
-            </exclusions>
         </dependency>
         <dependency>
             <groupId>com.vaadin</groupId>
@@ -85,6 +79,11 @@
             <groupId>org.springframework.boot</groupId>
             <artifactId>spring-boot-starter-oauth2-client</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-devtools</artifactId>
+            <optional>true</optional>
+        </dependency>
     </dependencies>
 
     <build>
@@ -93,6 +92,11 @@
             <plugin>
                 <groupId>org.springframework.boot</groupId>
                 <artifactId>spring-boot-maven-plugin</artifactId>
+                <configuration>
+                    <skip>false</skip>
+                    <finalName>Simple-UI</finalName>
+                    <mainClass>org.eclipse.hawkbit.ui.simple.SimpleUIApp</mainClass>
+                </configuration>
             </plugin>
 
             <plugin>
--- a/hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/MainLayout.java
+++ b/hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/MainLayout.java
@@ -81,7 +81,7 @@ public class MainLayout extends AppLayout {
     }
 
     private void addDrawerContent() {
-        final H1 appName = new H1("hawkBit UI (Experimental!)");
+        final H1 appName = new H1("hawkBit UI");
         final HorizontalLayout layout = new HorizontalLayout();
         layout.setPadding(true);
         layout.setJustifyContentMode(FlexComponent.JustifyContentMode.CENTER);
--- a/hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/MainLayout.java
+++ b/hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/MainLayout.java
@@ -9,8 +9,11 @@
  */
 package org.eclipse.hawkbit.ui.simple;
 
+import java.util.List;
 import java.util.Optional;
 
+import com.vaadin.flow.component.Component;
+import com.vaadin.flow.component.UI;
 import com.vaadin.flow.component.Unit;
 import com.vaadin.flow.component.applayout.AppLayout;
 import com.vaadin.flow.component.applayout.DrawerToggle;
@@ -47,9 +50,12 @@ import org.eclipse.hawkbit.ui.simple.view.TargetView;
  */
 public class MainLayout extends AppLayout {
 
+    static final List<Class<? extends Component>> DEFAULT_VIEW_PRIORITY = List.of(TargetView.class, DistributionSetView.class,
+            SoftwareModuleView.class, RolloutView.class);
     private final transient AuthenticatedUser authenticatedUser;
     private final AccessAnnotationChecker accessChecker;
     private H2 viewTitle;
+    private transient Optional<Class<? extends Component>> defaultView;
 
     public MainLayout(final AuthenticatedUser authenticatedUser, final AccessAnnotationChecker accessChecker) {
         this.authenticatedUser = authenticatedUser;
@@ -68,6 +74,9 @@ public class MainLayout extends AppLayout {
                 Optional.ofNullable(getContent().getClass().getAnnotation(PageTitle.class))
                         .map(PageTitle::value)
                         .orElse(""));
+        if (UI.getCurrent().getActiveViewLocation().getPath().isEmpty()) {
+            defaultView.ifPresent(c -> UI.getCurrent().navigate(c));
+        }
     }
 
     private void addHeaderContent() {
@@ -117,6 +126,7 @@ public class MainLayout extends AppLayout {
         if (accessChecker.hasAccess(AboutView.class)) {
             nav.addItem(new SideNavItem("About", AboutView.class, VaadinIcon.INFO_CIRCLE.create()));
         }
+        defaultView = DEFAULT_VIEW_PRIORITY.stream().filter(accessChecker::hasAccess).findFirst();
         return nav;
     }
 
--- a/hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/view/DistributionSetView.java
+++ b/hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/view/DistributionSetView.java
@@ -73,7 +73,6 @@ public class DistributionSetView extends TableView<MgmtDistributionSet, Long> {
 
                     @Override
                     protected void addColumns(Grid<MgmtDistributionSet> grid) {
-                        grid.addColumn(MgmtDistributionSet::getId).setHeader(Constants.ID).setAutoWidth(true).setKey("id").setSortable(true);
                         var createdAtCol = grid.addColumn(Utils.localDateTimeRenderer(MgmtDistributionSet::getCreatedAt)).setHeader(
                                 Constants.CREATED_AT).setAutoWidth(true).setKey("createdAt").setSortable(true);
                         grid.addColumn(MgmtDistributionSet::getName).setHeader(Constants.NAME).setAutoWidth(true).setKey("name").setSortable(
--- a/hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/view/DistributionSetView.java
+++ b/hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/view/DistributionSetView.java
@@ -177,7 +177,7 @@ public class DistributionSetView extends TableView<MgmtDistributionSet, Long> {
                     Map.of(
                             "name", name.getOptionalValue(),
                             "type", type.getSelectedItems().stream().map(MgmtDistributionSetType::getKey).toList(),
-                            "tag", tag.getSelectedItems()));
+                            "tag", tag.getSelectedItems().stream().map(MgmtTag::getName).toList()));
         }
     }
 
--- a/hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/view/DistributionSetView.java
+++ b/hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/view/DistributionSetView.java
@@ -11,7 +11,6 @@ package org.eclipse.hawkbit.ui.simple.view;
 
 import java.util.Collection;
 import java.util.Collections;
-import java.util.Date;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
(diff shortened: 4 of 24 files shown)


docker/build/build_dev.sh
```diff
@@ -9,7 +9,7 @@
 # SPDX-License-Identifier: EPL-2.0
 #
 
-#set -xe
+set -xe
```
strailov, line 12: Not still sure if we need to enable this or just leave it as it is

hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/MainLayout.java
```diff
@@ -68,6 +73,9 @@ protected void afterNavigation() {
                 Optional.ofNullable(getContent().getClass().getAnnotation(PageTitle.class))
                         .map(PageTitle::value)
                         .orElse(""));
+        if(UI.getCurrent().getActiveViewLocation().getPath().isEmpty()){
```
strailov: leave some whitespaces between the if , () and {} : 
`if (<condition>) {`

hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/view/TargetView.java
```diff
@@ -552,202 +570,317 @@ private List<MgmtTag> fetchAvailableTags() {
...
         }
 
-        private List<ActionStatusEntry> fetchActions() {
-            return hawkbitClient.getTargetRestApi().getActionHistory(target.getControllerId(), null, 0, 30, null)
-                    .getBody()
-                    .getContent()
-                    .stream()
-                    .map(action -> new ActionStatusEntry(action, () -> setItems(fetchActions())))
+        @Slf4j
+        private static class TargetActions extends Grid<TargetActions.ActionStatusEntry> {
```
strailov: I would use this PR to rename this class maybe something like TargetActionHistory or sth like that. 
TargetActions indicates for manipulation possibilities over a target which is not the case here. 
I would also extract this logic into a separate class file for better readability ...

hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/view/TargetView.java
```diff
@@ -967,23 +1122,34 @@ private Icon targetUpdateStatusMapper(final String targetUpdateStatus) {
...
-            return pollIcon;
+    // todo change /targets api to reduce api calls ?
+    @EqualsAndHashCode(callSuper = true)
+    public static class MgmtTargetVersioned extends MgmtTarget{
+        MgmtTargetVersioned(){
+            super();
+        }
+        Optional<MgmtDistributionSet> ds;
+        static ObjectMapper objectMapper = new ObjectMapper();
+        @SneakyThrows
```
strailov: SneakyThrows could be dangerous to use, why we use it here ?

hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/view/TargetView.java
```diff
@@ -967,23 +1122,34 @@ private Icon targetUpdateStatusMapper(final String targetUpdateStatus) {
...
+        public static MgmtTargetVersioned from(final HawkbitMgmtClient hawkbitClient,MgmtTarget target) {
```
strailov: leave whitespaces after a comma

strailov: Why we name this class MgmtTargetVersioned ? Isn't it only extension for MgmtTarget with a DS object in it ?

hawkbit-simple-ui/pom.xml
```diff
@@ -85,6 +79,11 @@
             <groupId>org.springframework.boot</groupId>
             <artifactId>spring-boot-starter-oauth2-client</artifactId>
         </dependency>
+        <dependency>
```
avgustinmm, line 82: Why do we want a dev mode bu default?

hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/MainLayout.java
```diff
@@ -46,10 +49,12 @@
  * The main view is a top-level placeholder for other views.
  */
 public class MainLayout extends AppLayout {
-
+    static final List<Class<? extends Component>> defaultViewPriority =
```
avgustinmm: Const naming defaultViewPriority -> DEFAULT_VIEW_PRIORITY

hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/MainLayout.java
avgustinmm: please apply hawkbit codeformatters (at the project's root: if( -> if (

hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/view/DistributionSetView.java
```diff
@@ -64,24 +66,26 @@ public class DistributionSetView extends TableView<MgmtDistributionSet, Long> {
...
 
                     private final DistributionSetDetails details = new DistributionSetDetails(hawkbitClient);
 
                     @Override
                     protected void addColumns(Grid<MgmtDistributionSet> grid) {
-                        grid.addColumn(MgmtDistributionSet::getId).setHeader(Constants.ID).setAutoWidth(true);
-                        grid.addColumn(MgmtDistributionSet::getName).setHeader(Constants.NAME).setAutoWidth(true);
-                        grid.addColumn(MgmtDistributionSet::getVersion).setHeader(Constants.VERSION).setAutoWidth(true);
-                        grid.addColumn(MgmtDistributionSet::getTypeName).setHeader(Constants.TYPE).setAutoWidth(true);
+                        var createdAtCol = grid.addColumn(Utils.localDateTimeRenderer(MgmtDistributionSet::getCreatedAt)).setHeader(Constants.CREATED_AT).setAutoWidth(true).setKey("createdAt").setSortable(true);
```
avgustinmm: Why is the change? createdAtCol os never used?

docker/build/build_dev.sh
avgustinmm, line 12: is this uncommented intentionally?

hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/view/TargetView.java
```diff
@@ -967,23 +1122,34 @@ private Icon targetUpdateStatusMapper(final String targetUpdateStatus) {
...
-            } else if (pollStatus.isOverdue()) {
-                pollIcon = Utils.tooltip(VaadinIcon.EXCLAMATION_CIRCLE.create(), "Overdue");
-            } else {
-                pollIcon = Utils.tooltip(VaadinIcon.CLOCK.create(), "In Time");
-            }
-            pollIcon.addClassNames(LumoUtility.IconSize.SMALL);
-            return pollIcon;
+    // todo change /targets api to reduce api calls ?
+    @EqualsAndHashCode(callSuper = true)
+    public static class MgmtTargetVersioned extends MgmtTarget{
```
avgustinmm: Maybe TargetWithDs ?
1. MgmtSomething pattern is used for management rest classes and is little bit confusing here
2. Target version is not used as a term in hawkBit , and actually this is the installed DS

docker/build/build_dev.sh
flobz: If a command failed it's better to exit immediately don't you think ? 
"-x" isn't necessary but could be useful to understand an error.

flobz: yes

hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/view/TargetView.java
```diff
@@ -89,33 +93,41 @@
...
-                                .setFlexGrow(0);
-                        grid.addColumn(MgmtTarget::getControllerId).setHeader(CONTROLLER_ID).setAutoWidth(true);
-                        grid.addColumn(MgmtTarget::getName).setHeader(Constants.NAME).setAutoWidth(true);
-                        grid.addColumn(MgmtTarget::getTargetTypeName).setHeader(Constants.TYPE).setAutoWidth(true);
+                                .setFlexGrow(0).setKey("lastControllerRequestAt").setSortable(true);
+                        grid.addColumn(new ComponentRenderer<>(TargetUpdateStatusCell::new))
+                                .setHeader(UPDATE)
+                                .setAutoWidth(true)
+                                .setFlexGrow(0).setKey("updateStatus").setSortable(true);
+                        grid.addColumn(MgmtTarget::getControllerId).setHeader(FILTER).setAutoWidth(true).setKey("id").setSortable(true);
```
strailov: Why the header for controllerId here is "filter" ?

hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/view/TargetView.java
```diff
@@ -346,6 +363,7 @@ private static class TargetDetails extends FormLayout {
         private final TextField securityToken = Utils.textField(Constants.SECURITY_TOKEN);
         private final TextField lastPoll = Utils.textField(Constants.LAST_POLL);
         private final TextField group = Utils.textField(Constants.GROUP);
+        private final TextField ipAddress = Utils.textField(Constants.IP);
```
strailov: It's non-necessary that this is an IP address of a device - it could be other kind of address - so I would leave the name of the field here just "Address"

hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/view/util/Filter.java
```diff
@@ -84,34 +85,62 @@ public Filter(final Consumer<String> changeListener, final Rsql primaryRsql, fin
...
-                            } else {
-                                return e.getValue() != null;
-                            }
+                        .map( e->{
+                                if (e.getValue() instanceof Optional<?> opt) {
+                                    e.setValue(opt.orElse(null));
+                                }
+                                return e;
                         })
+                        .filter(e -> e.getValue() != null)
```
strailov: Can we sum those 2 filters into one ? 
We could get use of ObjectUtils.isEmpty() also for null & empty check at once ?

hawkbit-simple-ui/pom.xml
flobz: the [doc](https://docs.spring.io/spring-boot/reference/using/devtools.html) says : `Repackaged archives do not contain devtools by default` so it seems ok to add it to the pom.

hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/MainLayout.java
flobz: done

hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/MainLayout.java
flobz: done

hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/view/DistributionSetView.java
flobz: it's used to sort : 
```
grid.sort(List.of(new GridSortOrder<>(createdAtCol, SortDirection.DESCENDING)));
```

hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/view/TargetView.java
flobz: done ! 
Wdyt about adding installed DS name and version in /targets endpoint ?

hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/MainLayout.java
flobz: done

hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/view/TargetView.java
flobz: done

hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/view/TargetView.java
flobz: I delete it was useless

hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/view/TargetView.java
flobz: done

flobz: yes I rename it to TargetWithDs

hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/view/TargetView.java
flobz: I was a mistake, sorry !

hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/view/TargetView.java
flobz: done

hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/view/util/Filter.java
flobz: done

hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/view/TargetView.java
strailov: It's still visible as IP in the UI ...

hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/view/TargetView.java
strailov: Still the header in the grid is Filter in the UI ... You sure you pushed your changes ?

hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/view/util/Filter.java
strailov: Still I see the old approach, sure you've pushed your changes ?

hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/view/TargetView.java
flobz: forgot to push the last version, sorry !

pom.xml
```diff
@@ -368,6 +369,32 @@
...
             </plugin>
+            <plugin>
+                <!--
+                https://github.com/diffplug/spotless/tree/main/plugin-maven#-spotless-keep-your-code-spotless-with-maven
+
+                Usage:
+                mvn spotless:apply
+                mvn spotless:check
+                -->
+                <groupId>com.diffplug.spotless</groupId>
```
strailov: Introducing another plugin probably should be done in separate PR for the following reasons : 

- It's not related to this particular PR.
- Not 100% sure we want one more plugin in our pom.xml currently.
- Could have a separate discussion for pros and cons in the separate pr. 

Everything else seems fine IMHO and if you remove the plugin I believe we could merge the other part.

flobz: ok sure I removed the plugin and I will create another PR.

PR #2562, opened 2025-07-23
Fine grained repository permissions

avgustinmm:


hawkbit-mgmt/hawkbit-mgmt-resource/src/main/java/org/eclipse/hawkbit/mgmt/rest/resource/MgmtDistributionSetResource.java
```diff
@@ -260,7 +261,7 @@ public ResponseEntity<MgmtTargetAssignmentResponseBody> createAssignedTarget(
         }).toList();
 
         final List<DistributionSetAssignmentResult> assignmentResults = deployManagement.assignDistributionSets(deploymentRequests);
-        return ResponseEntity.ok(MgmtDistributionSetMapper.toResponse(assignmentResults));
+        return ResponseEntity.ok(mgmtDistributionSetMapper.toResponse(assignmentResults));
```
strailov, line 264: Why toResponse remains static, but fromRequest not ? Wouldn't be better both to be non-static ?

hawkbit-mgmt/hawkbit-mgmt-resource/src/main/java/org/eclipse/hawkbit/mgmt/rest/resource/MgmtDistributionSetResource.java
```diff
@@ -316,9 +317,8 @@ public ResponseEntity<PagedList<MgmtSoftwareModule>> getAssignedSoftwareModules(
             final Long distributionSetId,
             final int pagingOffsetParam, final int pagingLimitParam, final String sortParam) {
         final Pageable pageable = PagingUtility.toPageable(pagingOffsetParam, pagingLimitParam, sanitizeDistributionSetSortParam(sortParam));
-        final Page<SoftwareModule> softwareModules = softwareModuleManagement.findByAssignedTo(distributionSetId, pageable);
-        return ResponseEntity.ok(new PagedList<>(MgmtSoftwareModuleMapper.toResponse(
-                softwareModules.getContent()), softwareModules.getTotalElements()));
+        final Page<? extends SoftwareModule> softwareModules = softwareModuleManagement.findByAssignedTo(distributionSetId, pageable);
+        return ResponseEntity.ok(new PagedList<>(MgmtSoftwareModuleMapper.toResponse(softwareModules.getContent()), softwareModules.getTotalElements()));
```
strailov, line 321: Same for the static/non-static approach here.

hawkbit-repository/hawkbit-repository-core/src/main/java/org/eclipse/hawkbit/repository/RepositoryConfiguration.java
```diff
@@ -52,10 +56,37 @@ static RoleHierarchy roleHierarchy() {
...
+    PermissionEvaluator permissionEvaluator(final RoleHierarchy roleHierarchy) {
+        return new DenyAllPermissionEvaluator() {
+
+            @Override
+            public boolean hasPermission(final Authentication authentication, final Object targetDomainObject, final Object permission) {
+                if (targetDomainObject instanceof MethodSecurityExpressionOperations root) {
+                    final String neededPermission =
+                            permission + "_" + (root.getThis() instanceof RepositoryManagement<?, ?, ?> repositoryManagement
+                                    ? repositoryManagement.permissionGroup()
+                                    : "REPOSITORY"); // TODO - should not fall back here - all using parmissions should extend repository management interface
```
strailov, line 69: I guess it is not appropriate to throw exception here ? Than shall we leave it this way and remove the TODO or ?

strailov (approved):

hawkbit-mgmt/hawkbit-mgmt-resource/src/main/java/org/eclipse/hawkbit/mgmt/rest/resource/MgmtDistributionSetResource.java
avgustinmm: whatever could be static is better to be static.

hawkbit-repository/hawkbit-repository-core/src/main/java/org/eclipse/hawkbit/repository/RepositoryConfiguration.java
avgustinmm: it falls back to deny all which will throw exception. We handle whatever we know, and the rest leave to default implementation. At some point - this TODO, may be removed - too early

PR #2563, opened 2025-07-23
Remote Events migrated from Spring Bus to Spring Cloud Stream

vasilchev:
In progress... message converter check and doc remain


hawkbit-ddi/hawkbit-ddi-server/src/main/resources/application.properties
```diff
@@ -50,15 +49,15 @@ hawkbit.lock=inMemory
...
-# Configure communication between services
-endpoints.spring.cloud.bus.refresh.enabled=false
-endpoints.spring.cloud.bus.env.enabled=false
-spring.cloud.stream.bindings.springCloudBusInput.group=ddi-server
-
-# To use protostuff (for instance fot improved performance) you shall uncomment
-# the following two lines and add io.protostuff:protostuff-core and io.protostuff:protostuff-runtime to dependencies
-#spring.cloud.stream.bindings.springCloudBusInput.content-type=application/binary+protostuff
-#spring.cloud.stream.bindings.springCloudBusOutput.content-type=application/binary+protostuff
+# Enable communication between services - set 'org.eclipse.hawkbit.events.remote-enabled=true'
```
avgustinmm: Shouldn't it be enabled by default?
Also maybe comment shall be with less duplication, e.g.:
```properties
# Disables communication between services. To enable set to 'true'
```

hawkbit-ddi/hawkbit-ddi-server/src/main/resources/application.properties
```diff
@@ -50,15 +49,15 @@ hawkbit.lock=inMemory
...
+org.eclipse.hawkbit.events.remote-enabled=false
+spring.cloud.function.definition=fanoutEventConsumer;groupEventConsumer
+spring.cloud.stream.bindings.groupEventConsumer-in-0.group=${spring.application.name}
+# Optional: Use protostuff (if enabled)
```
avgustinmm, line 54: Since now we have more complex config maybe we could simplify by something like:
```properties
# to enable protobuf set content-type to application/binary+protostuff
hawkbit.stream.content-type=<default one, e.g. application/json?>
spring.cloud.stream.bindings.fanoutEventConsumer-in-0.content-type=${hawkbit.stream.content-type}
...
```
Maybe even spring provides a config for default message type?

hawkbit-mgmt/hawkbit-mgmt-server/README.md
```diff
@@ -24,37 +24,65 @@ run org.eclipse.hawkbit.app.mgmt.MgmtServerStart
...
+If multiple management server instances are needed, you must extend hawkBit to ensure that scheduled tasks do not run concurrently.
+Alternatively, configure all but one instance to disable scheduler execution.
 
-## Optional Protostuff for Spring cloud bus
+## Event Channel Types in Spring Cloud Stream
 
-The micro-service instances are configured to communicate via Spring Cloud Bus. Optionally, you could
+Remote events in hawkBit are distributed through **two distinct types of channels**:
+
+### 1. Fanout Event Channel
```
avgustinmm, line 43: shouldn't this be put into the /site content and here (and to other readme-s) to just have reference to there?

hawkbit-mgmt/hawkbit-mgmt-server/src/main/resources/application.properties
```diff
@@ -43,15 +42,15 @@ hawkbit.server.repository.publish-target-poll-event=false
 
 # Disable discovery client of spring-cloud-commons
```
avgustinmm, line 43: this config, nevertheless if in this way or in proposed - isn't the same for (almost) all services - should be put as default in some in-jar properties?

hawkbit-repository/hawkbit-repository-api/src/main/java/org/eclipse/hawkbit/repository/event/EventPublisherHolder.java
```diff
@@ -9,63 +9,158 @@
...
- * A singleton bean which holds the event publisher and service origin id in order to publish remote application events.
- * It can be used in beans not instantiated by spring e.g. JPA entities which cannot be auto-wired.
- */
-@NoArgsConstructor(access = AccessLevel.PRIVATE)
-@SuppressWarnings("java:S6548") // java:S6548 - singleton holder ensures static access to spring resources in some places
+@Slf4j
 public final class EventPublisherHolder {
 
-    private static final EventPublisherHolder SINGLETON = new EventPublisherHolder();
+    @Value("${org.eclipse.hawkbit.events.remote-enabled:false}")
```
avgustinmm: shouldn't we have remote enabled by default? for all microservices it shall be true?
only disabled in monolith?

hawkbit-repository/hawkbit-repository-api/src/main/java/org/eclipse/hawkbit/repository/event/EventPublisherHolder.java
```diff
@@ -9,63 +9,158 @@
...
-        if (id == null && bus != null) {
-            id = bus.getId();
+
+        @Override
+        public void publishEvent(final ApplicationEvent event) {
+            routeEvent(event);
+        }
+
+        private void routeEvent(Object event) {
+            if (remoteEventsEnabled && streamBridge != null && event instanceof AbstractRemoteEvent remoteEvent) {
```
avgustinmm: shouldn't configuration be checked for consistency at the beginning to prevent starting wrongly configured app?
Also, do we have non-remove events? then we always need delegate? Do we want to throw in that case or skip?

hawkbit-repository/hawkbit-repository-api/src/main/java/org/eclipse/hawkbit/repository/event/EventPublisherHolder.java
```diff
@@ -9,63 +9,158 @@
...
+
+        private void publishLocally(final Object event) {
+            delegate.publishEvent(event);
+
+            // even though the event is published locally
+            // we still want to check if it should be forwarded as a grouped event - e.g. monolith
+            if (shouldForwardAsProcessingEvent(event)) {
+                final AbstractRemoteEvent processingEvent = toProcessingEvent((AbstractRemoteEvent) event);
+                if (processingEvent != null) {
+                    log.debug("Publishing grouped event: {} to locally: {}", processingEvent, groupChannel);
```
avgustinmm: Shouldn't term be "Group event" (not grouped)?
An maybe "Fanout event"?

hawkbit-repository/hawkbit-repository-core/src/main/java/org/eclipse/hawkbit/event/EventType.java
```diff
@@ -166,6 +176,15 @@ public class EventType {
         TYPES.put(44, TargetTypeCreatedEvent.class);
         TYPES.put(45, TargetTypeUpdatedEvent.class);
         TYPES.put(46, TargetTypeDeletedEvent.class);
+
+        // processing events
+        TYPES.put(47, ProcessingTargetCreatedEvent.class);
```
avgustinmm: maybe leave some space for next db / fanout events? e.g. start heir order from 1000?

hawkbit-repository/hawkbit-repository-core/src/main/resources/hawkbit-events-defaults.properties
```diff
@@ -0,0 +1,25 @@
...
+# which is available at https://www.eclipse.org/legal/epl-2.0/
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+
+# Enable communication between services - set 'org.eclipse.hawkbit.events.remote-enabled=true'
+org.eclipse.hawkbit.events.remote-enabled=false
+
+# -- Consumer bindings --
+spring.cloud.stream.bindings.fanoutEventConsumer-in-0.destination=fanoutEventChannel
```
avgustinmm, line 17: maybe add here all repeating default config from servers?

hawkbit-repository/hawkbit-repository-jpa/src/test/java/org/eclipse/hawkbit/repository/jpa/management/ControllerManagementTest.java
```diff
@@ -1369,15 +1472,19 @@ void quotaExceededExceptionWhenControllerReportsTooManyUpdateActionStatusMessage
...
     @ExpectEvents({
             @Expect(type = TargetCreatedEvent.class, count = 1),
+            @Expect(type = ProcessingTargetCreatedEvent.class, count = 1),
             @Expect(type = DistributionSetCreatedEvent.class, count = 1),
             @Expect(type = SoftwareModuleCreatedEvent.class, count = 3),
             @Expect(type = DistributionSetUpdatedEvent.class, count = 1), // implicit lock
             @Expect(type = SoftwareModuleUpdatedEvent.class, count = 3), // implicit lock
             @Expect(type = ActionCreatedEvent.class, count = 1),
             @Expect(type = TargetAttributesRequestedEvent.class, count = 1),
+            @Expect(type = ProcessingTargetAttributesRequestedEvent.class, count = 1),
```
avgustinmm: Couldn't we just tweak the event expectation checker to expect by default the same "processing" event count as the "db/fanout" events and not to repeat them everywhere?

hawkbit-repository/hawkbit-repository-api/src/main/java/org/eclipse/hawkbit/repository/event/remote/AbstractProcessingRemoteEvent.java
```diff
@@ -0,0 +1,24 @@
...
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.hawkbit.repository.event.remote;
+
+import lombok.Getter;
+
+@Getter
+public abstract class AbstractProcessingRemoteEvent<T extends AbstractRemoteEvent> extends AbstractRemoteEvent {
```
avgustinmm: We have fanout and group events - since the 'regular' db events are fanout - maybe these events shall be called "GroupXXX" events, i.e. event for a group?
Also maybe we shall call fanout events just "events" (not fanout events) - regular events to all hosts

hawkbit-repository/hawkbit-repository-core/src/main/resources/hawkbit-events-defaults.properties
vasilchev: Spring Cloud does not rely on @PropertySource for binding setup. 
It reads configuration directly from the Environment during autoconfiguration, before @PropertySource is evaluated
some of them are needed early, some are evaluated after environment is ready, that is why we have mixture - whatever possible is in the hawkbit-events-defaults and what not in application.properties (like functional binding)

hawkbit-repository/hawkbit-repository-api/src/main/java/org/eclipse/hawkbit/repository/event/remote/AbstractProcessingRemoteEvent.java
vasilchev: Initially i've named them GroupXXX but somehow got the feeling that name is misleading e.g. GroupTargetCreatedEvent - > a group of targets created.
I agree Processing* is not the best, but rather choose a different prefix that does not have a misleading effect.

hawkbit-mgmt/hawkbit-mgmt-server/src/main/resources/application.properties
vasilchev: as mentioned in other comment some properties from spring cloud stream need to be processed early - i wll check options to implement central properties and in a way to be loaded in the needed phase

hawkbit-repository/hawkbit-repository-jpa/src/test/java/org/eclipse/hawkbit/repository/jpa/management/ControllerManagementTest.java
vasilchev: yes, event checker actually pareses the annotations and is in our hands.
fixed it

hawkbit-repository/hawkbit-repository-core/src/main/java/org/eclipse/hawkbit/event/EventType.java
vasilchev: started Processing* events from 1000

hawkbit-repository/hawkbit-repository-api/src/main/java/org/eclipse/hawkbit/repository/event/EventPublisherHolder.java
vasilchev: yes, makes sense

hawkbit-ddi/hawkbit-ddi-server/src/main/resources/application.properties
vasilchev: turned out it indeed has default

'spring.cloud.stream.default.content-type'

fixed it

hawkbit-mgmt/hawkbit-mgmt-server/src/main/resources/application.properties
vasilchev: I've added the needed properties in hawkbit-event-defaults.properties inside hawkbit-repository-jpa
now all services 'spring.config.import' it and reuse

hawkbit-repository/hawkbit-repository-core/src/main/resources/hawkbit-events-defaults.properties
vasilchev: 've added the needed properties in hawkbit-event-defaults.properties inside hawkbit-repository-jpa
now all services 'spring.config.import' it and reuse

hawkbit-repository/hawkbit-repository-api/src/main/java/org/eclipse/hawkbit/repository/event/EventPublisherHolder.java
vasilchev: > shouldn't configuration be checked for consistency at the beginning to prevent starting wrongly configured app?

Yes, definitely! fixed.

non-remote events are always published only locally via delegete (regardless of remoteEvents enabled)

Only when remoteEvents enabled and event is instance of AbstractRemoteEvent (i.e. RemoteEvent) only then we use the streamBridge (post it to broker, and depending on fanount/group replicas receive it and publisch it within the app context)

hawkbit-repository/hawkbit-repository-api/src/main/java/org/eclipse/hawkbit/repository/event/remote/AbstractProcessingRemoteEvent.java
avgustinmm: or TargetCreatedGroupEvent
we shall start from correct terms
* events & group events
* fanout & group events
* events & service events (service means dmf/ddi/update server)
* fanout & service events
where the first type is "all nodes", second "to single of group service"
Then we could specifiy naming ..

hawkbit-repository/hawkbit-repository-core/src/main/resources/hawkbit-events-defaults.properties
```diff
@@ -9,8 +9,11 @@
 #
 
 # Enable communication between services - set 'org.eclipse.hawkbit.events.remote-enabled=true'
-org.eclipse.hawkbit.events.remote-enabled=false
+org.eclipse.hawkbit.events.remote-enabled=true
```
avgustinmm, line 11: maybe now we shall disable it for monolith?

vasilchev: hawkbit/hawkbit-monolith/hawkbit-update-server/src/main/resources/application.properties

Isn't it the correct way disabled ?

hawkbit-repository/hawkbit-repository-api/src/main/java/org/eclipse/hawkbit/repository/event/remote/AbstractProcessingRemoteEvent.java
vasilchev: for me they are events and processingevents (or even PostProcessingEvent) that gives the closest meaning

Otherwise TargetCreatedGroupEvent events will not have a common prefix to easily review them (like Group* or Processing* or Post*)

otherwise they are all service events (naming the fanout just events and the others service events seem like not all events are coming from services)

vasilchev: So with latest propopsal we have

fanoutEventChannel -> *Events
serviceEventChannel -> *ServiceEvents

*Service events moved to separate package 'service'

Lets stick to 
- fanout events -> fanaoutEventChannel -> all replicas receive - e.g. cache invalidation db events
- service events -> serviceEventChannel -> wrapped events -> single replica recieve and process (e.g. dmf)

hawkbit-repository/hawkbit-repository-core/pom.xml
```diff
@@ -50,5 +50,9 @@
             <artifactId>protostuff-runtime</artifactId>
             <optional>true</optional>
         </dependency>
+        <dependency>
```
avgustinmm: this should not be needed - it should come via hawkbit-repository-api?

vasilchev: removed

PR #2568, opened 2025-07-28
Soft Module metadata as complex map value

avgustinmm:
* Removes SoftwareModuleMetadata - repository, builders
* Replace complex value (value & target visible) with MetadataValue in SoftwareModuleManagement, thus unify with Target and DS metadata - String key <-> value (string or more complex)
* clean up the methods - createMetadata -> create or update
* JpaSoftwareModule.metadate -> Map<String, JpaMetadataValue> 
* extend RSQL to spec to support filtering for value in complex object "value" field
* remove get target visible metadata by single software module - in reality we always want metadata for all software modules of a DS


hawkbit-repository/hawkbit-repository-api/src/main/java/org/eclipse/hawkbit/repository/model/SoftwareModule.java
```diff
@@ -81,4 +86,28 @@ default Optional<Artifact> getArtifactByFilename(final String fileName) {
...
     }
+
+    interface MetadataValue {
+
+        String getValue();
+        boolean isTargetVisible();
+    }
+
+    @NoArgsConstructor
+    @Data
```
vasilchev, line 97: MetadataValues could contain sensitive information (e.g. encrypt/decrypt information) not sure how to avoid but toString potentionaly at risk

hawkbit-repository/hawkbit-repository-api/src/main/java/org/eclipse/hawkbit/repository/SoftwareModuleManagement.java
```diff
@@ -84,52 +80,30 @@ default String permissionGroup() {
...
      * @param id where meta-data has to be deleted
      * @param key of the meta-data element
-     * @return true if really deleted, false if not
      * @throws EntityNotFoundException if software module with given ID does not exist or the key is not found
      */
     @PreAuthorize(SpringEvalExpressions.HAS_UPDATE_REPOSITORY)
-    void deleteMetadata(long id, @NotEmpty String key);
+    void deleteMetadata(@NotNull Long id, @NotEmpty String key);
+
+    @PreAuthorize(SpringEvalExpressions.IS_SYSTEM_CODE)
```
vasilchev, line 105: before this method needed READ_REPOSITORY, now it is SYSTEM_CODE
Is it always executed in system context ?

hawkbit-repository/hawkbit-repository-api/src/main/java/org/eclipse/hawkbit/repository/model/SoftwareModule.java
avgustinmm: JpaDistributionSet also have metadata and in toString the map shall be visible
metadata of a software module should not be something security sensitive. It could be (maybe ?) for target attributes or metadata, but for DS and SM I think it should not be

hawkbit-repository/hawkbit-repository-api/src/main/java/org/eclipse/hawkbit/repository/SoftwareModuleManagement.java
avgustinmm: DDI calls it (now) in runAsSystem and DMF should also do it in priviledged scope. it is not used by users with repository only scopes anyway

PR #2595, opened 2025-08-05
Introduce Stop of a rollout

strailov:
This PR is introducing the following changes:

1. Stop of a rollout - will soft cancel all active actions (put them in CANCELING state) and will notify the devices for this if they decide to perform a cancellation. This will put the rollout into STOPPING state and eventually in STOPPED state when no active actions are left (canceling state here is not considered active - which means that the rollout will stop and some actions could remain in canceling state)
2. Delete of rollout behavior changed - until now the deletion of rollout was leaving all the active actions in the state that they are, deleting only the scheduled ones. In this implementation the deletion will force cancel all the active action for the rollout (put them in CANCELED state)
3. Invalidation of DS changes 
- removed cancel rollout option
- NONE cancellation type will only invalidate the provided DS nothing more
- SOFT will perform soft cancel and will set all the rollouts that are associated with the provided DS to STOPPING state (which will lead to stop of a rollout case)
- FORCE will perform force cancel and will set all the rollouts that are associated with the provided DS to DELETING state (which will lead to delete of rollout)

### DIFF
--- a/hawkbit-repository/hawkbit-repository-jpa/src/test/java/org/eclipse/hawkbit/repository/jpa/management/RolloutManagementTest.java
+++ b/hawkbit-repository/hawkbit-repository-jpa/src/test/java/org/eclipse/hawkbit/repository/jpa/management/RolloutManagementTest.java
@@ -1918,7 +1918,7 @@ class RolloutManagementTest extends AbstractJpaIntegrationTest {
             @Expect(type = RolloutCreatedEvent.class, count = 1),
             @Expect(type = RolloutUpdatedEvent.class, count = 6),
             @Expect(type = RolloutDeletedEvent.class, count = 1),
-            @Expect(type = RolloutGroupUpdatedEvent.class, count = 11),
+            @Expect(type = RolloutGroupUpdatedEvent.class, count = 16),
             @Expect(type = RolloutGroupCreatedEvent.class, count = 5) })
     void deleteRolloutWhichHasBeenStartedBeforeIsSoftDeleted() {
         final int amountTargetsForRollout = 10;


hawkbit-repository/hawkbit-repository-api/src/main/java/org/eclipse/hawkbit/repository/DeploymentManagement.java
```diff
@@ -413,4 +414,13 @@ List<DistributionSetAssignmentResult> assignDistributionSets(
...
     void cancelActionsForDistributionSet(final CancelationType cancelationType, final DistributionSet set);
+
+    /**
+     * Cancels all actions that refer to a given rollout.
+     *
+     * @param cancelationType - type of cancellation - FORCE or SOFT (NONE is ignored)
+     * @param rollout - the rollout which actions are about to be cancelled
+     */
+    @PreAuthorize(SpringEvalExpressions.HAS_AUTH_ROLLOUT_MANAGEMENT_UPDATE)
+    void cancelActiveActionsForRollouts(final Rollout rollout, final CancelationType cancelationType);
```
avgustinmm: This is a rollouts related method - why not in RolloutManagement?
Maybe CancelationType could be moved on more central place - as for instance ActionManagement?
Since it is related to RolloutManagement, DistributionSetInvalidation & (?) DeploymentManagement

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaDeploymentManagement.java
```diff
@@ -551,6 +563,106 @@ protected boolean isActionsAutocloseEnabled() {
         return getConfigValue(REPOSITORY_ACTIONS_AUTOCLOSE_ENABLED, Boolean.class);
     }
 
+    private void softCancelActionsOfRollout(final Rollout rollout) {
+        final List<JpaAction> actions = actionRepository.findAll(
+                ActionSpecifications
+                        .byRolloutIdAndActiveAndStatusIsNot(rollout.getId(),
+                                Arrays.asList(Status.CANCELING, Status.CANCELED)), // avoid cancelling state here, because it is count as still active
+                Pageable.ofSize(999))
```
avgustinmm: better to be configurable. In JpaRolloutExecutor there are similar limits TRANSACTION_ACTIONS/TARGETS 5_000. So maybe this should be bigger? not 999. And better all (including TRANSACTION_ACTIONS/TARGETS 5_000) to be configurable (e.g. with system / spring property)

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaDeploymentManagement.java
```diff
@@ -551,6 +563,106 @@ protected boolean isActionsAutocloseEnabled() {
...
+        log.info("Found {} active actions for rollout {}, performing soft cancel.", actions.size(), rollout.getId());
+
+        final List<JpaActionStatus> cancellingStatuses = new ArrayList<>(actions.size());
+        actions.forEach(action -> {
+            action.setStatus(Status.CANCELING);
+
+            JpaActionStatus status = new JpaActionStatus();
+            status.setAction(action);
+            status.setStatus(Status.CANCELING);
+            status.setOccurredAt(System.currentTimeMillis());
```
avgustinmm: system time shall be got at once - so all cancels in transaction to have same occured at time

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaDeploymentManagement.java
```diff
@@ -551,6 +563,106 @@ protected boolean isActionsAutocloseEnabled() {
...
+        // send cancellation messages to event publisher
+        onlineDsAssignmentStrategy.cancelAssignments(actions, tenantAware.getCurrentTenant());
+    }
+
+    private void forceQuitActionsOfRollout(final Rollout rollout) {
+        final List<JpaAction> actions = findActiveActionsForRollout(rollout.getId(), Pageable.ofSize(999))
+                .getContent();
+        log.info("Found {} active actions for rollout {}", actions.size(), rollout.getId());
+
+        final List<JpaActionStatus> cancelledStatuses = new ArrayList<>(actions.size());
```
avgustinmm: same code as in previous method, just different status. message and active to false- should be reused

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaDeploymentManagement.java
```diff
@@ -551,6 +563,106 @@ protected boolean isActionsAutocloseEnabled() {
...
+
+        // if no active actions
+        // set assignedDs to previously installedDs and status to IN_SYNC
+        // otherwise set assigned ds to the active action ...
+        updateTargetAssignedDsWithInstalledIfNoActiveActions(targetIds);
+    }
+
+    private int updateTargetAssignedDsWithFirstActiveAction(List<Long> targetIds) {
+        final Query updateQuery = entityManager.createNativeQuery(
+                "UPDATE sp_target t " +
```
avgustinmm: does that set null for the targets without DS?

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaDeploymentManagement.java
```diff
@@ -551,6 +563,106 @@ protected boolean isActionsAutocloseEnabled() {
...
+        );
+        Jpa.setNativeQueryInParameter(updateQuery, "tid", targetIds);
+        final int updated = updateQuery.executeUpdate();
+        log.info("{} of target assigned distribution values updated for tenant {}",
+                updated, tenantAware.getCurrentTenant());
+        return updated;
+    }
+
+    private int updateTargetAssignedDsWithInstalledIfNoActiveActions(List<Long> targetIds) {
+        final Query updateQuery = entityManager.createNativeQuery(
```
avgustinmm: should we set assigned ds if not active? isn't better to have null?

hawkbit-repository/hawkbit-repository-api/src/main/java/org/eclipse/hawkbit/repository/DeploymentManagement.java
strailov: There is no ActionManagement no more I think. However I agree it should be in more central place.

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaDeploymentManagement.java
strailov: Yes. It sets null for targets with no previously assigned DS.

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaDeploymentManagement.java
strailov: This is the current behaviour of single forceQuitAction. I guess the bulk way should do the same. 
And yeah - this basically reverts the state before the assignment, so I think it is the right way.

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaDistributionSetInvalidationManagement.java
```diff
@@ -141,14 +141,14 @@ private void invalidateDistributionSet(final long setId, final CancelationType c
...
+        if (shouldRolloutsBeCanceled(cancelationType)) {
             log.debug("Cancel rollouts after ds invalidation. ID: {}", setId);
-            rolloutManagement.cancelRolloutsForDistributionSet(distributionSet);
+            rolloutManagement.cancelRolloutsForDistributionSet(distributionSet, cancelationType);
         }
 
         // Do run as system to ensure all actions (even invisible) are canceled due to invalidation.
         systemSecurityContext.runAsSystem(() -> {
-            if (cancelationType != CancelationType.NONE) {
+            if (cancelationType != ActionCancellationType.NONE) {
```
avgustinmm: Better check outside runAsSystem, so to skip context switch if not needed

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaRolloutManagement.java
```diff
@@ -459,6 +514,124 @@ public void triggerNextGroup(final long rolloutId) {
...
+        } else if (ActionCancellationType.SOFT.equals(cancelationType)) {
+            softCancelActionsOfRollout(rollout);
+        }
+    }
+
+    private void softCancelActionsOfRollout(final Rollout rollout) {
+        final List<JpaAction> actions = actionRepository.findAll(
+                        ActionSpecifications
+                                .byRolloutIdAndActiveAndStatusIsNot(rollout.getId(),
+                                        Arrays.asList(Action.Status.CANCELING, Action.Status.CANCELED)), // avoid cancelling state here, because it is count as still active
```
avgustinmm: are CANCELED active?
* if not - should be added in IsNot
* if yes - then in JpaRolloutExecutor.delete - we set all to canceled and then wait all actions to be inactive before delete - why?

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/OnlineDsAssignmentStrategy.java
```diff
@@ -184,6 +184,14 @@ void cancelAssignment(final JpaAction action) {
         }
     }
 
+    void cancelAssignments(final List<JpaAction> actions, final String tenant) {
```
avgustinmm: if this method just sends events - rename to something like "send events"

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaRolloutManagement.java
strailov: Hm - here CANCELED should be totally removed from the filter because its irrelevant (CANCELED are considered inactive) - We should search only for active actions that are not in CANCELING state.

PR #2660, opened 2025-09-09
Finalize and polish fine-grained permission

avgustinmm:
* Remove _REPOSITORY_ permissions -> replaced with _SOFTWARE_MODULE_, _SOFTWARE_MODULE_TYPE_, _DISTRIBUTION_SET_, _DISTRIBUTION_SET_TYPE_ permissions
* Still kept _ROLE_REPOSITORY_ADMIN_ role granting all repository fine-graned permissions
* Added dedicated _TARGET_TYPE_ permission set - the _TARGET_ permissions just grant _READ_TARGET_TYPE_ (analogically _SOFTWARE_MODULE_ permissions grant _READ_SOFTWARE_MODULE_TYPE_ and _DISTRIBUTION_SET_ grants _READ_DISTRIBUTON_SET_TYPE_
* Hierarcy is not configurable - could be completely replaced by setting spring application property org.eclipse.hawkbit.hierarchy or could be extended by adding rules using org.eclipse.hawkbit.hierarchy.ext


hawkbit-mgmt/hawkbit-mgmt-resource/src/main/java/org/eclipse/hawkbit/mgmt/rest/resource/mapper/MgmtDistributionSetMapper.java
```diff
@@ -170,36 +201,6 @@ public static List<MgmtDistributionSet> toResponseFromDsList(final List<? extend
...
-            modules.add(dsRest.getApplication().getId());
-        }
-        if (dsRest.getRuntime() != null) {
-            modules.add(dsRest.getRuntime().getId());
-        }
-        if (dsRest.getModules() != null) {
-            dsRest.getModules().forEach(module -> modules.add(module.getId()));
-        }
-        return DistributionSetManagement.Create.builder()
-                .type(Optional.ofNullable(dsRest.getType())
```
vasilchev, line 188: before it was possible to do a dsRest create request without specifiying the type ? in that case default one is used?
is this something legacy that now we do not support ? I cannot find anywhere we state it is required prop?

vasilchev: what is the usage of default ds type key in this case - i do not see need of this config now

avgustinmm: See the caller (AFAIR) the single caller - before calling that method all null-s are replaces with default type

hawkbit-monolith/hawkbit-update-server/src/test/java/org/eclipse/hawkbit/app/PreAuthorizeEnabledTest.java
```diff
@@ -69,8 +68,7 @@ void successIfHasRoleWithScope() throws Exception {
      */
     @Test
     @WithUser(authorities = {
-            SpPermission.CREATE_REPOSITORY,
-            SpPermission.READ_REPOSITORY,
+            "CREATE_DISTRIBUTION_SET", "READ_DISTRIBUTION_SET_TYPE",
```
vasilchev, line 71: for this test and others(create) maybe READ_*_TYPE permission should be better skipped, proofing that it is enough to have CREATE_DISTRIBUTION_SET to create (implying create* -> read_*_type)

hawkbit-repository/hawkbit-repository-jpa/src/test/java/org/eclipse/hawkbit/repository/jpa/management/DistributionSetInvalidationManagementTest.java
```diff
@@ -182,74 +176,63 @@ void verifyInvalidateIncompleteDistributionSetThrowsException() {
...
     }
 
     /**
-     * Verify that a user that has authority READ_REPOSITORY and UPDATE_REPOSITORY is allowed to invalidate a distribution set
+     * Verify that a user that has authority READ_DISTRIBUTION_SET and UPDATE_DISTRIBUTION_SET is allowed to invalidate a distribution set
      */
     @Test
-    @WithUser(authorities = { "READ_REPOSITORY", "UPDATE_REPOSITORY" })
+    @WithUser(authorities = { "READ_DISTRIBUTION_SET", "UPDATE_DISTRIBUTION_SET" })
     void verifyInvalidateWithReadAndUpdateRepoAuthority() {
```
vasilchev, line 187: shouldn't UPDATE_DISTRIBUTION_SET enough ? (implying update_ -> read_)

hawkbit-security-core/src/main/java/org/eclipse/hawkbit/im/authentication/SpPermission.java
```diff
@@ -124,27 +127,6 @@ public final class SpPermission {
             TENANT_CONFIGURATION + IMPLY_UPDATE + TENANT_CONFIGURATION + "\n" +
```
vasilchev, line 127: shouldn't we have CREATE/READ/UPDATE/DELETE_DISTRIBUTION_SET + IMPLY_READ + DISTRIBUTION_SET_TYPE as well 
(just like we have for the CREATE/READ/UPDATE/DELETE_SOFTWARE_MODULE + IMPLY_READ + SOFTWARE_MODULE_TYPE)?

vasilchev: (just an idea, not sure if better..)
Also maybe we could introduce DOWNLOAD_REPOSITORY_ARTIFACT + '>' SOFTWARE_MODULE_DOWNLOAD_ARTIFACT and remove DOWNLOAD_REPOSITORY_ARTIFACT from PreAuthorize methods

hawkbit-mgmt/hawkbit-mgmt-resource/src/main/java/org/eclipse/hawkbit/mgmt/rest/resource/mapper/MgmtDistributionSetMapper.java
vasilchev: yes, indeed, missed that

hawkbit-repository/hawkbit-repository-jpa/src/test/java/org/eclipse/hawkbit/repository/jpa/management/DistributionSetInvalidationManagementTest.java
avgustinmm: Indeed I would expect this method to require only update. BUT, the implementation requires READ permission (because it gets distributions set before updating it) which is not implied by the hawkbit at the moment.
It's up to the operators to apply READ permissions together with any CREATE/UPDATE/DELETEE. 
This could be changed by implication in hierarchy (could be made via properties now) or we could, in future:
1. Add out of the box CREATE/UPDATE/DELETE to imply READ
2. or, carefully fix methods, as invalidate DS, that after checking the methods @PreAuthorizie to make all calls to other services (e.g. get something) in doProtected style - this however could be error prone - e.g. forget permissions to request.

At least for now - this keeps the older behavior. In future, we could decide to change something

hawkbit-security-core/src/main/java/org/eclipse/hawkbit/im/authentication/SpPermission.java
avgustinmm: DISTRIBUTION_SET_HIERARCHY does that (at least in the master)
to add DOWNLOAD_REPOSITORY_ARTIFACT + IMPLY + SOFTWARE_MODULE_DOWNLOAD and remove it from "required" permissions is good idea - will do this with a followup

PR #2680, opened 2025-09-19
Add actions contxt management

avgustinmm:
* based on TARGET permissions - action is accessible iff it its target is accessible
* add some more (id) serch fields
* (backward incompatible) software module search supports now search by name (type.name) and the old search by type (indead by type key) is now with type.key


avgustinmm:
TODO: add support for deeper, e.g. target.type.id, jpa search keys

PR #2684, opened 2025-09-19
Refacto oauth2 granted authorities 

flobz:
I notice that if oauth2 provider don't provide a refresh token then hawkbit-ui will end up with a 401 error when access token expired instead of redirect to login page.
This refacto fix this issue by removing hacky class OAuth2TokenManager and use a filter to add grandted authorities to auth context.


hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/SimpleUIApp.java
```diff
@@ -21,52 +21,42 @@
...
 @Import(FeignClientsConfiguration.class)
 public class SimpleUIApp implements AppShellConfigurator {
 
     private static final String AUTHORIZATION_HEADER = "Authorization";
-    private static final Function<OAuth2TokenManager, RequestInterceptor> AUTHORIZATION = oAuth2TokenManager -> requestTemplate -> {
-        final Authentication authentication = SecurityContextHolder.getContext().getAuthentication();
-        if (authentication instanceof OAuth2AuthenticationToken authenticationToken) {
-            String bearerToken = oAuth2TokenManager.getToken(authenticationToken);
-            requestTemplate.header(AUTHORIZATION_HEADER, "Bearer " + bearerToken);
+    private static final RequestInterceptor AUTHORIZATION = (requestTemplate) -> {
```
avgustinmm: * (requestTemplate) -> brackets are not needed
* var - in most of the cases in hawkbit we explicitly state the type. It's not necessarily but in the sake of common approach maybe "var authentication" could be replaced with "final Authentication authentication" as it was before (plese use final when possible)
* there is AUTHORIZATION_HEADER constant introduced - please use it instead of "Authorization"

flobz: done

hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/security/SecurityConfiguration.java
```diff
@@ -24,19 +25,27 @@
...
+@Slf4j
 @Import(VaadinAwareSecurityContextHolderStrategyConfiguration.class)
 public class SecurityConfiguration {
 
     private Customizer<OAuth2LoginConfigurer<HttpSecurity>> oAuth2LoginConfigurerCustomizer;
+    @Autowired
+    private UserDetailsSetter userDetailsSetter;
+
+    @Autowired
+    private InMemoryClientRegistrationRepository clientRegistrationRepository;
```
desislava-marinova, line 48: This should be declared with `required=false`, as it will not be initialised when OAuth is not used

hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/security/UserService.java
```diff
@@ -0,0 +1,72 @@
...
+
+    private SecurityContextHolderStrategy securityContextHolderStrategy;
+    private HawkbitMgmtClient hawkbitClient;
+
+    @Override
+    @Cacheable(cacheNames = "userDetails")
+    public UserDetails loadUserByUsername(String username) throws UsernameNotFoundException {
+        Objects.requireNonNull(username);
+        var authentication = securityContextHolderStrategy.getContext().getAuthentication();
+        return new User(authentication.getName(), authentication.getCredentials().toString(), getGrantedAuthorities());
```
desislava-marinova: When using Basic Auth, the `authentication` at that point is `null`

hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/security/SecurityConfiguration.java
flobz: done

hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/security/UserService.java
flobz: I did a refacto to handle basic auth it should be fixed.

hawkbit-simple-ui/src/main/java/org/eclipse/hawkbit/ui/simple/security/GrantedAuthoritiesService.java
```diff
@@ -0,0 +1,58 @@
...
+import org.springframework.stereotype.Service;
+
+@Service
+@AllArgsConstructor
+@Slf4j
+public class GrantedAuthoritiesService {
+
+    private HawkbitMgmtClient hawkbitClient;
+
+    @Cacheable(cacheNames = "userDetails")
```
desislava-marinova: Consider clearing the cache after logout to be sure the view will be up-to-date on the next login.

flobz: done

PR #2686, opened 2025-09-23
First level suppor for RsqlQueryFields shortcut support

avgustinmm:
* now it is possible to have a showrtcut for a sub attributes (i.e. calling it directly with enum name, e.g. type -> type.key) with directly specifying the defaultSubEntityAttribute
* no need to have single sub attribute in order to have a default sub attribute
* added TYPE search field for TargetFields (sinonim of targettype)
* targettype is deprecated - to be decided if and when to be removed
* returned back "type" direct search (with meaning type.key) for DistributionSet and SoftwareModule as non-depricated
* add serche with "type" as type.key for Target

### DIFF
--- a/hawkbit-core/src/main/java/org/eclipse/hawkbit/repository/RsqlQueryField.java
+++ b/hawkbit-core/src/main/java/org/eclipse/hawkbit/repository/RsqlQueryField.java
@@ -40,11 +40,21 @@ public interface RsqlQueryField {
     }
 
     /**
+     * Return the default sub entity attribute if available. This allows to skip that sub-attribute and call with "shortcut" - the enum name
+     *
+     * @return the default sub-attribute or <code>null</code>> if no default is available
+     */
+    default String getDefaultSubEntityAttribute() {
+        final List<String> subAttributes = getSubEntityAttributes();
+        return subAttributes.size() == 1 ? subAttributes.get(0) : null;
+    }
+
+    /**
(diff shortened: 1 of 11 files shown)


hawkbit-core/src/main/java/org/eclipse/hawkbit/repository/RsqlQueryField.java
```diff
@@ -39,12 +39,22 @@ default List<String> getSubEntityAttributes() {
...
     }
 
+    /**
+     * Return the default sub entity attribute if available. This allows to skip that sub-attribute and call with "shortcut" - the enum name
+     *
+     * @return the default sub-attribute or <code>null</code>> if no default is available
+     */
+    default String getDefaultSubEntityAttribute() {
+        final List<String> subAttributes = getSubEntityAttributes();
+        return subAttributes.size() == 1 ? subAttributes.get(0) : null;
```
strailov, line 49: Isn't it possible to have more than 1 subAttribute here ? 
In that case we want to return null ?

avgustinmm: Yes, by default, if you have more then one attribute - you don't have default (as it was before). If you have more then one sub attributes and you want to have default - you should specify it explicitly as it is done in DistributionSetFields and SoftwareModuleFields

PR #2707, opened 2025-09-29
Cleanup QL parse and tranform API

avgustinmm:


hawkbit-repository/hawkbit-repository-jpa-ql/src/main/java/org/eclipse/hawkbit/repository/jpa/ql/QLSupport.java
```diff
@@ -126,6 +130,15 @@ void setQueryParser(final QueryParser parser) {
         this.parser = parser;
     }
 
+    @Override
+    public void onApplicationEvent(@NonNull final ContextRefreshedEvent event) {
+        nodeTransformers = event.getApplicationContext().getBeansOfType(NodeTransformer.class).values().stream().sorted((b1, b2) -> {
+            final Order o1 = b1.getClass().getAnnotation(Order.class);
+            final Order o2 = b2.getClass().getAnnotation(Order.class);
```
vasilchev, line 137: we are always sure that there will be exactly 2 beans registered NodeTransformer?

avgustinmm: no, it is a comparator

PR #2709, opened 2025-09-30
Remove deprecated DOWNLOAD_REPOSITORY_ARTIFACT permission

avgustinmm:
use READ_SOFTWARE_MODULE_DOWNLOAD


hawkbit-security-core/src/main/java/org/eclipse/hawkbit/im/authentication/SpPermission.java
```diff
@@ -161,9 +153,6 @@ public final class SpPermission {
         // system permission, (!) take care with
         allPermissions.add(SYSTEM_ADMIN);
 
-        // add deprecated permissions
-        allPermissions.add(DOWNLOAD_REPOSITORY_ARTIFACT);
```
vasilchev, line 165: what is the plan here ? there are users that still have this permission.

1. wouldn't this fail if they try to modify their existing permissions ? e.g. you want to add/change some permission on a user that has this DOWNLOAD_REPOSITORY_ARTIFACT - if user only append some permissino to the already existing (this one will remain in the list) - wouldnt we fail on some check that this permission is not supported anymore ?

2. Also i cannot see anything implying this now, so meaning only if you have it explicitly only then you will be able to download artifacts (or REPOSITORY_ADMIN)

3. If users alreadyh have the old SOFTWARE_MODULE_DOWNLOAD permission - they are suppose to migrate db or ?

avgustinmm: in hawkbit users specify explicitly permissions. 
with removal of `repository` permissions the users shall replace them with more fine grained for ds, ds types, sm and sm types.
the same way DOWNLOAD_REPOSITORY_ARTIFACT shall be replaced with READ_SOFTWARE_MODULE_DOWNLOAD

Issue #2713, opened 2025-10-01
Docs: Fix documentation around target token limitations


jnodorp-jaconi:
The documentation states this about target tokens:

https://github.com/eclipse-hawkbit/hawkbit/blob/37f2ceb8efda355c1cfdc6b9ee4337771846b4d7/site/content/concepts/authentication.md?plain=1#L25

This is wrong.

Target tokens can have [up to 128 characters](https://github.com/eclipse-hawkbit/hawkbit/blob/0.9.0/hawkbit-repository/hawkbit-repository-api/src/main/java/org/eclipse/hawkbit/repository/model/Target.java#L33). I tested this and could not find any limitation when using a 128 character target token.

Also there is no limitation to alphanumeric characters. I tested base64 encoded tokens (containing `=`, `-`, and `/`) and could not find any issues.

Given that the limitations seem arbitrary, I'd suggest updating the docs instead of adjusting the code.

denislavprinov:
Hi @jnodorp-jaconi, 

Thank you for bringing this to our attention!
We are currently in the process of **migrating** the documentation **to GitHub Pages**, and this update will be reflected in the new documentation.

PR #2714, opened 2025-10-01
Clarify target token docs (#2713)

jnodorp-jaconi:


avgustinmm:
It was a little bit hard to me to understand the statement. Maybe it shall be something like "This token could be provided (with size up to 128) at creation. If not specified - it would be auto generated with size 32 characters."
Also, if we do this doc so detailed it's worth to mention that the security token could also be modified via Management API - PUT /rest/v1/targets/{targetId}

jnodorp-jaconi:
@denislavprinov said on #2713, that the issue will be addressed during the migration to GitHub pages. Closing.

PR #2725, opened 2025-10-07
Fix claims NPE

flobz:
Fix error:
```
java.lang.NullPointerException: Cannot invoke "Object.getClass()" because "current" is null
	at org.eclipse.hawkbit.autoconfigure.mgmt.MgmtSecurityConfiguration$DefaultOAuth2ResourceServerCustomizer.followPathInJwtClaims(MgmtSecurityConfiguration.java:207) ~[hawkbit-mgmt-starter-0.9.0.jar:0.9.0]
```

### DIFF
--- a/hawkbit-mgmt/hawkbit-mgmt-starter/src/main/java/org/eclipse/hawkbit/autoconfigure/mgmt/MgmtSecurityConfiguration.java
+++ b/hawkbit-mgmt/hawkbit-mgmt-starter/src/main/java/org/eclipse/hawkbit/autoconfigure/mgmt/MgmtSecurityConfiguration.java
@@ -61,7 +61,7 @@ import org.springframework.security.web.session.SessionManagementFilter;
  */
 @Slf4j
 @Configuration
-@EnableConfigurationProperties({HawkbitSecurityProperties.class, OidcProperties.class})
+@EnableConfigurationProperties({ HawkbitSecurityProperties.class, OidcProperties.class })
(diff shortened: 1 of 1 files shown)


hawkbit-mgmt/hawkbit-mgmt-starter/src/main/java/org/eclipse/hawkbit/autoconfigure/mgmt/MgmtSecurityConfiguration.java
```diff
@@ -192,11 +189,12 @@ public void customize(OAuth2ResourceServerConfigurer<HttpSecurity> oauth2Resourc
...
             final String[] chunks = path.split("\\.");
             Object current = jwt.getClaims();
+            if (current == null) {
+                return null;
+            }
             for (final String chunk : chunks) {
                 if (current instanceof Map<?, ?> map) {
                     current = map.get(chunk);
-                } else if (current == null) {
-                    return null;
```
avgustinmm, line 199: Removing this check may lead to misleading warn log - "unexpected claim type" instead of null ("not found")

flobz: Fixed !

avgustinmm (approved):

PR #2728, opened 2025-10-08
Action history cleanup/purge initial

strailov:
The following PR introduces a functionality to clean/purge action history whenever needed by couple of ways:

### Action resource

- Delete action by id using newly introduced **_DELETE /rest/v1/actions/{actionId}_**
- Delete multiple actions by IDs OR RSQL filter using **_DELETE /rest/v1/actions_**

### Target resource

- Delete/Purge target actions EXCEPT the latest <N> actions using **_DELETE /rest/v1/targets/{targetId}/actions_**?numberOfActions=N
- Delete target actions by provided action id list using again **_DELETE /rest/v1/targets/{targetId}/actions_** with the list into the request body.

### Action purge/cleanup functionality on max actions per target quota hit

A configuration (in Tenant configuration) option which provides a way to configure a percentage of target actions to be deleted when hitting max actions per target quota. Default value is 0% which is the current behaviour. If changed however - will purge the oldest actions matching the percentage and won't fail on action creation.
For example if 2000 is the action quota limit and we have configured 10% actions cleanup on quota hit - it will delete the 200 oldest actions and create a new one.

### DIFF
--- a/hawkbit-repository/hawkbit-repository-api/src/main/java/org/eclipse/hawkbit/repository/DeploymentManagement.java
+++ b/hawkbit-repository/hawkbit-repository-api/src/main/java/org/eclipse/hawkbit/repository/DeploymentManagement.java
@@ -413,5 +413,5 @@ public interface DeploymentManagement extends PermissionSupport {
     void cancelActionsForDistributionSet(final ActionCancellationType cancelationType, final DistributionSet set);
 
     @PreAuthorize(SpringEvalExpressions.IS_SYSTEM_CODE)
-    void handleMaxAssignmentsExceeded(Long targetId, Integer requested, AssignmentQuotaExceededException ex);
+    void handleMaxAssignmentsExceeded(Long targetId, Long requested, AssignmentQuotaExceededException ex);
 }
--- a/hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/AbstractDsAssignmentStrategy.java
+++ b/hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/AbstractDsAssignmentStrategy.java
@@ -14,8 +14,6 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
 import java.util.Set;
-import java.util.function.BiConsumer;
-import java.util.function.BiFunction;
 import java.util.function.BooleanSupplier;
 
 import jakarta.persistence.criteria.JoinType;
@@ -69,7 +67,7 @@ public abstract class AbstractDsAssignmentStrategy {
     private final BooleanSupplier multiAssignmentsConfig;
     private final BooleanSupplier confirmationFlowConfig;
     private final RepositoryProperties repositoryProperties;
-    private final TriConsumer<Long, Integer, AssignmentQuotaExceededException> maxAssignmentExeededHandler;
+    private final TriConsumer<Long, Long, AssignmentQuotaExceededException> maxAssignmentExceededHandler;
 
     @SuppressWarnings("java:S107")
     AbstractDsAssignmentStrategy(
@@ -78,7 +76,7 @@ public abstract class AbstractDsAssignmentStrategy {
             final ActionRepository actionRepository, final ActionStatusRepository actionStatusRepository,
             final QuotaManagement quotaManagement, final BooleanSupplier multiAssignmentsConfig,
             final BooleanSupplier confirmationFlowConfig, final RepositoryProperties repositoryProperties,
-            final TriConsumer<Long, Integer, AssignmentQuotaExceededException> maxAssignmentExeededHandler) {
+            final TriConsumer<Long, Long, AssignmentQuotaExceededException> maxAssignmentExceededHandler) {
         this.targetRepository = targetRepository;
         this.afterCommit = afterCommit;
         this.actionRepository = actionRepository;
@@ -87,7 +85,7 @@ public abstract class AbstractDsAssignmentStrategy {
         this.multiAssignmentsConfig = multiAssignmentsConfig;
         this.confirmationFlowConfig = confirmationFlowConfig;
         this.repositoryProperties = repositoryProperties;
-        this.maxAssignmentExeededHandler = maxAssignmentExeededHandler;
+        this.maxAssignmentExceededHandler = maxAssignmentExceededHandler;
     }
 
     public JpaAction createTargetAction(
@@ -299,12 +297,12 @@ public abstract class AbstractDsAssignmentStrategy {
                 .publishEvent(new CancelTargetAssignmentEvent(tenant, actions)));
     }
 
-    private void assertActionsPerTargetQuota(final Target target, final int requested) {
+    private void assertActionsPerTargetQuota(final Target target, final long requested) {
         final int quota = quotaManagement.getMaxActionsPerTarget();
         try {
             QuotaHelper.assertAssignmentQuota(target.getId(), requested, quota, Action.class, Target.class, actionRepository::countByTargetId);
         } catch (AssignmentQuotaExceededException ex) {
-            maxAssignmentExeededHandler.accept(target.getId(), requested, ex);
+            maxAssignmentExceededHandler.accept(target.getId(), requested, ex);
         }
     }
 }
--- a/hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaDeploymentManagement.java
+++ b/hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaDeploymentManagement.java
@@ -37,6 +37,7 @@ import jakarta.persistence.criteria.Root;
 
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.collections4.ListUtils;
+import org.apache.commons.lang3.ObjectUtils;
 import org.apache.commons.lang3.function.TriConsumer;
 import org.eclipse.hawkbit.repository.ActionFields;
 import org.eclipse.hawkbit.repository.DeploymentManagement;
@@ -59,16 +60,15 @@ import org.eclipse.hawkbit.repository.jpa.acm.AccessController;
 import org.eclipse.hawkbit.repository.jpa.configuration.Constants;
 import org.eclipse.hawkbit.repository.jpa.executor.AfterTransactionCommitExecutor;
 import org.eclipse.hawkbit.repository.jpa.model.AbstractJpaBaseEntity_;
-import org.eclipse.hawkbit.repository.jpa.model.AbstractJpaNamedEntity_;
 import org.eclipse.hawkbit.repository.jpa.model.JpaAction;
 import org.eclipse.hawkbit.repository.jpa.model.JpaActionStatus;
 import org.eclipse.hawkbit.repository.jpa.model.JpaDistributionSet;
 import org.eclipse.hawkbit.repository.jpa.model.JpaTarget;
 import org.eclipse.hawkbit.repository.jpa.model.JpaTarget_;
+import org.eclipse.hawkbit.repository.jpa.ql.QLSupport;
 import org.eclipse.hawkbit.repository.jpa.repository.ActionRepository;
 import org.eclipse.hawkbit.repository.jpa.repository.ActionStatusRepository;
 import org.eclipse.hawkbit.repository.jpa.repository.TargetRepository;
-import org.eclipse.hawkbit.repository.jpa.ql.QLSupport;
 import org.eclipse.hawkbit.repository.jpa.specifications.ActionSpecifications;
 import org.eclipse.hawkbit.repository.jpa.specifications.TargetSpecifications;
 import org.eclipse.hawkbit.repository.jpa.utils.DeploymentHelper;
@@ -163,9 +163,11 @@ public class JpaDeploymentManagement extends JpaActionManagement implements Depl
         this.auditorProvider = auditorProvider;
         this.txManager = txManager;
         onlineDsAssignmentStrategy = new OnlineDsAssignmentStrategy(targetRepository, afterCommit, actionRepository, actionStatusRepository,
-                quotaManagement, this::isMultiAssignmentsEnabled, this::isConfirmationFlowEnabled, repositoryProperties, maxAssignmentsExceededHandler);
+                quotaManagement, this::isMultiAssignmentsEnabled, this::isConfirmationFlowEnabled, repositoryProperties,
+                maxAssignmentsExceededHandler);
         offlineDsAssignmentStrategy = new OfflineDsAssignmentStrategy(targetRepository, afterCommit, actionRepository, actionStatusRepository,
-                quotaManagement, this::isMultiAssignmentsEnabled, this::isConfirmationFlowEnabled, repositoryProperties, maxAssignmentsExceededHandler);
+                quotaManagement, this::isMultiAssignmentsEnabled, this::isConfirmationFlowEnabled, repositoryProperties,
+                maxAssignmentsExceededHandler);
         this.tenantConfigurationManagement = tenantConfigurationManagement;
         this.systemSecurityContext = systemSecurityContext;
         this.tenantAware = tenantAware;
@@ -407,9 +409,14 @@ public class JpaDeploymentManagement extends JpaActionManagement implements Depl
     @Transactional
     public void deleteAction(final long actionId) {
         log.info("Deleting action {}", actionId);
-        actionRepository.getAccessController().ifPresent(accessController ->
-                actionRepository.findAll(accessController.appendAccessRules(AccessController.Operation.UPDATE,
-                        ((root, q, cb) -> cb.equal(root.get(AbstractJpaNamedEntity_.id), actionId)))));
+        actionRepository.getAccessController().ifPresent(accessController -> {
+            // check update access
+            if (ObjectUtils.isEmpty(actionRepository.findAll(accessController.appendAccessRules(
+                    AccessController.Operation.UPDATE, ((root, q, cb) -> cb.equal(root.get(AbstractJpaBaseEntity_.id), actionId)))))) {
+                // could be also InsufficientPermissionException but for security reasons we do not reveal that the entity exists
+                throw new EntityNotFoundException(Action.class, actionId);
+            }
+        });
         actionRepository.deleteById(actionId);
     }
 
@@ -418,9 +425,10 @@ public class JpaDeploymentManagement extends JpaActionManagement implements Depl
     public void deleteActionsByRsql(final String rsql) {
         log.info("Deleting actions matching rsql {}", rsql);
         final Specification<JpaAction> specification = QLSupport.getInstance().buildSpec(rsql, ActionFields.class);
-        actionRepository.getAccessController().ifPresent(accessController ->
-                actionRepository.findAll(accessController.appendAccessRules(AccessController.Operation.UPDATE, specification)));
-        actionRepository.delete(specification);
+        actionRepository.delete(
+                actionRepository.getAccessController()
+                        .map(accessController -> accessController.appendAccessRules(AccessController.Operation.UPDATE, specification))
+                        .orElse(specification));
     }
 
     @Override
@@ -428,7 +436,8 @@ public class JpaDeploymentManagement extends JpaActionManagement implements Depl
     public void deleteActionsByIds(final List<Long> actionIds) {
         log.info("Deleting actions with ids {}", actionIds);
         actionRepository.getAccessController().ifPresent(accessController ->
-                actionRepository.findAll(accessController.appendAccessRules(AccessController.Operation.UPDATE, ActionSpecifications.byIdIn(actionIds))));
+                actionRepository.findAll(
+                        accessController.appendAccessRules(AccessController.Operation.UPDATE, ActionSpecifications.byIdIn(actionIds))));
         actionRepository.deleteByIdIn(actionIds);
     }
 
@@ -498,7 +507,7 @@ public class JpaDeploymentManagement extends JpaActionManagement implements Depl
                 startScheduledActions(groupScheduledActions.getContent());
                 return groupScheduledActions.getTotalElements();
             }
-        }) > 0);
+        }) > 0) ;
     }
 
     @Override
@@ -590,11 +599,11 @@ public class JpaDeploymentManagement extends JpaActionManagement implements Depl
     }
 
     @Override
-    public void handleMaxAssignmentsExceeded(Long targetId, Integer requested, AssignmentQuotaExceededException ex) {
+    public void handleMaxAssignmentsExceeded(final Long targetId, final Long requested, final AssignmentQuotaExceededException ex) {
         maxAssignmentsExceededHandler.accept(targetId, requested, ex);
     }
 
-    private final TriConsumer<Long, Integer, AssignmentQuotaExceededException> maxAssignmentsExceededHandler = (targetId, requested,quotaExceededException) -> {
+    private final TriConsumer<Long, Long, AssignmentQuotaExceededException> maxAssignmentsExceededHandler = (targetId, requested, quotaExceededException) -> {
         int actionsPurgePercentage = getActionsPurgePercentage();
         int quota = quotaManagement.getMaxActionsPerTarget();
         if (actionsPurgePercentage > 0 && actionsPurgePercentage < 100) {
@@ -617,6 +626,7 @@ public class JpaDeploymentManagement extends JpaActionManagement implements Depl
 
     /**
      * Deletes the first n target actions of a target
+     *
      * @param targetId - target id
      * @param oldestToDelete - number of oldest actions to be deleted
      */
@@ -630,7 +640,7 @@ public class JpaDeploymentManagement extends JpaActionManagement implements Depl
                 " ORDER BY id ASC" +
                 " LIMIT " + oldestToDelete
                 + ") AS sub"
-                +")";
+                + ")";
         Query query = entityManager.createNativeQuery(SQL);
         query.setParameter("target", targetId);
         query.executeUpdate();
@@ -894,22 +904,19 @@ public class JpaDeploymentManagement extends JpaActionManagement implements Depl
         countOfTargetInRequest.forEach(this::checkMaxAssignmentQuota);
     }
 
-    private void checkMaxAssignmentQuota(final String controllerId, long requested) {
+    private void checkMaxAssignmentQuota(final String controllerId, final long requested) {
         final int quota = quotaManagement.getMaxActionsPerTarget();
-        systemSecurityContext.runAsSystem(() -> {
-            final Optional<JpaTarget> opt = targetRepository.findByControllerId(controllerId);
-            if (opt.isPresent()) {
-                final Target target = opt.get();
-                try {
-                    QuotaHelper.assertAssignmentQuota(controllerId, requested, quota, Action.class, Target.class,
-                            actionRepository::countByTargetControllerId);
-                } catch (final AssignmentQuotaExceededException ex) {
+        try {
+            systemSecurityContext.runAsSystem(() -> QuotaHelper.assertAssignmentQuota(
+                    controllerId, requested, quota, Action.class, Target.class, actionRepository::countByTargetControllerId));
+        } catch (final AssignmentQuotaExceededException ex) {
+            targetRepository.findByControllerId(controllerId).ifPresentOrElse(
                     // assume requested are always smaller than int size
-                    maxAssignmentsExceededHandler.accept(target.getId(), Math.toIntExact(requested), ex);
-                }
-            }
-            return null;
-        });
+                    target -> maxAssignmentsExceededHandler.accept(target.getId(), requested, ex),
+                    () -> {
+                        throw new EntityNotFoundException(Target.class, controllerId);
+                    });
+        }
     }
 
     private void closeOrCancelActiveActions(final AbstractDsAssignmentStrategy assignmentStrategy, final List<List<Long>> targetIdsChunks) {
--- a/hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaRolloutExecutor.java
+++ b/hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/JpaRolloutExecutor.java
@@ -858,7 +858,7 @@ public class JpaRolloutExecutor implements RolloutExecutor {
      * @param target the target
      * @param requested number of actions to check
      */
-    private void assertActionsPerTargetQuota(final Target target, final int requested) {
+    private void assertActionsPerTargetQuota(final Target target, final long requested) {
         final int quota = quotaManagement.getMaxActionsPerTarget();
         try {
             QuotaHelper.assertAssignmentQuota(target.getId(), requested, quota, Action.class, Target.class, actionRepository::countByTargetId);
--- a/hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/OfflineDsAssignmentStrategy.java
+++ b/hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/OfflineDsAssignmentStrategy.java
@@ -54,9 +54,9 @@ class OfflineDsAssignmentStrategy extends AbstractDsAssignmentStrategy {
             final ActionRepository actionRepository, final ActionStatusRepository actionStatusRepository,
             final QuotaManagement quotaManagement, final BooleanSupplier multiAssignmentsConfig,
             final BooleanSupplier confirmationFlowConfig, final RepositoryProperties repositoryProperties,
-            final TriConsumer<Long, Integer, AssignmentQuotaExceededException> maxAssignmentExeededHandler) {
+            final TriConsumer<Long, Long, AssignmentQuotaExceededException> maxAssignmentExceededHandler) {
         super(targetRepository, afterCommit, actionRepository, actionStatusRepository,
-                quotaManagement, multiAssignmentsConfig, confirmationFlowConfig, repositoryProperties, maxAssignmentExeededHandler);
+                quotaManagement, multiAssignmentsConfig, confirmationFlowConfig, repositoryProperties, maxAssignmentExceededHandler);
     }
 
     @Override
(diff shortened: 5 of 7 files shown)


hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/actions/TargetAssignmentsChecker.java
```diff
@@ -0,0 +1,82 @@
...
+import org.eclipse.hawkbit.repository.model.Action;
+import org.eclipse.hawkbit.repository.model.Target;
+import org.eclipse.hawkbit.security.SystemSecurityContext;
+import org.eclipse.hawkbit.tenancy.configuration.TenantConfigurationProperties;
+
+import java.io.Serializable;
+import java.util.function.ToLongFunction;
+
+@Slf4j
+public class TargetAssignmentsChecker {
```
avgustinmm: Why dedicated class, package?
Why not in QuotaHelper?
I see here some beans are needed and doesn't fit in the current QuotaHelper static pattern - but you could use the singleton pattern, as for instance MdcHelper. This will also remove need to extend some constructors with additional arguments.

Another option would be to keep quota helper fully static, but callers to pass also
"quota hit handler" e.g. Consumer<AssignmentQuotaExceededException>
which will recover from AssignmentQuotaExceededException (with cleanup) or will re-throw exception.
Then having an deploymentManagement.cleanup (or whatever) could be passed as handler at all placess - deploymentManagement::cleanup

hawkbit-repository/hawkbit-repository-core/src/main/resources/hawkbit-repository-defaults.properties
```diff
@@ -94,4 +94,8 @@ hawkbit.server.tenant.configuration.implicit-lock-enabled.keyName=implicit.lock.
 hawkbit.server.tenant.configuration.implicit-lock-enabled.defaultValue=true
 hawkbit.server.tenant.configuration.implicit-lock-enabled.dataType=java.lang.Boolean
 
+hawkbit.server.tenant.configuration.percentage-of-actions-cleaned-on-quota-hit.keyName=percentage.of.actions.cleaned.on.quota.hit
```
avgustinmm: naming is not compatible with the other action properties as action.cleanup.actionExpiry
proposal: actions-cleanup-on-quota-hit-percent.keyName=actions.cleanup.onQuotaHit.percent

hawkbit-repository/hawkbit-repository-api/src/main/java/org/eclipse/hawkbit/repository/DeploymentManagement.java
```diff
@@ -298,6 +298,48 @@ List<DistributionSetAssignmentResult> assignDistributionSets(
     @PreAuthorize(SpringEvalExpressions.HAS_UPDATE_REPOSITORY)
     Action forceTargetAction(long actionId);
 
+    /**
+     * Deletes the current action by id.
+     * 
+     * @param actionId - action id
+     */
+    @PreAuthorize(SpringEvalExpressions.HAS_DELETE_REPOSITORY)
```
avgustinmm: I'm not sure regarding the permissions you need in order to be allowed to clean actions. Actually actions permissions are managed by TARGET permission and you "modify" target. So, wouldn't HAS_UPDATE_REPOSITORY (i.e. UPDATE_TARGET) be the needed permission?

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaDeploymentManagement.java
```diff
@@ -121,15 +122,21 @@ public class JpaDeploymentManagement extends JpaActionManagement implements Depl
...
     private static final int ACTION_PAGE_LIMIT = 1000;
-    private static final String QUERY_DELETE_ACTIONS_BY_STATE_AND_LAST_MODIFIED_DEFAULT = "DELETE FROM sp_action " + "WHERE tenant=" + Jpa.nativeQueryParamPrefix() + "tenant" + " AND status IN (%s)" + " AND last_modified_at<" + Jpa.nativeQueryParamPrefix() + "last_modified_at LIMIT " + ACTION_PAGE_LIMIT;
+    private static final String QUERY_DELETE_ACTIONS_BY_STATE_AND_LAST_MODIFIED_DEFAULT = "DELETE FROM sp_action " + "WHERE tenant=" + Jpa
+            .nativeQueryParamPrefix() + "tenant" + " AND status IN (%s)" + " AND last_modified_at<" + Jpa
+                    .nativeQueryParamPrefix() + "last_modified_at LIMIT " + ACTION_PAGE_LIMIT;
     private static final EnumMap<Database, String> QUERY_DELETE_ACTIONS_BY_STATE_AND_LAST_MODIFIED;
 
     static {
         QUERY_DELETE_ACTIONS_BY_STATE_AND_LAST_MODIFIED = new EnumMap<>(Database.class);
         QUERY_DELETE_ACTIONS_BY_STATE_AND_LAST_MODIFIED.put(Database.SQL_SERVER,
```
avgustinmm: we don't support SQL_SERVER anymore. It's not new in that PR, but could be used to be removed.

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaDeploymentManagement.java
```diff
@@ -401,6 +413,45 @@ public Action forceTargetAction(final long actionId) {
...
+
+    @Override
+    public void deleteTargetActionsByIds(final String target, final List<Long> actionsIds) {
+        log.info("Delete actions for target {} with action ids {}", target, actionsIds);
+        actionRepository.delete(ActionSpecifications.byControllerIdAndIdIn(target, actionsIds));
+    }
+
+    @Override
+    @Transactional
+    public void deleteTargetActions(final String target, final int numberOfActions) {
```
avgustinmm: AFAIU this method is actually "deletes count of  OLDEST action of a STATE to delete".  It's not very intuitive by the name.
So, I propose to rename it (at places) the method to someting like deleteLastTarget(Finished)Actions, with comment "of state ...". 
and maybe numberOfActions parameter to  simply number/count

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaDeploymentManagement.java
```diff
@@ -401,6 +413,45 @@ public Action forceTargetAction(final long actionId) {
...
+        final JpaTarget jpaTarget = targetRepository.findByControllerId(target)
```
avgustinmm, line 457: do we really wan't to check if there are >= actions?
If there are 10 actions and someone want to delete 100 - let delete them?

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/actions/TargetAssignmentsChecker.java
```diff
@@ -0,0 +1,82 @@
...
+
+    public TargetAssignmentsChecker(final EntityManager entityManager,
+            final SystemSecurityContext systemSecurityContext, final TenantConfigurationManagement tenantConfigurationManagement) {
+        this.entityManager = entityManager;
+        this.systemSecurityContext = systemSecurityContext;
+        this.tenantConfigurationManagement = tenantConfigurationManagement;
+    }
+
+    public void checkActionsPerTargetQuota(final Long id, final long requested, final int quota, final ToLongFunction<Long> func) {
+        int actionsPurgePercentage = getActionsPurgePercentage();
```
avgustinmm: never get something that may not be needed. Especially when it could be heavy operation - potential DB call, and very unexpected to be needed - only if quota is hit.
actionsPurgePercentage must be initialized in catch only if quota is hit

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/actions/TargetAssignmentsChecker.java
```diff
@@ -0,0 +1,82 @@
...
+
+        try {
+            QuotaHelper.assertAssignmentQuota(id, requested, quota, Action.class, Target.class, func);
+        } catch (final AssignmentQuotaExceededException quotaExceededException) {
+            if (actionsPurgePercentage > 0 && actionsPurgePercentage < 100) {
```
avgustinmm: better move all quota hit handling in deployment management.
I.e. move init of actionsPurgePercentage handling, cleanup, and return true / false if cleaned

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/actions/TargetAssignmentsChecker.java
strailov: You wanna make QuotaHelper singleton or talking about a whole new Singleton helper class ?

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/actions/TargetAssignmentsChecker.java
strailov: Why in deployment management, this handling is also present in RolloutExecutor (when creating actions) 
We need the logic in one place.

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaDeploymentManagement.java
strailov: Actually currently implemeted this method is more like : 
"deletes oldest actions and LEAVE numberOfActions latest" - it is mentioned to the PR description and the javadoc
The idea behind it was more like of use case for "leave me 100 actions of this particular target", I guess both are reasonable ways. At least in the pre development phase it was mentioned to be done that way, but I guess the option with oldest is also fine.

avgustinmm:
And one general API note. Methods that get rsql xor action lists seems little bit awkward
1. At least they shall have very good doc
2. Together with checking that at least one of them is set, the implementation MUST check then only one is set. If both are set maybe user doesn't know what he wants (maybe to delete by both criteria?) - and may get unexpected results.

you could consider something like:

```java
@DeleteMapping(value = MgmtRestConstants.ACTION_V1_REQUEST_MAPPING)
ResponseEntity<Void> deleteActions(
        @RequestParam(value = MgmtRestConstants.REQUEST_PARAMETER_SEARCH, required = false, defaultValue = "") String rsqlParam);

@DeleteMapping(value = MgmtRestConstants.ACTION_V1_REQUEST_MAPPING consumes = { MediaTypes.HAL_JSON_VALUE, MediaType.APPLICATION_JSON_VALUE })
ResponseEntity<Void> deleteActions(
        @RequestBody(required = false) List<Long> actionIds);
```
This actually created 2 same delete endpoints that could be called with param or with json body.
The only problem is that it generates an openapi doc (or swagger?) with single DELETE method and mix both arguments.
But maybe this could be somehow fixed - or reported to spring org.springdoc or to swagger and be (someday fixed)

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaDeploymentManagement.java
avgustinmm: no, idea is fine, all is about making thins more intuitive. 
In that case it could be:
* deleteTargetActions -> deleteOldestTargetActions 
* numberOfActions -> keepLast/Newest(Count/Number)

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/actions/TargetAssignmentsChecker.java
avgustinmm: JpaRolloutsExecutor anyway already depends/uses deploymentManagement

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/actions/TargetAssignmentsChecker.java
avgustinmm: Make QuotaHelper singleton (in the mentioned pattern) or keep it static with passing "quita hit handler"
There are many options.
If you move quota handling in deploymentManagement you could also keep QuotaHelper as is and the callers could do something like:

try {
  check quota
} catch (assgnmentexception) {
  deploymentManagement.handle(assgnmentexception)  
}

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/actions/TargetAssignmentsChecker.java
strailov: And what about AbstractDsAssignmentStrategy.createTargetAction ? There is also quota hit check and this will lead to cycling dependency with deploymentManagement.

avgustinmm: Assignment strategies are created by deployment management. so it could pass them whatever handlers it wants.
Or, handle exceptions instead of the strategies.

strailov:
This approach (with splitting the deleteActions rest logic) at least at my side does not work 100% properly. 
I invoke the method with rsql parameter but It always makes a call to the one with List wether I call it with null body, or empty does not matter. So I would suggest in order to escape additional headaches to remain as it is .

hawkbit-mgmt/hawkbit-mgmt-api/src/main/java/org/eclipse/hawkbit/mgmt/rest/api/MgmtTargetRestApi.java
```diff
@@ -403,6 +403,42 @@ The sequence of the sort criteria (multiple can be used) defines the sort order
...
+                    content = @Content(mediaType = "application/json", schema = @Schema(hidden = true))),
+            @ApiResponse(responseCode = "429", description = "Too many requests. The server will refuse further attempts " +
+                    "and the client has to wait another second.",
+                    content = @Content(mediaType = "application/json", schema = @Schema(hidden = true)))
+    })
+    @DeleteMapping(value = MgmtRestConstants.TARGET_V1_REQUEST_MAPPING + "/{targetId}/actions",
+            produces = { MediaTypes.HAL_JSON_VALUE, MediaType.APPLICATION_JSON_VALUE })
+    ResponseEntity<Void> deleteActionsForTarget(
+            @PathVariable("targetId") String targetId,
+            @RequestParam(name = "numberOfActions", required = false, defaultValue = "0") int numberOfActions,
```
avgustinmm: Rename to keepLast. Ant then, having this semantic - won't be better default value to be -1?

hawkbit-mgmt/hawkbit-mgmt-api/src/main/java/org/eclipse/hawkbit/mgmt/rest/api/MgmtTargetRestApi.java
```diff
@@ -403,6 +403,42 @@ The sequence of the sort criteria (multiple can be used) defines the sort order
                     in the result.""")
             String sortParam);
 
+
+    /**
+     * Deletes all actions for the provided target EXCEPT the latest N actions OR by provided action IDs list.
```
avgustinmm: Could be interpreted as: EXCEPT (the latest N actions) OR (by provided action IDs list).
Maybe will be more safe to switch: "provided action IDs list OR EXCEPT the latest N actions"

hawkbit-mgmt/hawkbit-mgmt-resource/src/main/java/org/eclipse/hawkbit/mgmt/rest/resource/MgmtActionResource.java
```diff
@@ -71,6 +75,27 @@ public ResponseEntity<MgmtAction> getAction(final Long actionId) {
...
+    public ResponseEntity<Void> deleteAction(Long actionId) {
+        deploymentManagement.deleteAction(actionId);
+        return ResponseEntity.ok().build();
+    }
+
+    @Override
+    @AuditLog(entity = "Actions", type = AuditLog.Type.DELETE, description = "Delete Actions", logResponse = true)
+    public ResponseEntity<Void> deleteActions(String rsqlParam, List<Long> actionIds) {
+        if (!ObjectUtils.isEmpty(rsqlParam)) {
+            deploymentManagement.deleteActionsByRsql(rsqlParam);
```
avgustinmm, line 96: If !ObjectUtils.isEmpty(actionIds) throw exception.
If user has specified both - maybe he hasn't understood the method behavior - he may expect that both specified sets (or intersection) will be deleted

hawkbit-mgmt/hawkbit-mgmt-resource/src/main/java/org/eclipse/hawkbit/mgmt/rest/resource/MgmtTargetResource.java
```diff
@@ -226,6 +226,23 @@ public ResponseEntity<PagedList<MgmtAction>> getActionHistory(
...
+    @Override
+    @AuditLog(entity = "Target", type = AuditLog.Type.DELETE, description = "Delete Actions For Target")
+    public ResponseEntity<Void> deleteActionsForTarget(final String targetId, final int numberOfActions, final List<Long> actionIds) {
+
+        if (numberOfActions <= 0 && actionIds == null) {
+            return ResponseEntity.badRequest().build();
+        }
+
+        if (actionIds != null) {
+            deploymentManagement.deleteTargetActionsByIds(targetId, actionIds);
```
avgustinmm, line 242: If numberOfActions > 0 (keepLast) throw exception.
If user has specified both - maybe he hasn't understood the method behavior - he may expect that both specified sets (or intersection) will be deleted

hawkbit-repository/hawkbit-repository-api/src/main/java/org/eclipse/hawkbit/repository/DeploymentManagement.java
```diff
@@ -372,4 +411,7 @@ List<DistributionSetAssignmentResult> assignDistributionSets(
      */
     @PreAuthorize(SpringEvalExpressions.HAS_UPDATE_REPOSITORY)
     void cancelActionsForDistributionSet(final ActionCancellationType cancelationType, final DistributionSet set);
+
+    @PreAuthorize(SpringEvalExpressions.IS_SYSTEM_CODE)
+    void maxAssignmentsExceededHandle(Long targetId, AssignmentQuotaExceededException ex);
```
avgustinmm: better handleMaxAssignmentsExceeded - usually methods doing something starts with verb

hawkbit-repository/hawkbit-repository-core/src/main/resources/hawkbit-repository-defaults.properties
```diff
@@ -94,4 +94,8 @@ hawkbit.server.tenant.configuration.implicit-lock-enabled.keyName=implicit.lock.
 hawkbit.server.tenant.configuration.implicit-lock-enabled.defaultValue=true
 hawkbit.server.tenant.configuration.implicit-lock-enabled.dataType=java.lang.Boolean
 
+hawkbit.server.tenant.configuration.actions-cleanup-on-quota-hit-percent.keyName=actions.cleanup.onQuotaHit.percent
```
avgustinmm: maybe better move it up together with the other action-cleanup property, in order to be at one place - no need to be in order of introduction

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/AbstractDsAssignmentStrategy.java
```diff
@@ -65,14 +67,16 @@ public abstract class AbstractDsAssignmentStrategy {
...
 
     @SuppressWarnings("java:S107")
     AbstractDsAssignmentStrategy(
             final TargetRepository targetRepository,
             final AfterTransactionCommitExecutor afterCommit,
             final ActionRepository actionRepository, final ActionStatusRepository actionStatusRepository,
             final QuotaManagement quotaManagement, final BooleanSupplier multiAssignmentsConfig,
-            final BooleanSupplier confirmationFlowConfig, final RepositoryProperties repositoryProperties) {
+            final BooleanSupplier confirmationFlowConfig, final RepositoryProperties repositoryProperties,
+            final BiFunction<Long, AssignmentQuotaExceededException, Void> assignmentsExceededHandler) {
```
avgustinmm: since the result is not used it should be BiConsumer, 
and better stick to single var/argument naming. In current PR it is named:
maxAssignmentExeededHandler, assignmentExeededHandler, handler

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaDeploymentManagement.java
```diff
@@ -529,6 +584,50 @@ public void cancelActionsForDistributionSet(final ActionCancellationType cancela
         }
     }
 
+    @Override
+    public void maxAssignmentsExceededHandle(Long targetId, AssignmentQuotaExceededException ex) {
+        maxAssignmentsExceededHandler.apply(targetId, ex);
+    }
+
+    public BiFunction<Long, AssignmentQuotaExceededException, Void> maxAssignmentsExceededHandler = (targetId,  quotaExceededException) -> {
```
avgustinmm: make it private

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaDeploymentManagement.java
```diff
@@ -529,6 +584,50 @@ public void cancelActionsForDistributionSet(final ActionCancellationType cancela
...
+        int actionsPurgePercentage = getActionsPurgePercentage();
+        int quota = quotaManagement.getMaxActionsPerTarget();
+        if (actionsPurgePercentage > 0 && actionsPurgePercentage < 100) {
+            int numberOfActions = (int) ((actionsPurgePercentage / 100.0) * quota);
+            log.info("Actions purge percentage {}, will delete {} oldest actions for target {}",
+                    actionsPurgePercentage, numberOfActions, targetId);
+            deleteOldestTargetActions(targetId, numberOfActions);
```
avgustinmm, line 621: after deleting the oldest N actions we could be sure that requested actions wont break the limit.
E.g. if called via checkMaxAssignmentQuota checking if we could add 1000 assignments and we clean 100 - we could break the quota.
We could:
1. of called - by externals - rollouts and strategies always check for 1 assume that at least one is removed - however better fix the method to always request 1
2. if called by checkMaxAssignmentQuota the method check if remove sufficient number of actions

strailov: What if we integrate requested attribute into the handler and check there ? 
if requested is bigger than configured number of actions we just throw the quota exception ?

avgustinmm: Will become too complicated
we may pass requested if needed .. maybe it could be even in the exeception - "exeeded count" if the purger purge less - exeception will be thrown anyway

strailov: Why would become too complicated ? 
How would we check how many actions could break the quota otherwise ? We would need to extract the purge percentage earlier and calculate the number of actions to be deleted before invoking the handler

avgustinmm: I understood that you want to do exactly that. Maybe have misunderstood. Yes, the check could be in handleXXX
but for that it must know many are requested.
Another option would be handle to be changed to "purge" (called only on quota hit) and return how many are purged. then the callers will throw or continue depending of purged count and requested.
Not that purge shall be called only on quota hit.
Also if quota is 1000, percent is 10 we shall not delete 100 but keep 900 - i.e. if there are 950 - only last 50 shall be removed. I think it is more natural - we purge to  some level. Then won't be a problem even if by mistake someone call purge on, e.g. 100 actions

strailov: Added logic for that - also a test case to check that it keeps the actions that do not need to be purged.

hawkbit-mgmt/hawkbit-mgmt-resource/src/main/java/org/eclipse/hawkbit/mgmt/rest/resource/MgmtTargetResource.java
avgustinmm: If the default is -1, then 0 should be acceptable value - meaning keep 0 will delete all target actions.
so here in the checks (2 checks) shall have changes

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaDeploymentManagement.java
```diff
@@ -529,6 +583,57 @@ public void cancelActionsForDistributionSet(final ActionCancellationType cancela
...
+        }
+    };
+
+    /**
+     * Deletes the first n target actions of a target
+     * @param targetId - target id
+     * @param oldestToDelete - number of oldest actions to be deleted
+     */
+    public void deleteOldestTargetActions(long targetId, int oldestToDelete) {
+        // Workaround for the case where JPQL or Criteria API do not support LIMIT
```
avgustinmm, line 634: I think that purge should work a little bit different. 
If quota is 1000, percent is 10 we shall not delete 100 but keep 900 - i.e. if there are 950 - only last 50 shall be removed. I think it is more natural - we purge to  some level. Then won't be a problem even if by mistake someone call purge on, e.g. 100 actions
React

avgustinmm (changes requested):

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaDeploymentManagement.java
strailov: Actually this is what is implemented but in the quota exception handler.

strailov: Added a test case also for this use case.

hawkbit-mgmt/hawkbit-mgmt-resource/src/main/java/org/eclipse/hawkbit/mgmt/rest/resource/MgmtTargetResource.java
avgustinmm: if (keepLast > 0)  is not enough, if 0 you should also throw exception - 0 means delete all
also I startet to wonder - what if actionIds is empty list? should it be treated maybe as null? so maybe check at both places with ObjectUtils.isEmpty

strailov: made -1 acceptable

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaDeploymentManagement.java
avgustinmm: aha, I got it.

hawkbit-mgmt/hawkbit-mgmt-resource/src/main/java/org/eclipse/hawkbit/mgmt/rest/resource/MgmtTargetResource.java
strailov: added ObjectUtils checks + throw exception also if keeplast is 0

hawkbit-mgmt/hawkbit-mgmt-resource/src/main/java/org/eclipse/hawkbit/mgmt/rest/resource/MgmtActionResource.java
```diff
@@ -71,6 +75,34 @@ public ResponseEntity<MgmtAction> getAction(final Long actionId) {
...
+    public ResponseEntity<Void> deleteAction(Long actionId) {
+        deploymentManagement.deleteAction(actionId);
+        return ResponseEntity.ok().build();
+    }
+
+    @Override
+    @AuditLog(entity = "Actions", type = AuditLog.Type.DELETE, description = "Delete Actions", logResponse = true)
+    public ResponseEntity<Void> deleteActions(String rsqlParam, List<Long> actionIds) {
+
+        final boolean isActionIdsEmpty = !ObjectUtils.isEmpty(actionIds);
```
avgustinmm: isActionIdsEmpty confusing name of variable. it is isActionIdsNOTEmpty

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaDeploymentManagement.java
```diff
@@ -777,14 +888,28 @@ private void enforceMaxAssignmentsPerRequest(final int requestedActions) {
...
+        systemSecurityContext.runAsSystem(() -> {
+            final Optional<JpaTarget> opt = targetRepository.findByControllerId(controllerId);
+            if (opt.isPresent()) {
+                final Target target = opt.get();
+                try {
+                    QuotaHelper.assertAssignmentQuota(controllerId, requested, quota, Action.class, Target.class,
+                            actionRepository::countByTargetControllerId);
+                } catch (final AssignmentQuotaExceededException ex) {
+                    // assume requested are always smaller than int size
+                    maxAssignmentsExceededHandler.accept(target.getId(), Math.toIntExact(requested), ex);
```
avgustinmm: handling of exception - e.g. purge of some action should be made outside systemSecurityContext.runAsSystem in order to be run with the caller's permission
Otherwise, all callers shall check in front that the user has target update permission for that target

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaDeploymentManagement.java
```diff
@@ -401,6 +403,64 @@ public Action forceTargetAction(final long actionId) {
         return action;
     }
 
+    @Override
+    @Transactional
+    public void deleteAction(final long actionId) {
+        log.info("Deleting action {}", actionId);
+        actionRepository.getAccessController().ifPresent(accessController ->
+                actionRepository.findAll(accessController.appendAccessRules(AccessController.Operation.UPDATE,
```
avgustinmm: should check that it return the controller - this check doesn't check anything
if no update permission - will return just the empty result

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaDeploymentManagement.java
```diff
@@ -401,6 +403,64 @@ public Action forceTargetAction(final long actionId) {
...
+        actionRepository.deleteById(actionId);
+    }
+
+    @Override
+    @Transactional
+    public void deleteActionsByRsql(final String rsql) {
+        log.info("Deleting actions matching rsql {}", rsql);
+        final Specification<JpaAction> specification = QLSupport.getInstance().buildSpec(rsql, ActionFields.class);
+        actionRepository.getAccessController().ifPresent(accessController ->
+                actionRepository.findAll(accessController.appendAccessRules(AccessController.Operation.UPDATE, specification)));
```
avgustinmm: should, if there is access controller call actionRepository with accessController.appendAccessRules(AccessController.Operation.UPDATE, specification
this findall does nothing.

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaDeploymentManagement.java
strailov: Why ? If I as a user want to assign a DS to a target and my tenant admin have configured purge of actions isn't this like a "done deal" ? I mean I only need to have rights to assign the ds, the configuration for deletion is already done by the admin and in this context system deletes/purges actions ? At least this is my point of view.

avgustinmm: yes. If he has target update in order to assign and create action - it probably is the case. If he has not the action creation shall fail. But it will fail the action create transaction. So, we shall always be sure that we are in that transaction or failing that will fail delete transaction. I'm not really sure here.
Anyway all "system code" shall be as small as possible - that's why, since it's not big work - better move handling out.

PR #2809, opened 2025-11-17
Remove 'Simple' from Simple UI

avgustinmm:


b4u-mw:
Hello,

this commit breaks https://github.com/eclipse-hawkbit/hawkbit/blob/master/docker/mysql/docker-compose-monolith-with-ui-mysql.yml because hawkbit/hawkbit-ui does not exist on docker hub.

avgustinmm:
Yes. This hawkbit/hawkbit-ui has will be released on docker hub with next release. 
At the moment, on docker hub you have hawkbit/hawkbit-simple-ui/0.9.0 (the last UI version). 
So, for running it, you should use the correspondent docker compose files from the 0.9.0 tag, e.g. https://github.com/eclipse-hawkbit/hawkbit/blob/0.9.0/docker/mysql/docker-compose-monolith-with-simple-ui-mysql.yml

PR #2811, opened 2025-11-17
Fix simple UI docker image

strailov:
Turned out that when In Docker environment in Simple UI container tomcat uses Atmosphere framework under the hood and Atmosphere Annotations processing constantly recursively scans the filesystem which leads to high cpu constantly. 
This PR changes the WORKDIR in the Docker container which preserves scanning of dirs like sys, tmp etc ...

### DIFF
--- a/docker/build/Dockerfile_dbinit
+++ b/docker/build/Dockerfile_dbinit
@@ -20,8 +20,7 @@ RUN set -x &&\
  apk del build-dependencies
 
 FROM eclipse-temurin:${JAVA_VERSION}-jre-alpine
-ARG APP_DIR=/app
-WORKDIR ${APP_DIR}
+WORKDIR /app
 RUN addgroup -S hawkbit_group && adduser -D hawkbit -G hawkbit_group
 USER hawkbit
 
@@ -31,7 +30,7 @@ ARG HAWKBIT_VERSION=pass_as_buils_arg
 ENV APP=${HAWKBIT_APP}
 ENV VERSION=${HAWKBIT_VERSION}
 
-COPY --from=build /${APP_DIR}/${APP}-${VERSION}.jar ./
+COPY --from=build ${APP}-${VERSION}.jar ./
 
 ARG X_MS=256m
 ENV X_MS=${X_MS}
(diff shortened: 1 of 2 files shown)


docker/build/Dockerfile
```diff
@@ -15,20 +15,21 @@ COPY KEY .
...
- rm /${APP}.jar
+ rm //${APP}.jar
 
 FROM eclipse-temurin:${JAVA_VERSION}-jre-alpine
-
+ARG APP_DIR=/app
+WORKDIR ${APP_DIR}
 RUN addgroup -S hawkbit_group && adduser -D hawkbit -G hawkbit_group
-RUN mkdir -p /artifactrepo && chown -R hawkbit /artifactrepo
+RUN mkdir -p /${APP_DIR}/artifactrepo && chown -R hawkbit /${APP_DIR}/artifactrepo
```
avgustinmm: maybe it shall be just artifactrepo - since we are in /app

docker/build/Dockerfile_dbinit
```diff
@@ -30,7 +31,7 @@ ARG HAWKBIT_VERSION=pass_as_buils_arg
 ENV APP=${HAWKBIT_APP}
 ENV VERSION=${HAWKBIT_VERSION}
 
-COPY --from=build /${APP}-${VERSION}.jar ./
+COPY --from=build /${APP_DIR}/${APP}-${VERSION}.jar ./
```
avgustinmm: no need of /${APP_DIR}/ - could be just ${APP}...
even could just set WORKDIR and work in it

docker/build/Dockerfile_dbinit
```diff
@@ -47,4 +48,4 @@ ENV GC=${GC}
 ARG AND_THEN="exit 0"
 ENV AND_THEN=${AND_THEN}
 
-ENTRYPOINT ["sh", "-c", "java -classpath /${APP}-${VERSION}.jar ${JAVA_OPTS} -Xms${X_MS} -Xmx${X_MX} -XX:MaxMetaspaceSize=${XX_MAX_METASPACE_SIZE} -XX:MetaspaceSize=${XX_METASPACE_SIZE} -Xss${XSS} -XX:+Use${GC}GC -XX:+UseStringDeduplication -XX:+UseCompressedOops -XX:+HeapDumpOnOutOfMemoryError org.eclipse.hawkbit.repository.jpa.init.HawkbitFlywayDbInit ${@} && ${AND_THEN}"]
+ENTRYPOINT ["sh", "-c", "java -classpath /${APP_DIR}/${APP}-${VERSION}.jar ${JAVA_OPTS} -Xms${X_MS} -Xmx${X_MX} -XX:MaxMetaspaceSize=${XX_MAX_METASPACE_SIZE} -XX:MetaspaceSize=${XX_METASPACE_SIZE} -Xss${XSS} -XX:+Use${GC}GC -XX:+UseStringDeduplication -XX:+UseCompressedOops -XX:+HeapDumpOnOutOfMemoryError org.eclipse.hawkbit.repository.jpa.init.HawkbitFlywayDbInit ${@} && ${AND_THEN}"]
```
avgustinmm: you are in WORKDIR so just ${APP}-${VERSION} instead of /${APP}-${VERSION} shall do the work

PR #2846, opened 2025-12-09
Support for action cancellation in ddi controller sdk

strailov:


hawkbit-sdk/hawkbit-sdk-device/src/main/java/org/eclipse/hawkbit/sdk/device/DdiController.java
```diff
@@ -165,31 +183,21 @@ private void poll() {
...
                                             } else if (currentActionId != actionId) {
-                                                // TODO - cancel and start new one?
+                                                // currentActionId had failed to be processed and new one had been initiated
+                                                // try cancel current and process new one
                                                 log.info(LOG_PREFIX + "Action {} is canceled while in process (new {})!", getTenantId(),
                                                         getControllerId(), currentActionId, actionId);
+                                                cancelActionByCancellationLink(controllerBase, currentActionId);
+                                                currentActionId = null;
+                                                // then process the new one
+                                                poll();
```
desislava-marinova, line 195: Is `poll()` needed here? Isn't `processAction(actionId, actionWithDeployment, executor)` enough?

hawkbit-sdk/hawkbit-sdk-device/src/main/java/org/eclipse/hawkbit/sdk/device/DdiController.java
```diff
@@ -165,31 +183,21 @@ private void poll() {
...
                                             } // else same action - already processing
                                         }, () -> {
-                                            if (currentActionId != null) {
-                                                // TODO - cancel current?
-                                                log.info(LOG_PREFIX + "Action {} is canceled while in process (not returned)!", getTenantId(),
-                                                        getControllerId(), getCurrentActionId());
-                                            }
+                                            cancelActionByCancellationLink(controllerBase, -1);
```
desislava-marinova, line 198: This will cancel action with `actionId`, not with `currentActionId` if they differ, am I right?
Is this the required behaviour or action with `currentActionId` should be cancelled if `currentActionId != null`?

strailov: If -1 is passed as an argument actionId should be obtained from the cancellation link provided into the response - `getActionIdFromCancellationLink` is applied in that case. 
Particulary this line of code should be executed only when : 

1. device polls
2. checks that there is no deploymentBase link present
3. Then it searches for cancelation link
4. if such is present - confirms cancelation
5. if not - does nothing.

hawkbit-sdk/hawkbit-sdk-device/src/main/java/org/eclipse/hawkbit/sdk/device/DdiController.java
strailov: Not necessarily. I was thinking about this and if we use processAction here this could eliminate the case where we have a couple concequitive actions in canceling state waiting to be confirmed as canceled. Process action just gets the deployment base and passes the info to an update handler - it expects to have assigned action, not canceled. 

Potentially we also could do nothing and just wait for the next device poll (to avoid possible loop with recursion also)

hawkbit-sdk/hawkbit-sdk-device/src/main/java/org/eclipse/hawkbit/sdk/device/DdiController.java
desislava-marinova: What about the case when `currentActionId != null` and is not the same as the `actionId ` from the cancellation link, should currentAction be cancelled?

strailov: Yes - in theory it should be the same(it also could differ) - this use case is when : 
currentActionId had failed to be processed (like download error which sdk does not handle correctly) and new action had been initiated on Backend side - then we reach this part of the code. So the only option I see is to try cancel the one present in the cancellation link (if such exists) and process the next one.

desislava-marinova: Does it mean that the one with the `currentActionId` should not be cancelled in this case (when the `currentActionId` != action from the cancellation link?

strailov: Yes - even if currentActionId is different - if it is not present in cancellation link on poll - it could not be cancelled. 
Meaning - always the present action on poll should be cancelled/processed first. So thats why we just obtain the cancellation from the.

desislava-marinova (approved):

PR #2870, opened 2026-01-15
Support target metadata management in the UI

momobel:
This PR adds target metadata view in the target details where they can be deleted with a confirmation.

view
del

And added:
add

### DIFF
--- a/hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/view/TargetView.java
+++ b/hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/view/TargetView.java
@@ -38,6 +38,7 @@ import com.vaadin.flow.component.button.Button;
 import com.vaadin.flow.component.button.ButtonVariant;
 import com.vaadin.flow.component.checkbox.CheckboxGroup;
 import com.vaadin.flow.component.combobox.ComboBox;
+import com.vaadin.flow.component.confirmdialog.ConfirmDialog;
 import com.vaadin.flow.component.datetimepicker.DateTimePicker;
 import com.vaadin.flow.component.dependency.Uses;
 import com.vaadin.flow.component.formlayout.FormLayout;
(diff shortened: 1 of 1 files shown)


hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/view/TargetView.java
```diff
@@ -586,6 +592,76 @@ private List<MgmtTag> fetchAvailableTags() {
...
+            metadataArea.addColumn(MgmtMetadata::getValue).setHeader(VALUE).setAutoWidth(true);
+            metadataArea.addComponentColumn(metadata -> {
+                final Button deleteBtn = Utils.tooltip(new Button(VaadinIcon.TRASH.create()), "Delete Metadata");
+                deleteBtn.addClickListener(e -> confirmDeleteDialog(metadata.getKey()));
+                return deleteBtn;
+            }).setHeader("Actions").setAutoWidth(true).setFlexGrow(0);
+            metadataArea.setWidthFull();
+            add(metadataArea);
+
+            final Button addBtn = new Button("Add Metadata");
```
strailov: I would exclude this button from the current vertical layout because of the setWidthFull() method invocation from above. I would prefer the button NOT to be with full width. 
So my suggestion here is to wrap it into a HorizontalLayout - this way if there are new actions to be added - they just have to be added into the layout

strailov: I would leave the name to just "Add" (without the metadata). Maybe "Add Metadata" is for a tooltip (could be added also)

momobel: Changed to a separate HorizontalLayout

momobel: Changed to "Add"

strailov:
Looks good!
Thanks @momobel  for you contribution in hawkBit!

PR #2872, opened 2026-01-16
ui: Use simple target filter as primary

momobel:
This PR makes it easier to search targets by preferring the simple search and keeping the RSQL as backup.


strailov:
Looks good! 
Thanks @momobel for your contribution!

PR #2882, opened 2026-01-22
ui: Make clearer which target's detail is currently open

momobel:
This MR clarifies which target's details is currently displayed by:
- pining the action button (view/close)
- show the target name and ID in the details panel

### DIFF
--- a/hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/view/TargetView.java
+++ b/hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/view/TargetView.java
@@ -342,8 +342,9 @@ public class TargetView extends TableView<TargetView.TargetWithDs, String> {
         }
     }
 
-    protected static class TargetDetailedView extends TabSheet {
+    protected static class TargetDetailedView extends VerticalLayout {
 
+        private final Span targetId;
         private final TargetDetails targetDetails;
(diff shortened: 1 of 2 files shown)


hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/view/TargetView.java
```diff
@@ -342,30 +342,36 @@ public void setFilter(String filter) {
...
+            tabSheet.add("Details", targetDetails);
+            tabSheet.add("Assigned / Installed", targetAssignedInstalled);
+            tabSheet.add("Tags", targetTags);
+            tabSheet.add("Metadata", targetMetadata);
+            tabSheet.add("Action History", targetActionsHistoryLayout);
+            add(tabSheet);
         }
 
         private void setItem(final MgmtTarget target) {
+            this.targetId.setText(target.getName() + " (" + target.getControllerId() + ")");
```
strailov: I would leave only onе of name / controllerId here because now it seems quite confusing which is the id and which is the name ... 
Most prefarably I would only leave the controllerId.

momobel: Sure! I updated to only display the controllerId.

Issue #2745, opened 2025-10-10
Question: POST /rest/v1/distributionsets clarifications


fkromer:
`MgmtDistributionSetRequestBodyPost` does not define at least one required property and no default values applied implicitly if not passed explicitly schema wise. The example in the schema is:

```
[
  {
    "name": "dsOne",
    "description": "Description of the distribution set.",
    "version": "1.0.0",
    "locked": true,
    "requiredMigrationStep": false,
    "modules": [
      {
        "id": 108
      }
    ],
    "type": "test_default_ds_type"
  }
]
```

If I create a distribution set with the following request body (**code wise the `name` and `version` are the only required props**)

```
{
    "name": "ds1",
    "version": "1.0.0"
}
```

I get a response containing

```
        "name": "ds1",
        "id": 1,
        "version": "1.0.0",
        "type": "os_app",
        "typeName": "OS with app(s)",
        "complete": false,
        "locked": false,
        "deleted": false,
        "valid": true,
        "requiredMigrationStep": false,
```

which answers the question about implicit default values.

The lack of the info about required props and description about how `complete`, `locked`, `deleted`, `valid` work in the openapi spec schema is not critical. It implies the need for implicit knowledge however which makes working with the API pretty difficult. If someone knows about the behavior still involved in the project: Would it be possible to add this info into the related code (the openapi spec is generated from). I have guesses only...

fkromer:
If one tries to create a rollout with a distribution set as input this might lead to

```
{
    "exceptionClass": "org.eclipse.hawkbit.repository.exception.IncompleteDistributionSetException",
    "errorCode": "hawkbit.server.error.distributionset.incomplete",
    "message": "Distribution set of type os_app is incomplete: 1"
}
```

It's not critical but user-unfriendly and we can improve the API usage experience by providing more complete API usage info in the openapi spec.

vasilchev:
Hello,
regarding the Creation of DistributionSet ('POST') Request:

'modules' is important part of a DistributionSet - it defines the 'complete' field.
Distribution Sets that are Not 'complete' cannot be used in a Rollout (In the end Distribution Sets must contain the Software Modules defined by the DistributionSet Type (required/optional sw modules).
Distribtuion Set 'complete' flag also is calculated dynamically by checking the 'complete' flag of Software Modules assigned to the DistributionSet.

Currently it is not mandatory as you can create the DistributionSet first and later adjust accordingly (i.e. assign Software Modules). 
[...]

PR #2891, opened 2026-01-26
Addressing some of the recent created Issues

vasilchev:
Addressing:

- #2745 - fix open api missing schema for SoftwareModules in DistributionSet open api schema
- #2713 - Fix doc of SecurityToken length - 128 max length

### DIFF
--- a/docs/authentication.md
+++ b/docs/authentication.md
@@ -18,8 +18,9 @@ tenant.
 
 #### Target Security Token Authentication
 
-There is a 128 alphanumeric character security-token for each created target within IoT hawkBit. This token can be used
-to authenticate the target at hawkBit through the HTTP-Authorization header with the custom scheme _TargetToken_.
+If not set by the user, the security token is automatically generated with 32 alphanumeric characters. 
+However, the user can provide a custom token with a maximum length of 128 alphanumeric characters.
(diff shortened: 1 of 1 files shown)


docs/authentication.md
```diff
@@ -18,7 +18,7 @@ tenant.
 
 #### Target Security Token Authentication
 
-There is a 32 alphanumeric character security-token for each created target within IoT hawkBit. This token can be used
+There is a 128 alphanumeric character security-token for each created target within IoT hawkBit. This token can be used
```
avgustinmm: is it really so? SecurityToken generator generates 16 bites and the token is hex encoded. So, it is 32 alphanumeric - not 128. But is actually even less - not all 32 alphanumeric are possible. In order to get correctly key strength maybe more concrete description will be better - "hex encoded 128 bit key" will be better

vasilchev: This is the SecurityToken generator. Couldn't one use Rest API and just update Target to whatever you want ?
On jpa level we have it configured with max length 128

`       /**
     * Maximum length of securityToken.
     */
    int SECURITY_TOKEN_MAX_SIZE = 128;
`

avgustinmm: but it is a max - not 128. Said like this user will expect always 128 chars. So, we have 32 generated but could be set up to 128..

PR #2892, opened 2026-01-27
feat: add target filter query view

MikaMauger:
This MR introduces the `Target Filter Queries` view, aligned with the previous Hawkbit UI.
Creation is not included yet, as target filter queries can already be created from the Targets view.

Currently, you can:

- list target filter queries
- automatically assign a distribution set to a filter
- unassign a distribution set
- delete a single filter query
- bulk delete multiple filter queries

Screenshot From 2026-01-26 19-05-24
Screenshot From 2026-01-27 11-54-48

### DIFF
--- a/hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/view/TargetFilterQueryView.java
+++ b/hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/view/TargetFilterQueryView.java
@@ -1,14 +1,26 @@
+/**
+ * Copyright (c) 2026 Contributors to the Eclipse Foundation
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
 package org.eclipse.hawkbit.ui.view;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.vaadin.flow.component.Component;
 import com.vaadin.flow.component.Key;
+import com.vaadin.flow.component.Unit;
 import com.vaadin.flow.component.button.Button;
 import com.vaadin.flow.component.button.ButtonVariant;
 import com.vaadin.flow.component.combobox.ComboBox;
 import com.vaadin.flow.component.confirmdialog.ConfirmDialog;
 import com.vaadin.flow.component.dependency.Uses;
+import com.vaadin.flow.component.formlayout.FormLayout;
 import com.vaadin.flow.component.grid.Grid;
+import com.vaadin.flow.component.html.Div;
 import com.vaadin.flow.component.html.Paragraph;
 import com.vaadin.flow.component.html.Span;
 import com.vaadin.flow.component.icon.Icon;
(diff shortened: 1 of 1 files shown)


avgustinmm:
@MikaMauger , 
I don't mind the double brackets the SonarCloud fail for - we could accept it.
But if you don't resolve **eclipsefdn/eca** compliance issue - we could not accept the PR. While this issue is present, we won't start any review.

MikaMauger:
Done! Also, the double brackets are from copied and pasted existing code. I can remove them, but I'll wait for the full review before doing so.

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/view/TargetFilterQueryView.java
```diff
@@ -0,0 +1,292 @@
...
+            if (filter.getAutoAssignDistributionSet() == null) {
+                add(Utils.tooltip(new Button(VaadinIcon.LINK.create()) {
+                    {
+                        addClickListener(v -> {
+                            new AutoAssignDialog(filter.getId(), hawkbitClient, () -> refresh(filter.getId())).open();
+                        });
+                    }
+                }, "Auto assign"));
+            } else {
+                add(Utils.tooltip(new Button(VaadinIcon.UNLINK.create()) {
```
strailov: see above comment

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/view/TargetFilterQueryView.java
```diff
@@ -0,0 +1,292 @@
...
+@RolesAllowed({ "TARGET_READ" })
+@Uses(Icon.class)
+public class TargetFilterQueryView extends TableView<TargetFilterQueryView.TargetFilterQueryGridItem, Long>  {
+    public TargetFilterQueryView(final HawkbitMgmtClient hawkbitClient) {
+        super(
+                new TargetFilterQueryFilter(),
+                new SelectionGrid.EntityRepresentation<>(TargetFilterQueryGridItem.class, TargetFilterQueryGridItem::getId) {
+
+                    @Override
+                    protected void addColumns(final Grid<TargetFilterQueryGridItem> grid) {
```
strailov, line 69: I would like to see the actual query somehow - after all it is the most important thing of the target filter. Wether it would be as a tooltip or on click with a modal window IDK yet which one will be best. I guess if it is added as a column could be too long sometimes ...

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/view/TargetFilterQueryView.java
```diff
@@ -0,0 +1,292 @@
+package org.eclipse.hawkbit.ui.view;
```
strailov, line 10: Add missing headers. See [here](https://github.com/eclipse-hawkbit/hawkbit/blob/master/licenses/LICENSE_HEADER_TEMPLATE_CONTRIBUTORS_25.txt) for example.

hawkbit-ui/src/main/java/org/eclipse/hawkbit/ui/view/TargetFilterQueryView.java
```diff
@@ -0,0 +1,292 @@
...
+import org.eclipse.hawkbit.ui.view.util.Utils;
+
+import java.util.*;
+import java.util.concurrent.CompletableFuture;
+
+@PageTitle("Target Filter Queries")
+@Route(value = "target_filter_queries", layout = MainLayout.class)
+@RolesAllowed({ "TARGET_READ" })
+@Uses(Icon.class)
+public class TargetFilterQueryView extends TableView<TargetFilterQueryView.TargetFilterQueryGridItem, Long>  {
```
strailov, line 61: Overall looks good. However currently we created target filters in the target view - because we didn't have a target filter view. Now that it is introduced it makes sense for it to have a create option there maybe ... What do you think ?

MikaMauger:
OK, I have taken your comments into account, so I changed a little bit the table to be able to see the query (with ellipsis and a tooltip):
new_screen
I also added the details panel:
details
What do you think?

PS: The creation flow will require additional work. A small modal with only the required fields wouldn’t be sufficient; we also need a way to display and review the actual query results. Since we already have it in the "Targets" view, I think this could be addressed in a separate PR.

strailov:
Looks good! Thanks @MikaMauger for your contribution!

PR #2901, opened 2026-01-30
Some small UI tweaks

momobel:
- Extend target details tab sheet to cover remaining of the screen
- Delimitate target details from the target list


strailov (approved):
Looks good! Thanks @momobel !

PR #2904, opened 2026-02-03
docker: Update hawkBit UI image

yashi:
`hawkbit/hawkbit-ui` doesn't exist any more.  Replace it with `hawkbit/hawkbit-simple-ui`.


yashi:
Nevermind then.

PR #2906, opened 2026-02-04
Unify flyway schemas and fix PostgreSQL v1_12_35

avgustinmm:


hawkbit-repository/hawkbit-repository-jpa-flyway/src/main/resources/db/migration/POSTGRESQL/V1_12_35__sm_type_min_artifacts__POSTGRESQL.sql
```diff
@@ -1,4 +1,4 @@
 ALTER TABLE sp_software_module_type ADD COLUMN min_artifacts integer default 0 NOT NULL;
-DROP INDEX sp_idx_distribution_set_01;
+DROP INDEX sp_idx_distribution_set_01_sp_distribution_set;
```
vasilchev, line 2: this edits intermediate migration - is that even ok ?
what happens if already processed but now it is with different checksum ?


I see two indexes 

Create in

POSTGRESQL/V1_12_15__baseline___POSTGRESQL.sql
`CREATE INDEX **sp_idx_distribution_set_01_sp_distribution_set**
ON sp_distribution_set
USING BTREE (tenant, deleted, complete);`

RE-CREATE in 
POSTGRESQL/V1_12_35__sm_type_min_artifacts__POSTGRESQL.sql
`DROP INDEX **sp_idx_distribution_set_01**;`
`CREATE INDEX **sp_idx_distribution_set_01** ON sp_distribution_set USING BTREE (tenant, deleted);`


So maybe in 1_12_35 where it was suppose to drop the one in 1_12_15 and recreate it it actually dropped nothing and just created the new one ? or it was wrong and actually migration would fail because could not find the 'sp_idx_distribution_set_01' and actually this never worked

avgustinmm: migration used to fail - because the index to delete does not exist - so it is assumed it has never been applied. 
so if we don't fix it - the whole chain could not be applied. So:
- for those that have never applied 1_12_35 - this fix will just work
- those hat have tried 1_12_35 have failed (none reported that) and probably shall do some manual fixing anywat and apply the new version

PR #2941, opened 2026-02-26
[#2845] Bump Spring boot to 4.x

avgustinmm:
Notes:
1. (!) Eclipselink shall be migrated to 5.0 (in 4.0.8 there are incompatible classes, e.g EJBQueryImpl doesn't implement some newer methods). In the moment is with beta (5.0.0-B12) - JUST for testing!
2. (!) Ethlo plugin doesn't work with Eclipselink 5.0, it builds with Eclipselink 4.0.8 (could be a problem)
3. Dependencies - new starters, test starters changes, some dependencies refactoring
4. Auto-configs split - package changes, some properties classes changes
5. Spring nullable org.springframework.lang.Nullable/NonNull are depecated and replaced with jspcify -> org.jspecify.annotations.Nullable/NonNull (NullMarked)
6. Lombok config - adding lombok.addNullAnnotations=jspecify - to do not mess annotations
7. Distributed lock table changes - SP_LOCK table db migration
8. Spring Retry replaced with Spring Core Retry - does repace retry in hawkbit
9. Specifications -> added Update/Delete(/Predicate) Specifications and JpaSpecificationExecutor changed
10. HawkbitBaseRepositoryFactoryBean modified to register properly
11. Jackson - 2 -> 3, package migrations, finals are not deserialized by default(enable finals deserialization, consider make non-final), too ‘smart’ tries to set complex objects instead of using non args constructor (-> @JsonIgnore), some other default configs made


hawkbit-repository/hawkbit-repository-jpa-flyway/src/main/resources/db/migration/POSTGRESQL/V1_20_1__spring_boot_4__POSTGRESQL.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE SP_LOCK ADD COLUMN EXPIRED_AFTER TIMESTAMP NOT NULL;
```
strailov, line 1: Wouldn't this fail in cases where SP_LOCK has some rows in it ? 
DB would try to populate every existing row with that column and its value, but since there is NOT NULL and no default is specified wouldn't that query fail ?

hawkbit-repository/hawkbit-repository-jpa/src/test/resources/jpa-test.properties
```diff
@@ -58,8 +58,8 @@ logging.level.org.eclipse.persistence=${NOISE_SUPPRESS_LEVEL}
 ## enable / disable case sensitiveness of the DB when playing around
 #hawkbit.ql.case-insensitive-db=true
 
-hawkbit.repository.cluster.lock.ttl=3000
-hawkbit.repository.cluster.lock.refreshOnRemainMS=2000
+hawkbit.repository.cluster.lock.ttl=1000
```
strailov, line 61: Is there a reason to revert those ?

hawkbit-dmf/hawkbit-dmf-amqp/src/main/java/org/eclipse/hawkbit/amqp/DmfApiConfiguration.java
```diff
@@ -91,13 +93,16 @@ public ErrorHandler errorHandler(
      * @return {@link RabbitTemplate} with automatic retry, published confirms and {@link Jackson2JsonMessageConverter}.
      */
     @Bean
-    public RabbitTemplate rabbitTemplate() {
+    public RabbitTemplate rabbitTemplate(final JsonMapper jsonMapper) {
         final RabbitTemplate rabbitTemplate = new RabbitTemplate(rabbitConnectionFactory);
-        rabbitTemplate.setMessageConverter(new Jackson2JsonMessageConverter());
+        rabbitTemplate.setMessageConverter(new JacksonJsonMessageConverter(jsonMapper));
```
strailov, line 98: Maybe we can test that in our case both converters are compatible somehow ? 
My concerns are when performing a new deployment with rolling-update if there are some messages that are incompatible with the new converter ?

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/configuration/Constants.java
```diff
@@ -28,12 +27,13 @@ public final class Constants {
      * number.
      */
     public static final int MAX_ENTRIES_IN_STATEMENT = 999;
+
     /**
-     * @see Retryable#maxAttempts()
+     * See {@link Retryable#maxRetries()}
```
strailov, line 32: Are we aware that attempt != retry ?
old way assumes 10 attempts, but the new approach adds 10 retries to an attempt = 11 ? Just to double check.

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaDeploymentManagement.java
```diff
@@ -144,15 +148,20 @@ protected JpaDeploymentManagement(
...
         this.targetRepository = targetRepository;
         this.entityManager = entityManager;
         this.txManager = txManager;
         this.database = jpaProperties.getDatabase();
 
-        retryTemplate = createRetryTemplate();
+        retryTemplate = new RetryTemplate(RetryPolicy.builder()
+                .includes(ConcurrencyFailureException.class)
+                .maxRetries(maxRetries)
+                .maxDelay(Duration.ofMillis(delay))
```
strailov, line 163: The old approach used 
`backOffPolicy.setBackOffPeriod(Constants.TX_RT_DELAY);` 
This was the initial delay right ? Current max Delay seems different in semanthics ? 
Maybe we could add .delay() into the builder with some configured value ?

strailov (approved):

hawkbit-repository/hawkbit-repository-jpa-flyway/src/main/resources/db/migration/POSTGRESQL/V1_20_1__spring_boot_4__POSTGRESQL.sql
avgustinmm: SP_LOCK table is used by spring database lock and we strictly follow the spring table definition. 
Probably spring lock has set timestamp if the column has existed. If there are nulls - db cleanup may be needed.
it is possible to add table cleanup before - but not sure about concurrency - it will unlock table eventually ot

hawkbit-repository/hawkbit-repository-jpa/src/test/resources/jpa-test.properties
avgustinmm: test faild with the 3000/2000 config when build by verify github actions. Also use to always fail localy. 
So I returned it back to previous config - with that it works

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/configuration/Constants.java
avgustinmm: I suppose that the number is just a "guess" or something random. and 10 is as good as 11

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaDeploymentManagement.java
avgustinmm: Good point. Actually the old approach was with backoff new is with delay - an attempt timeout ..
not sure what is the proper way to configure..

avgustinmm: from https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/util/backoff/FixedBackOff.html it seems that Fixed backoff is delay between attempts which probably is the same as now - delay beteewn attempts

PR #2948, opened 2026-03-04
Remove the ROLE_CONTROLLER_ANONYMOUS

avgustinmm:


hawkbit-dmf/hawkbit-dmf-amqp/src/main/java/org/eclipse/hawkbit/amqp/AmqpMessageHandlerService.java
```diff
@@ -172,7 +172,7 @@ private static void setSecurityContext(final Authentication authentication) {
     private static void setTenantSecurityContext(final String tenant) {
         final AnonymousAuthenticationToken authenticationToken = new AnonymousAuthenticationToken(
```
vasilchev, line 173: maybe could be renamed to ControllerAuthenticationToken as well

avgustinmm: it's not exactly controller token since it doesn't identify the controller but the tenant

PR #2965, opened 2026-03-20
Modify TargetPollEvent to be per batch update (not per target)

strailov:


hawkbit-repository/hawkbit-repository-api/src/main/java/org/eclipse/hawkbit/repository/event/remote/TargetPollEvent.java
```diff
@@ -29,16 +31,22 @@ public class TargetPollEvent extends RemoteTenantAwareEvent {
     @Serial
     private static final long serialVersionUID = 1L;
 
-    private String controllerId;
     private String targetAddress;
+    private long lastTargetPoll;
```
vasilchev: do we need this lastTargetPoll ? each event has 'timestamp' defining "when this event occurred/published"
in the end it will have same value as 'timestamp' as well (or very close, some miliseconds of the actual 'commit')

if idea is to capture the actual 'lastTargetPoll' should be done when initially the poll is received (and the event put in queue for delayed handle), but not sure we need/want that.

vasilchev: It seems initial impl was a better approach - separate property 'lastTargetPoll' (or 'lastPollTime')
1) 'timestamp' remains the RemoteEvent publish timestamp
2) 'lastTargetPoll' (or 'lastPollTime) body of the event - i.e. the last poll time - semantically different, eventhough that with current impl will be close or equal

+ the value('lastTargetPoll) has to be the same value which is written in DB (not different one on the time of creation of the event.
 
 Apologies for the confusion, initial approach was better - please revert https://github.com/eclipse-hawkbit/hawkbit/pull/2965/commits/27e3b740e2bb0b02cad1d5a6137db7928901f069 commit from this PR and go with the initial

PR #3007, opened 2026-04-09
Fix test assertions that depend on undefined row ordering

clayly:
## Summary

Several integration tests use order-sensitive assertions (`containsExactly`, index-based access) on query results from queries with no `ORDER BY` clause. SQL does not guarantee row ordering without explicit `ORDER BY`, and this causes deterministic test failures on databases that don't return rows in insertion order (e.g. YugabyteDB, CockroachDB, or any distributed PostgreSQL-compatible DB).

The tests verify set membership (correct targets assigned, correct actions stored), not ordering. Changed to order-independent assertions:

- **AutoAssignTest** (3 tests): `containsExactly` → `containsExactlyInAnyOrder` on `findByAssignedDistributionSet()` results
- **ControllerManagementTest.updatedExternalRefOnActionIsReallyUpdated**: index-based loop → `containsExactlyInAnyOrderElementsOf`
- **TargetFilterQueryManagementTest**: `containsExactly` → `containsExactlyInAnyOrder` in `verifyExpectedFilterQueriesInList` helpers

## Test plan

- [x] Verified passing on H2 (default)
- [x] Verified passing on YugabyteDB (PostgreSQL 15.12-YB-2025.2.2.2-b0)
- [ ] Verify passing on PostgreSQL
- [ ] Verify passing on MySQL

### DIFF
--- a/hawkbit-repository/hawkbit-repository-jpa/src/test/java/org/eclipse/hawkbit/repository/jpa/acm/AutoAssignTest.java
+++ b/hawkbit-repository/hawkbit-repository-jpa/src/test/java/org/eclipse/hawkbit/repository/jpa/acm/AutoAssignTest.java
@@ -76,6 +76,6 @@ class AutoAssignTest extends AbstractAccessControllerManagementTest {
                 .as("Only updatable targets should be part of the rollout")
                 // all targets are distribution set type 2 compatible, but since user has UPDATE_TARGET only for targets of type 2
                 // only target2 and target3 shall be assigned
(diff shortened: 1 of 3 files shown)


clayly:
Hi everyone! Main purpose of such PR is to make it more easy to adapt hawkbit for other Postgres-compatible DB's. Also looks like eclipsedfn/eca check succeeds now for me, do i need to recreate PR?

strailov (approved):

PR #3024, opened 2026-04-16
Cleanup/fix jackson 2 -> 3 migration

avgustinmm:


hawkbit-core/src/main/java/org/eclipse/hawkbit/auth/StaticAuthenticationProvider.java
```diff
@@ -58,15 +60,13 @@ private static UserDetailsService userDetailsService(
...
-            final List<GrantedAuthority> credentials =
-                    createAuthorities(user.getRoles(), user.getPermissions(), Collections::emptyList);
+            final List<GrantedAuthority> credentials = createAuthorities(user.getRoles(), user.getPermissions(), Collections::emptyList);
             userPrincipals.add(ObjectUtils.isEmpty(user.getTenant())
                     ? new User(username, password, credentials)
                     : new TenantAwareUser(username, password, credentials, user.getTenant()));
         });
 
-        if (securityProperties != null && securityProperties.getUser() != null && !securityProperties.getUser().isPasswordGenerated()) {
+        if (securityProperties != null && !securityProperties.getUser().isPasswordGenerated()) {
```
vasilchev, line 69: why null check removed

vasilchev (approved):

hawkbit-core/src/main/java/org/eclipse/hawkbit/auth/StaticAuthenticationProvider.java
avgustinmm: could not be null - see SecurityProperties:
private final User user = new User();

	public User getUser() {
		return this.user;
	}

PR #3082, opened 2026-05-13
Fix swallow OptimisticLockingFailureException as GenericSpServerExcep…

vasilchev:
…tion that prevents @Retryable(includes = ConcurrencyFailureException.class) to work as expected

Having entry 'OptimisticLockingFailureException' in 'MAPPED_EXCEPTION_ORDER' but not in 'EXCEPTION_MAPPING' (leftover) forces ExceptionMapper to map it to 'GenericSpServerException' which then breaks completely all the annotations for retry:
`@Retryable(includes = ConcurrencyFailureException.class,`

Removing it from 'MAPPED_EXCEPTION_ORDER' now just returns it as is -> 'ConcurrencyFailureException' that will be handled by the '@Retryable' annotations.


vasilchev:
Wrong assumption - @retryable is executed - it searches/compares the Exception with traversing=true (looping all exception causes as well, not a direct match).

Having this said, retry is there, just Exception after the retries is mapped to Generic, because of missing mapper.

If we completely remove the MAPPER, OptimisticLockException would be return to caller (could be REST) which is not a great design (at least custom exception should be returned). Better would be to actually introduce back a custom mapped exception for it

PR #3093, opened 2026-05-19
Introduce soft deleted list option for soft deletable entities

strailov:
### Summary

Introduces a `soft_deleted_mode` filter parameter to all soft-deletable entity listing endpoints, allowing API consumers to query soft-deleted entities that were previously hidden from listing results.

### Motivation

Soft-deleted entities (Distribution Sets, Software Modules, their Types, and Rollouts) were previously only accessible by direct ID lookup. There was no way to list or search them via the REST API, making it difficult for UI and integrations to display or manage deleted resources.

### Changes

**New API contract:**
- `SoftDeletedMode` enum (`only_soft_deleted`, `exclude_not_soft_deleted`, `include_soft_deleted`) with safe `fromValue()` parsing
- `SoftDeletableRepositoryManagement` interface extending `RepositoryManagement` with `findAll(SoftDeletedMode, Pageable)` and `findByRsql(String, SoftDeletedMode, Pageable)`
- `count(SoftDeletedMode)` added to `RepositoryManagement` with fallback for non-soft-deletable entities

**Affected REST endpoints:**
- `GET /rest/v1/distributionsets`
- `GET /rest/v1/distributionsettypes`
- `GET /rest/v1/softwaremodules`
- `GET /rest/v1/softwaremoduletypes`
- `GET /rest/v1/rollouts` (compact mode only; full mode intentionally excluded)

**Mutation guards:**
- `update`, `lock`, `unlock`, `invalidate`, `assignTag`, `unassignTag` now reject operations on soft-deleted entities with `DeletedException`

**Implementations:**
- `JpaDistributionSetManagement` — refactored `findByRsql` to support deleted filter while preserving legacy `complete` field handling
- `JpaSoftwareModuleManagement`, `JpaSoftwareModuleTypeManagement`, `JpaDistributionSetTypeManagement` — added `findAll`/`findByRsql` with soft-delete filtering and update guards
- `JpaRolloutManagement` — added `findAll`/`findByRsql` with `SoftDeletedFilter`
- Chosen that approach instead of introducing this logic in `AbstractJpaRepositoryManagement` due to that some entities are not soft-deletable by definition. So the current approach may have some duplications with `findAll` or `findByRsql` implementations but seems with more clear architecture approach and distinguish those entities.
- Soft-delete mutation guards - Soft-deleted entities must reject mutation operations (like Rollout already does). Users can stumble into "Entity exists" errors due to unique key constraint violations on soft-deleted entities they cannot see. Here I rejected the approach with intermediate class due to `AbstractJpaRepositoryManagement` → `AbstractJpaRepositoryWithMetadataManagement` - Would need two variants (with/without metadata) or force unrelated coupling. Not worth the complexity. Decided to override the updates for each entity which is easy to review, no risk for other entities and straightforward .
- Reject all for bulk updates - to be still considered ...


hawkbit-mgmt/hawkbit-mgmt-api/src/main/java/org/eclipse/hawkbit/mgmt/rest/api/MgmtRestConstants.java
```diff
@@ -39,6 +39,17 @@ public final class MgmtRestConstants {
      * The request parameter for specifying the representation mode. The value of this parameter can either be "full" or "compact".
      */
     public static final String REQUEST_PARAMETER_REPRESENTATION_MODE = "representation";
+
+    /**
+     * The request parameter for specifying the soft deletion listing of entities. The value of this parameter
+     * can be soft_deleted, not_soft_deleted or all
+     */
+    public static final String REQUEST_PARAMETER_LIST_SOFT_DELETED_MODE = "soft_deleted_mode";
```
avgustinmm, line 47: The naming of the 3 modes maybe should be reconsidered. What about:
SOFT_DELETED: INCLUDE (all), ONLY (only soft deleted), EXCLUDE (only not deleted)?

hawkbit-mgmt/hawkbit-mgmt-api/src/main/java/org/eclipse/hawkbit/mgmt/rest/api/MgmtRolloutRestApi.java
```diff
@@ -93,7 +95,14 @@ ResponseEntity<PagedList<MgmtRolloutResponseBody>> getRollouts(
...
             @RequestParam(value = REQUEST_PARAMETER_REPRESENTATION_MODE, defaultValue = REQUEST_PARAMETER_REPRESENTATION_MODE_DEFAULT)
-            String representationModeParam);
+            String representationModeParam,
+            @RequestParam(value = REQUEST_PARAMETER_LIST_SOFT_DELETED_MODE, defaultValue = REQUEST_PARAMETER_LIST_SOFT_DELETED_MODE_DEFAULT, required = false)
+            @Schema(description = "Controls whether soft-deleted rollouts are included in the result. " +
+                    "Possible values: 'not_soft_deleted' (default) - returns only non-deleted rollouts; " +
+                    "'soft_deleted' - returns only soft-deleted rollouts; " +
+                    "'all' - returns both active and soft-deleted rollouts.",
+                    allowableValues = { "not_soft_deleted", "soft_deleted", "all" })
+            String softDeletedMode);
```
avgustinmm: why string? It should be enumeration, e.g. SoftDeletedMode:
INCLUDE, EXCLUDE, ALL/ANY or,
ONLY_SOFT_DELETED, EXCLUDE_SOFT_DELETED (default), INCLUDE_SOFT_DELETED, or (with current naming)
SOFT_DELETED, NOT_SOFT_DELETED, ALL/ANY

hawkbit-mgmt/hawkbit-mgmt-resource/src/main/java/org/eclipse/hawkbit/mgmt/rest/resource/MgmtDistributionSetResource.java
```diff
@@ -112,16 +113,18 @@ public class MgmtDistributionSetResource implements MgmtDistributionSetRestApi {
 
     @Override
     public ResponseEntity<PagedList<MgmtDistributionSet>> getDistributionSets(
-            final String rsqlParam, final int pagingOffsetParam, final int pagingLimitParam, final String sortParam) {
+            final String rsqlParam, final int pagingOffsetParam, final int pagingLimitParam, final String sortParam, final String softDeletedModeParam) {
         if (rsqlParam != null && rsqlParam.toLowerCase().contains("complete")) {
             LogUtility.logDeprecated("Usage of MgmtDistributionSetResource.getActions with 'complete': 'complete' distribution set search field is limited and may be removed.");
         }
         final Pageable pageable = PagingUtility.toPageable(pagingOffsetParam, pagingLimitParam, sanitizeDistributionSetSortParam(sortParam));
+        final SoftDeletedFilter softDeletedFilter = SoftDeletedFilter.fromValue(softDeletedModeParam)
```
avgustinmm: Maybe SoftDeltedMode? "Filter" implies something like RSQL in scope of hawkbit

hawkbit-repository/hawkbit-repository-api/src/main/java/org/eclipse/hawkbit/repository/qfields/DistributionSetFields.java
```diff
@@ -36,7 +36,8 @@ public enum DistributionSetFields implements QueryField {
     MODULE("modules", SoftwareModuleFields.ID.getName(), SoftwareModuleFields.NAME.getName()),
     TAG("tags", "name"),
     METADATA("metadata"),
-    VALID("valid");
+    VALID("valid"),
+    DELETED("deleted");
```
avgustinmm: do we want to add deleted field? if we add deleted field we could query with RSQL deleted=true and soft deleted mode "not deleted"? which will take prefernce? or will let user to be responsible for collisions?

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaDistributionSetManagement.java
```diff
@@ -105,46 +107,23 @@ protected JpaDistributionSetManagement(
...
-                                : new Node.Logical(Node.Logical.Operator.AND, sanitizedChildren),
-                        DistributionSetFields.class));
-            }
-            if (completedComparison.get() != null) { // really a comparison
-                log.warn("Usage of 'complete' is limited and may be removed: {}", node);
-                final boolean completed = completeComparison(completedComparison);
-                return filter(JpaManagementHelper.findAllWithCountBySpec(jpaRepository, specList, pageable), completed);
-            }
-        }
+        return findByRsqlAndDeleted(rsql, SoftDeletedFilter.NOT_SOFT_DELETED, pageable);
```
avgustinmm: just idea - what about if we don't add rest soft deleted mode, but just DELETED filter fields?
And here just check - if RSQL contains (parse the filter) deleted=** - we dont add in specification deleted=false.
if don't - like now - we add deleted = false?
This could keep the changes and the api expansion really small as keeping all new features.

hawkbit-repository/hawkbit-repository-api/src/main/java/org/eclipse/hawkbit/repository/qfields/DistributionSetFields.java
strailov: I guess this is the price if we want sort option by deleted if ALL is applied ... Personally I would not include that and not include sort at all.

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaDistributionSetManagement.java
strailov: I think it will be almost the same change - it won't be that small - we still have to filter the entities through the database (which currently is the biggest change), so we will skip maybe 50% of the Controller layer which i don't think is that big now... Also the solution won't be that clean, it feels like a stitch. Idk ...

hawkbit-mgmt/hawkbit-mgmt-resource/src/main/java/org/eclipse/hawkbit/mgmt/rest/resource/MgmtDistributionSetResource.java
strailov: Yeah, I am open for discussion for the names ... I left it softDeletedMode on the Rest API side, but on the implementation side it is still a filter though which goes directly to the db ... So that's why currently I left it that way.

hawkbit-mgmt/hawkbit-mgmt-api/src/main/java/org/eclipse/hawkbit/mgmt/rest/api/MgmtRolloutRestApi.java
avgustinmm: decided for ONLY_SOFT_DELETED, EXCLUDE_SOFT_DELETED (default), INCLUDE_SOFT_DELETED

strailov: Why String - SoftDeletedMode is in the repository layer and operating there (core logic for soft delete). 
Meaning if we want enum in API layer directly that would mean duplicate enum + add mapping (more boilerplate)

avgustinmm: yes, but for feign clients it will be much easier and safe to use API without risk of passing wrong values. 
Also by having enum it will be part of the schema - described in open api - which will also benefits the Open API based clients.

strailov: Ok, will introduce enum on api layer. In open api it is already described using appropriate properties. But yeah, on feign clients this does not apply...

hawkbit-mgmt/hawkbit-mgmt-api/src/main/java/org/eclipse/hawkbit/mgmt/rest/api/MgmtSoftDeletedMode.java
```diff
@@ -0,0 +1,38 @@
...
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.hawkbit.mgmt.rest.api;
+
+import java.util.Arrays;
+import java.util.Optional;
+
+public enum MgmtSoftDeletedMode {
```
avgustinmm, line 14: isn't it sufficient to have just enum without mode? 
I mean spring shall be able to detect ONLY_SOFT_DELETED based on 'only_soft_deleted'? At most case shall be followed. 
Do we really need mode?

hawkbit-mgmt/hawkbit-mgmt-api/src/main/java/org/eclipse/hawkbit/mgmt/rest/api/MgmtRolloutRestApi.java
```diff
@@ -93,7 +95,13 @@ ResponseEntity<PagedList<MgmtRolloutResponseBody>> getRollouts(
                     "The sequence of the sort criteria (multiple can be used) defines the sort order of the entities in the result.")
             String sortParam,
             @RequestParam(value = REQUEST_PARAMETER_REPRESENTATION_MODE, defaultValue = REQUEST_PARAMETER_REPRESENTATION_MODE_DEFAULT)
-            String representationModeParam);
+            String representationModeParam,
+            @RequestParam(value = REQUEST_PARAMETER_LIST_SOFT_DELETED_MODE, defaultValue = REQUEST_PARAMETER_LIST_SOFT_DELETED_MODE_DEFAULT, required = false)
+            @Schema(description = "Controls whether soft-deleted rollouts are included in the result. " +
+                    "Possible values: 'exclude_soft_deleted' (default) - returns only non-deleted rollouts; " +
```
avgustinmm: is there need to enumerate here the values? the enum shall be properly interpreted into the openapi doc. Enumerating here means we shall keep in sync if we decide to change something

hawkbit-repository/hawkbit-repository-api/src/main/java/org/eclipse/hawkbit/repository/SoftDeletedMode.java
```diff
@@ -0,0 +1,27 @@
...
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.hawkbit.repository;
+
+public enum SoftDeletedMode {
+    ONLY_SOFT_DELETED("only_soft_deleted"),
```
avgustinmm: the same - do we need to have mode? the name of the enum value shall be sufficient

licenses/LICENSE_HEADER_TEMPLATE_CONTRIBUTORS_26.txt
```diff
@@ -0,0 +1,7 @@
+Copyright (c) 2026 Contributors to the Eclipse Foundation
```
avgustinmm: hawkbit/licenses/LICENSE_HEADER_TEMPLATE.txt is for 2026 - no need to add this - it will be added next year. for now it's just duplication

pom.xml
```diff
@@ -607,6 +607,7 @@
                                         <validHeader>licenses/LICENSE_HEADER_TEMPLATE_CONTRIBUTORS_23.txt</validHeader>
                                         <validHeader>licenses/LICENSE_HEADER_TEMPLATE_CONTRIBUTORS_24.txt</validHeader>
                                         <validHeader>licenses/LICENSE_HEADER_TEMPLATE_CONTRIBUTORS_25.txt</validHeader>
+                                        <validHeader>licenses/LICENSE_HEADER_TEMPLATE_CONTRIBUTORS_26.txt</validHeader>
```
avgustinmm: hawkbit/licenses/LICENSE_HEADER_TEMPLATE.txt is for 2026 - no need to add this - it will be added next year. for now it's just duplication

hawkbit-repository/hawkbit-repository-api/src/main/java/org/eclipse/hawkbit/repository/RepositoryManagement.java
```diff
@@ -133,6 +133,16 @@ public interface RepositoryManagement<T extends BaseEntity, C, U extends Identif
...
 
+    /**
+     * Counts {@link BaseEntity}s in the repository filtered by their soft-deleted state.
+     * If the {@link BaseEntity} is non-soft deletable, fallbacks to {@link #count()}.
+     *
+     * @param softDeletedMode the filter defining which entities to count based on their soft-deleted status
+     * @return number of {@link BaseEntity}s matching the given soft-deleted filter
+     */
+    @PreAuthorize(SpringEvalExpressions.HAS_READ_REPOSITORY)
+    long count(@NotNull SoftDeletedMode softDeletedMode);
```
avgustinmm: If you extract soft deletable interface SoftDeletableRepositoryManagement - this count method shall go there

hawkbit-repository/hawkbit-repository-api/src/main/java/org/eclipse/hawkbit/repository/RolloutManagement.java
```diff
@@ -188,6 +198,20 @@ Rollout create(
...
+     * @param softDeletedMode the filter defining which rollouts to return based on their soft-deleted status
+     * @param pageable the page request to sort and limit the result
+     * @return a page of found rollouts
+     * @throws RSQLParameterUnsupportedFieldException if a field in the RSQL string is used but not provided by the given
+     *         {@code fieldNameProvider}
+     * @throws RSQLParameterSyntaxException if the RSQL syntax is wrong
+     */
+    @PreAuthorize(SpringEvalExpressions.HAS_READ_REPOSITORY)
+    Page<Rollout> findByRsql(@NotNull String rsql, SoftDeletedMode softDeletedMode, @NotNull Pageable pageable);
+
```
avgustinmm, line 214: don't we want here count with soft deleted mode?

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/AbstractJpaRepositoryManagement.java
```diff
@@ -189,6 +190,20 @@ public long count() {
         return jpaRepository.count(isNotDeleted().orElseGet(Specification::unrestricted));
     }
 
+    @Override
+    public long count(SoftDeletedMode softDeletedMode) {
```
avgustinmm: currently there is softe deleted support here, see supportSoftDelete0, isNotDeleted0 that automatically resolve if there is soft deleted support or not.
So there are two options:
A) stick to this support, and implement using that mechanism, eventually AbstractJpaSoftDeletableRepostioryManagement (or this class to extend and implement SoftDeletableRepostioryManagement)
B) remove soft deletable mechanism from here and move the support entirely into AbstractJpaSoftDeletableRepostioryManagement
Anyway I think a dedicated abstract class with impl of SoftDeletableRepostioryManagement will help to minimise boilerplate code

avgustinmm (changes requested):

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/AbstractJpaRepositoryManagement.java
strailov: It's not that simple I think - DS & SM extend AbstractJpaRepositoryWithMetadataManagement but other entities don't - they lay on AbstractJpaRepositoryManagement . So it again leads to not so clean separation ... 
We would have to add also AbstractJpaSoftDeletableRepositoryWithMetadataManagement for DS & SM eventually ?

hawkbit-repository/hawkbit-repository-api/src/main/java/org/eclipse/hawkbit/repository/RolloutManagement.java
strailov: I even started to think - why we need count at all ? Previously Slice<T> it was used - that's why count was invoked, but It seems everything is page now - meaning why do we need count at all since we have it into the page ?

strailov: This probably would eliminate the need of count(SoftDeletedMode) at all.

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/AbstractJpaRepositoryManagement.java
avgustinmm: if e add abstract class - it will fight with java limitation for extending multiple classes - maybe we could try something with interfaces and default method implementation.
A) however is still possible.

hawkbit-mgmt/hawkbit-mgmt-resource/src/main/java/org/eclipse/hawkbit/mgmt/rest/resource/MgmtApiConfiguration.java
```diff
@@ -28,4 +30,12 @@
 @ComponentScan
 @Import({ RestConfiguration.class, OpenApi.class })
 @PropertySource("classpath:/hawkbit-mgmt-api-defaults.properties")
-public class MgmtApiConfiguration implements WebMvcConfigurer {}
\ No newline at end of file
+public class MgmtApiConfiguration implements WebMvcConfigurer {
+
+    @Override
+    public void addFormatters(final FormatterRegistry registry) {
```
avgustinmm: Is this really needed? Spring shall be able to convert string values to enums

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/AbstractJpaRepositoryManagement.java
strailov: I think it is not relevant anymore ? I have removed the need of the count method as mentioned in the above comments. Without it - the architecture looks more clear I think ...

avgustinmm: count is not exposed via rest so then, for now we probably don't need to export count with soft deleted support

avgustinmm (approved):

PR #3110, opened 2026-06-01
Fix Error/Success Conditions not evaluated case. Fix 'Cancel' final s…

vasilchev:
### Error/Success Conditions skipped if no match, still group marked as finish and nextGroup is triggered

There is a bug in our Rollout  handling whenever Error/Success conditions are not met - and thus not executed, but All Actions have finished - group is marked as finished which leads to.

Error/Success Condition is not executed
Next scheduler run just starts nextGroup
 

Proposal:

If neither Error nor Success Condition/Action triggered but all Actions finished - execute success Action (no change in behavior, only fixes the issue when Success Action is PAUSE) - Only if Error Condition is met then mark group in ERROR state - otherwise consider it success (success action is triggered whenever either Success Condition or Group has finished)

--- 

### If Action is deleted group percentage is not updated correctly -
we count assigned Targets
we count finished Actions
if we delete Actions, but target remain - percentage calculations is broken and could never be met in some cases (i.e. 100% success, on 5 targets, 5 Actions - but delete an Action -> 5 Targets with, 4 Success Actions -> this in theory is 100%, but not handled)
 

Proposal:

**Option 1:** Do nothing - this seems a corner case - Usually if user wants should "force-quit" the Action, not delete it.
**Option 2: Chosen** Count Actions not target (i.e. like Dynamic groups - else condition below)
Current impl:

```
private long countTargetsFrom(final JpaRolloutGroup rolloutGroup) {
    if (rolloutGroup.isDynamic()) {
        return countByActionsInRolloutGroup(rolloutGroup.getId());
    } else {
        return rolloutGroupManagement.countTargetsOfRolloutsGroup(rolloutGroup.getId());
    }
} 
```

---

### Cancel result is not mapped to Error nor Success

Same percentage break whenever we have Cancelled Action. It is not considered neither as Error nor Success. This could also break Error/Success evaluation - i.e. skip Actions, triggerNext group
In case we have 5 Targets, 4 Actions Success, 1 Action Cancelled. on 

Success condition 100%, with PAUSE Success ACTION -> Success action is never executed as condition is not met, but yet triggerNextGroup 

- Ignore Cancel as it is neither Error nor Success -> it will not be calculated in Error/Success Condition.

### DIFF
--- a/hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/scheduler/JpaRolloutExecutor.java
+++ b/hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/scheduler/JpaRolloutExecutor.java
@@ -449,7 +449,7 @@ public class JpaRolloutExecutor implements RolloutExecutor {
                 // 'success' is either group completed or success condition reached - execute 'success' Action
                 boolean groupCompleted = !(rolloutGroup == lastGroup && rolloutGroup.isDynamic()) && isRolloutGroupComplete(rollout, rolloutGroup);
                 checkSuccessCondition(rollout, rolloutGroup, evalProxy, rolloutGroup.getSuccessCondition(), groupCompleted);
-                if (groupCompleted) {// if
+                if (groupCompleted) {
                     rolloutGroup.setStatus(RolloutGroupStatus.FINISHED);
                     rolloutGroupRepository.save(rolloutGroup);
                 }
--- a/hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/scheduler/JpaRolloutExecutor.java
+++ b/hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/scheduler/JpaRolloutExecutor.java
@@ -447,7 +447,7 @@ public class JpaRolloutExecutor implements RolloutExecutor {
                 callErrorAction(rollout, rolloutGroup);
             } else {// not in error so check success condition and group completed
                 // 'success' is either group completed or success condition reached - execute 'success' Action
-                boolean groupCompleted = !(rolloutGroup == lastGroup && rolloutGroup.isDynamic()) && isRolloutGroupComplete(rollout, rolloutGroup);
+                final boolean groupCompleted = !(rolloutGroup == lastGroup && rolloutGroup.isDynamic()) && isRolloutGroupComplete(rollout, rolloutGroup);
                 checkSuccessCondition(rollout, rolloutGroup, evalProxy, rolloutGroup.getSuccessCondition(), groupCompleted);
                 if (groupCompleted) {
                     rolloutGroup.setStatus(RolloutGroupStatus.FINISHED);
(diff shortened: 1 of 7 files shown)


hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/scheduler/JpaRolloutExecutor.java
```diff
@@ -445,10 +445,11 @@ private void executeRunningGroups(final JpaRollout rollout, final List<JpaRollou
             if (isError) {
                 log.info("Rollout {} {} has error, calling error action", rollout.getName(), rollout.getId());
                 callErrorAction(rollout, rolloutGroup);
-            } else {
-                // not in error so check finished state, do we need to start the next group?
-                checkSuccessCondition(rollout, rolloutGroup, evalProxy, rolloutGroup.getSuccessCondition());
-                if (!(rolloutGroup == lastGroup && rolloutGroup.isDynamic()) && isRolloutGroupComplete(rollout, rolloutGroup)) {
+            } else {// not in error so check success condition and group completed
+                // 'success' is either group completed or success condition reached - execute 'success' Action
+                boolean groupCompleted = !(rolloutGroup == lastGroup && rolloutGroup.isDynamic()) && isRolloutGroupComplete(rollout, rolloutGroup);
```
avgustinmm: could make var final

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/scheduler/JpaRolloutExecutor.java
```diff
@@ -507,14 +506,14 @@ private boolean checkErrorState(final Rollout rollout, final RolloutGroup rollou
     }
 
     private void checkSuccessCondition(final Rollout rollout, final RolloutGroup rolloutGroup, final RolloutGroup evalProxy,
-            final RolloutGroupSuccessCondition successCondition) {
+            final RolloutGroupSuccessCondition successCondition, final boolean groupCompleted) {
         log.trace("Checking finish condition {} on rolloutgroup {}", successCondition, rolloutGroup);
         try {
-            final boolean isFinished = evaluationManager
+            final boolean isSuccessEvaluation = evaluationManager
```
avgustinmm: if groupCompleted == true - there is no need to evaluate success

hawkbit-repository/hawkbit-repository-jpa/src/test/java/org/eclipse/hawkbit/repository/jpa/rollout/condition/ThresholdRolloutGroupErrorConditionTest.java
```diff
@@ -0,0 +1,86 @@
+/**
```
avgustinmm, line 1: Not sure we need these tests. This entire test tests just "((float) error / (float) totalGroup) > ((float) threshold / 100F)"
My feeling is that this is just on big tests to test something trivial and not sure even it test border conditions - e.g. totalGroup == 0? 
Also other tests, as new RolloutManagementTest does actually provides coverage.

avgustinmm (changes requested):

avgustinmm (approved):

PR #3116, opened 2026-06-04
Fix target group assignment with RSQL negation filters (e.g. tag!=tag…

strailov:
…1) failing on EclipseLink/MySQL

- EclipseLink's UpdateAllQuery cannot handle NOT EXISTS subqueries with inherited @Id from @MappedSuperclass — generates invalid SQL (cross-join, missing ManyToMany join table)
- When RSQL contains negation (!=, =out=, =notlike=), use chunked SELECT+UPDATE approach that bypasses the EclipseLink bug
- Positive RSQL filters continue using the direct single-query CriteriaUpdate path
- Chunk size configurable via hawkbit.target-group.assign.chunk-size (default: 1000)

### DIFF
--- a/hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaTargetManagement.java
+++ b/hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaTargetManagement.java
@@ -29,18 +29,16 @@ import jakarta.persistence.criteria.CriteriaBuilder;
 import jakarta.persistence.criteria.CriteriaQuery;
 import jakarta.persistence.criteria.CriteriaUpdate;
 import jakarta.persistence.criteria.MapJoin;
-import jakarta.persistence.criteria.Predicate;
 import jakarta.persistence.criteria.Root;
-import jakarta.persistence.criteria.Subquery;
 import jakarta.persistence.metamodel.MapAttribute;
 import jakarta.validation.constraints.NotEmpty;
 
+import lombok.extern.slf4j.Slf4j;
 import org.eclipse.hawkbit.ql.jpa.QLSupport;
 import org.eclipse.hawkbit.repository.QuotaManagement;
 import org.eclipse.hawkbit.repository.TargetManagement;
 import org.eclipse.hawkbit.repository.exception.EntityAlreadyExistsException;
 import org.eclipse.hawkbit.repository.exception.EntityNotFoundException;
-import org.eclipse.hawkbit.repository.jpa.Jpa;
 import org.eclipse.hawkbit.repository.jpa.JpaManagementHelper;
 import org.eclipse.hawkbit.repository.jpa.acm.AccessController;
 import org.eclipse.hawkbit.repository.jpa.configuration.Constants;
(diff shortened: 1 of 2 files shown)


hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaTargetManagement.java
```diff
@@ -335,30 +338,62 @@ public Target unassignType(final String controllerId) {
...
+                    .setMaxResults(CHUNK_SIZE)
+                    .getResultList();
+
+            if (!chunk.isEmpty()) {
+                final CriteriaUpdate<JpaTarget> update = cb.createCriteriaUpdate(JpaTarget.class);
+                update.set("group", group);
+                update.where(update.getRoot().get(AbstractJpaBaseEntity_.ID).in(chunk));
+                entityManager.createQuery(update).executeUpdate();
+            }
+            offset += CHUNK_SIZE;
```
avgustinmm: is there granted db will return as much as possible up to CHUNK_SIZE? or db could decide to return less (as for instance InputStream with buffer works?) - maybe not
so better increment with the chunk.size() and exit if empty, not when != CHUNK_SIZE

avgustinmm (changes requested):

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaTargetManagement.java
```diff
@@ -335,30 +338,63 @@ public Target unassignType(final String controllerId) {
...
-            subquery.select(subRoot.get(AbstractJpaBaseEntity_.ID));
-            subquery.where(rsqlSpecification.toPredicate(subRoot, cb.createQuery(JpaTarget.class), cb));
-            criteriaUpdateQuery.where(updateRoot.get(AbstractJpaBaseEntity_.ID).in(subquery));
+        // Switch back to UpdateAllQuery if switching back to hibernate. (EclipseLink does not work well with UpdateAllQuery)
+        // EclipseLink: using subquery approach — applying predicate directly to the UPDATE root
+        // fails for NOT EXISTS due to UpdateAllQuery's @Id resolution bug
+        // BUG Reported: https://github.com/eclipse-ee4j/eclipselink/issues/2757
+        // Hibernate: applies predicate directly to the UPDATE root — Hibernate handles
+        // NOT EXISTS subqueries correctly in CriteriaUpdate context. So this problem does not exist there.
+        if (containsNegation(rsql)) {
```
avgustinmm:
```suggestion
        if (Jpa.JPA_VENDOR == Jpa.JpaVendor.ECLIPSELINK && containsNegation(rsql)) {
```

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaTargetManagement.java
```diff
@@ -335,30 +338,63 @@ public Target unassignType(final String controllerId) {
...
+        final CriteriaBuilder cb = entityManager.getCriteriaBuilder();
+        // SELECT Target IDs
+        final CriteriaQuery<Long> select = cb.createQuery(Long.class);
+        final Root<JpaTarget> root = select.from(JpaTarget.class);
+        select.select(root.get(AbstractJpaBaseEntity_.ID));
+        select.where(spec.toPredicate(root, select, cb));
+
+        List<Long> chunk;
+        int offset = 0;
+        final int CHUNK_SIZE = assignTargetGroupChunkSize;
```
avgustinmm: `CHUNK_SIZE` -> `chunkSize`, and it's not actually needed - you could use `assignTargetGroupChunkSize` directly

avgustinmm (approved):

PR #3125, opened 2026-06-11
Bump to Spring Boot 4.1.0

avgustinmm:


hawkbit-dmf/hawkbit-dmf-rabbitmq-test/src/main/java/org/eclipse/hawkbit/rabbitmq/test/AmqpTestConfiguration.java
```diff
@@ -67,4 +71,19 @@ ConnectionFactory rabbitConnectionFactory(RabbitMqSetupService rabbitMqSetupServ
     RabbitMqSetupService rabbitMqSetupService() {
         return new RabbitMqSetupService();
     }
+
+    // note - it MUST be the same as DmfApiConfiguration#messageConverter for the test to work properly (to test the real AMQP)
```
vasilchev, line 75: why duplicating it instead of using DmfApiConfiguration#messageConverter (change visibility)

avgustinmm: this dmf-rabbitmq-test has no dependency on dmf-amqp, it's the opposite

vasilchev (approved):

PR #3143, opened 2026-06-15
Move DMF message converter in amqp-api

avgustinmm:
- move in amqp to be in single place
- public DmfMessageConverter that could be used everywhere directly (instead of factory methods)
- defualt amqpMessageConverter bean renamed to dmfMessageConverter
- trusted packages configured (for dmfMessageConverter) with hawkbit.dmf.trusted-packages - default hwakbit dmf model package

### DIFF
--- a/hawkbit-dmf/hawkbit-dmf-api/src/main/java/org/eclipse/hawkbit/dmf/DmfMessageConverter.java
+++ b/hawkbit-dmf/hawkbit-dmf-api/src/main/java/org/eclipse/hawkbit/dmf/DmfMessageConverter.java
@@ -1,0 +1,45 @@
+/**
+ * Copyright (c) 2026 Contributors to the Eclipse Foundation
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.hawkbit.dmf;
+
+import org.jspecify.annotations.NonNull;
+import org.jspecify.annotations.Nullable;
+import org.springframework.amqp.core.Message;
+import org.springframework.amqp.support.converter.JacksonJsonMessageConverter;
+import tools.jackson.databind.json.JsonMapper;
+
+// - since spring-amqp 4.0.4 not all packages are assumed trusted for type converter (only java.land and java.util)
+// so e need to add hawkbit DMF model package (and eventual extension packages) as trusted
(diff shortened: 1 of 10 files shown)


hawkbit-sdk/hawkbit-sdk-dmf/src/main/java/org/eclipse/hawkbit/sdk/dmf/amqp/VHost.java
```diff
@@ -60,7 +60,7 @@ public VHost(final ConnectionFactory connectionFactory, final AmqpProperties amq
         // It is necessary to define rabbitTemplate as a Bean and set JacksonJsonMessageConverter explicitly here in order to convert only
```
vasilchev, line 60: "...Bean and set (JacksonJsonMessageConverter -> DmfMessageConverter).."

hawkbit-dmf/hawkbit-dmf-api/src/main/java/org/eclipse/hawkbit/dmf/DmfMessageConverter.java
```diff
@@ -0,0 +1,45 @@
...
+public class DmfMessageConverter extends JacksonJsonMessageConverter {
+
+    public static final String DMF_JSON_MODEL_PACKAGE = "org.eclipse.hawkbit.dmf.json.model";
+
+    public DmfMessageConverter(final String... trustedPackages) {
+        this(new JsonMapper(), trustedPackages);
+    }
+
+    public DmfMessageConverter(final JsonMapper jsonMapper, final String... trustedPackages) {
+        super(jsonMapper, trustedPackages.length == 0 ? new String[] { DMF_JSON_MODEL_PACKAGE } : trustedPackages);
```
vasilchev, line 32: shouldn't we append to and not replace the default with the configured trustedPackages :? 
otherwise default is dropped and we should explicitly add "org.eclipse.hawkbit.dmf.json.model" in the trustedPackages if we want to add something in addition to the default?

avgustinmm: this way there is more secure and flexibility - someone may decide for using this with different purposes, or somehow change packages or whatever ..

vasilchev: for me the main use case for this would be if someone adds additional packages with events outside hawkbit and still use. the core implementation and dmf listeners (just adding custom ones that are outside hawkbit api).
i don't see it as valid option to only use listeners(dmf core) but not use default DMF APIs or have refactored them completely without refactoring just the constant

PR #3149, opened 2026-06-16
Migrate to spring AI 2

denislavprinov:


hawkbit-mcp/hawkbit-mcp-starter/src/main/java/org/eclipse/hawkbit/mcp/server/tools/HawkbitMcpToolProvider.java
```diff
@@ -208,15 +196,7 @@ public PagedResponse<MgmtTargetFilterQuery> listTargetFilters(final ListRequest
...
-                    "UPDATE (modify existing target by controllerId), " +
-                    "DELETE (remove target by controllerId). " +
-                    "Use 'type' field to select operation: " +
-                    "{\"type\":\"Create\",\"body\":{\"controllerId\":\"id\",\"name\":\"name\"}}, " +
-                    "{\"type\":\"Update\",\"controllerId\":\"id\",\"body\":{...}}, " +
-                    "{\"type\":\"Delete\",\"controllerId\":\"id\"}")
+    @McpTool(name = "manage_target", description = "Create, update, or delete targets (devices). " + "Operations: CREATE (new target with controllerId, name, description. When creating a target without a specific target type, set \"targetType\": null), " + "UPDATE (modify existing target by controllerId), " + "DELETE (remove target by controllerId). " + "Use 'type' field to select operation: " + "{\"type\":\"Create\",\"body\":{\"controllerId\":\"id\",\"name\":\"name\"}}, " + "{\"type\":\"Update\",\"controllerId\":\"id\",\"body\":{...}}, " + "{\"type\":\"Delete\",\"controllerId\":\"id\"}")
```
avgustinmm: wouldn't it be more readable with: """ 
<multi line string>
""" ? And same for all JSONs? 
1. remove escaping (\") and
2. pretty json is easier to read and maintain
Even if we don't like new lines and indents (if this is somehow performance related, though new lines at least are not big overhead an somewhere are explicitly added) - at least for 1 i it is better to use """ string definition

avgustinmm (approved):

PR #3238, opened 2026-07-29
Allow RUNNING -> WAIT_FOR_CONFIRMATION via DDI 'denied' feedback

### LINKED ISSUE #3220: Allow action state change `RUNNING` → `WAIT_FOR_CONFIRMATION` via DDI `denied` feedback
### Summary

Extend the action state machine to permit a transition from `RUNNING` back to `WAIT_FOR_CONFIRMATION` so that a device can **revoke previously granted consent** — whether it was given automatically (auto-confirmation) or manually (explicit confirm) — for an action that has been confirmed but not yet installed (e.g. still waiting for a maintenance window).

Rather than introducing a new API, this **reuses the existing DDI confirmation feedback endpoint**:

```
POST /{tenant}/controller/v1/{controllerId}/confirmationBase/{actionId}/feedback
{
  "confirmation": "denied"
}
```

### Background / Current behavior

With the confirmation flow enabled, a new action is created in `WAIT_FOR_CONFIRMATION` (`OnlineDsAssignmentStrategy.createActionStatus`).

An action leaves `WAIT_FOR_CONFIRMATION` for `RUNNING` in two ways:
- **Manual consent:** `JpaConfirmationManagement.confirmAction` → `RUNNING`
- **Auto consent:** `JpaConfirmationManagement.autoConfirmAction` → `RUNNING`

The DDI feedback endpoint already maps `confirmation` values to management calls in `DdiRootController.postConfirmationActionFeedback`:
- `confirmed` → `confirmAction` (`WAIT_FOR_CONFIRMATION` → `RUNNING`)
- `denied` → `denyAction` (stays in `WAIT_FOR_CONFIRMATION`)

However, `denyAction` guards with `assertActionCanAcceptFeedback`, which throws `NOT_AWAITING_CONFIRMATION` unless the action is currently `WAIT_FOR_CONFIRMATION`. So a `denied` feedback on an already-`RUNNING` action is rejected with **`404 Not Found`** — there is currently no way to reverse consent once the action is `RUNNING`.

### Scenarios (the gap)

**A) Auto-consent revoked**
1. Confirmation flow enabled + auto-confirmation given → assigned action goes straight to `RUNNING`.
2. A **maintenance window** defers installation; the device is only notified it will install during the next window.
3. Before installing, the device revokes consent and sends `confirmation: "denied"` to `confirmationBase/{actionId}/feedback`.
4. **Expected:** the action returns to `WAIT_FOR_CONFIRMATION` and does not install during the maintenance window.

**B) Manual consent revoked**
1. Confirmation flow enabled; action starts in `WAIT_FOR_CONFIRMATION`.
2. Device sends `confirmation: "confirmed"` → action moves to `RUNNING`, installation deferred to the next **maintenance window**.
3. Before installing, the device changes its mind and sends `confirmation: "denied"` for the same action.
4. **Expected:** the action returns to `WAIT_FOR_CONFIRMATION` and does not install during the maintenance window.

**Actual (both cases):** the `denied` feedback is rejected with `404` and the action stays `RUNNING`, installing in the next maintenance window regardless of the revoked consent.

### Proposed change

1. Allow the state machine to transition an active action from `RUNNING` back to `WAIT_FOR_CONFIRMATION`.
2. Relax `denyAction` / `assertActionCanAcceptFeedback` so that a `denied` feedback is accepted for an active action in either `WAIT_FOR_CONFIRMATION` (unchanged) or `RUNNING` (reverts to `WAIT_FOR_CONFIRMATION`).
3. **No eligibility boundary.** The transition is driven entirely by the device's own `denied` feedback — the device is the authority on whether it has started installing, so if it reports a denial it is by definition still able to return to `WAIT_FOR_CONFIRMATION`. hawkBit does not need to gate this on maintenance-window timing or prior progress feedback.
4. No new endpoint, DTO, or management method signature — the existing `POST confirmationBase/{actionId}/feedback` with `"confirmation": "denied"` is reused for both the auto- and manually-confirmed cases.
5. Add an audit-trail `ActionStatus` entry documenting the revocation.

### Out of scope

- **Server-side (management-initiated) revoke.** Reverting consent from the server side (e.g. an operator/UI withdrawing a previously granted confirmation) is intentionally not covered here. It would require a new DDI **`revokeConfirmation`** flow — analogous to `cancelAction` on DDI — so the device can acknowledge/confirm the server-side revocation before the action is treated as re-armed. Tracked separately.

### Open questions / discussion

- **Audit trail wording:** Confirm the desired `ActionStatus` message for a `denied`-driven `RUNNING` → `WAIT_FOR_CONFIRMATION` revert, so it is clearly distinguishable from a first-time deny (e.g. "Target revoked previously granted confirmation. Action returned to confirmation-pending state.").

### State machine change

```mermaid
flowchart LR
    S(( )) -->|assign: confirmation flow on| WFC[WAIT_FOR_CONFIRMATION]
    S -->|assign: confirmation flow off| RUN[RUNNING]

    WFC -->|"(Auto-) Confirmed"| RUN
    RUN -->|"feedback &quot;denied&quot;<br/>revokes manual OR auto consent<br/>(NEW)"| WFC

    RUN -->|success| FIN[FINISHED]
    RUN -->|failure| ERR[ERROR]

    %% highlight only the new transition (edge index 3)
    linkStyle 3 stroke:#d6336c,stroke-width:3px,color:#d6336c
```

*(The `RUNNING --> WAIT_FOR_CONFIRMATION` edge is the new transition; all others already exist.)*

### Acceptance criteria

- [ ] The action state machine permits `RUNNING` → `WAIT_FOR_CONFIRMATION` for active actions.
- [ ] `POST confirmationBase/{actionId}/feedback` with `"confirmation": "denied"` reverts a `RUNNING` action (auto- or manually confirmed) to `WAIT_FOR_CONFIRMATION`, with no maintenance-window/progress gating.
- [ ] A descriptive `ActionStatus` entry records the revocation-driven transition, distinct from a first-time deny.
- [ ] No new DDI endpoint or management API method is introduced.
- [ ] Server-side revoke remains unimplemented and is explicitly out of scope.
- [ ] Tests cover: deny after auto-confirm (reverted to `WAIT_FOR_CONFIRMATION`), deny after manual confirm (reverted), subsequent base-resource poll offers `confirmationBase` again, and re-confirmation afterwards returns to `RUNNING`.

herdt-michael:
Extend the confirmation flow so a device can revoke a previously granted consent (auto- or manually confirmed) for an action that has not been installed yet, by sending 'confirmation: "denied"' to the existing confirmationBase/{actionId}/feedback endpoint.

- Relax denyAction guard so a 'denied' feedback is accepted for any active action, not only while WAIT_FOR_CONFIRMATION. An already-confirmed active action reverts to WAIT_FOR_CONFIRMATION with no maintenance-window or progress gating.
- Record a distinct ActionStatus audit-trail message for the revocation, separate from a first-time deny.
- No new DDI endpoint or management API method is introduced.

Tests cover deny after auto-confirm, deny after manual confirm, subsequent base-resource poll offering confirmationBase again, and re-confirmation returning to RUNNING.

Fixes eclipse-hawkbit/hawkbit#3220

### DIFF
--- a/hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaConfirmationManagement.java
+++ b/hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaConfirmationManagement.java
@@ -122,12 +122,19 @@ public class JpaConfirmationManagement extends JpaActionManagement implements Co
     public Action denyAction(final long actionId, final Integer code, final Collection<String> deviceMessages) {
         log.trace("Action with id {} deny request is triggered.", actionId);
         final Action action = actionRepository.getById(actionId);
-        assertActionCanAcceptFeedback(action);
+        assertActionCanBeDenied(action);
         final List<String> messages = new ArrayList<>();
         if (deviceMessages != null) {
             messages.addAll(deviceMessages);
         }
-        messages.add(SERVER_MESSAGE_PREFIX + "Target rejected action. Action will stay in confirmation pending state.");
+        if (action.getStatus() == Status.WAIT_FOR_CONFIRMATION) {
+            messages.add(SERVER_MESSAGE_PREFIX + "Target rejected action. Action will stay in confirmation pending state.");
+        } else {
+            // the target revokes a previously granted confirmation (auto- or manually confirmed) for an action that has not
+            // been installed yet - revert it back to WAIT_FOR_CONFIRMATION
+            messages.add(SERVER_MESSAGE_PREFIX
+                    + "Target revoked previously granted confirmation. Action returned to confirmation-pending state.");
+        }
         return addActionStatus(createConfirmationActionStatus(action.getId(), code, messages).status(Status.WAIT_FOR_CONFIRMATION).build());
     }
 
@@ -155,8 +162,13 @@ public class JpaConfirmationManagement extends JpaActionManagement implements Co
 
     @Override
     protected void onActionStatusUpdate(final JpaActionStatus newActionStatus, final JpaAction action) {
-        if (newActionStatus.getStatus() == Status.RUNNING && action.isActive()) {
-            action.setStatus(Status.RUNNING);
+        if (action.isActive()) {
+            if (newActionStatus.getStatus() == Status.RUNNING) {
+                action.setStatus(Status.RUNNING);
+            } else if (newActionStatus.getStatus() == Status.WAIT_FOR_CONFIRMATION) {
+                // a denied feedback reverts an already confirmed (RUNNING) action back to the confirmation-pending state
+                action.setStatus(Status.WAIT_FOR_CONFIRMATION);
+            }
         }
     }
 
@@ -173,6 +185,20 @@ public class JpaConfirmationManagement extends JpaActionManagement implements Co
             log.warn(msg);
             throw new InvalidConfirmationFeedbackException(
                     InvalidConfirmationFeedbackException.Reason.NOT_AWAITING_CONFIRMATION, msg);
+        }
+    }
+
+    private static void assertActionCanBeDenied(final Action action) {
+        // a denied feedback is accepted while awaiting confirmation (stays in confirmation-pending state) or for an
+        // already confirmed but not yet installed action, in which case the previously granted confirmation is revoked
+        // and the action reverts to the confirmation-pending state. As the device is the authority on whether it may
+        // still revoke consent, there is no maintenance-window or progress gating - only closed actions are rejected.
+        if (!action.isActive()) {
+            final String msg = String.format(
+                    "Denying action %s is not possible since the action is not active anymore.", action.getId());
+            log.warn(msg);
+            throw new InvalidConfirmationFeedbackException(InvalidConfirmationFeedbackException.Reason.ACTION_CLOSED,
+                    msg);
         }
     }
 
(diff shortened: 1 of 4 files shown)


hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaConfirmationManagement.java
```diff
@@ -122,12 +122,19 @@ public Action confirmAction(final long actionId, final Integer code, final Colle
     public Action denyAction(final long actionId, final Integer code, final Collection<String> deviceMessages) {
         log.trace("Action with id {} deny request is triggered.", actionId);
         final Action action = actionRepository.getById(actionId);
-        assertActionCanAcceptFeedback(action);
+        assertActionCanBeDenied(action);
```
strailov, line 126: If we send denied status for action in cancelling currently it will go to WFC... I wonder if this is appropriate.
Main reason is when you have cancelling on poll target recieves cancelAction, not confirmation base... Idk, maybe we should add a guard for cancelling state here ? Meaning - actions soft cancelled can't be denied and sent to WFC again. What do you think ?

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaConfirmationManagement.java
```diff
@@ -122,12 +122,19 @@ public Action confirmAction(final long actionId, final Integer code, final Colle
     public Action denyAction(final long actionId, final Integer code, final Collection<String> deviceMessages) {
```
strailov, line 123: Also what stops us currently to deny an action if we have not enabled the confirmation flow ? Previously there was a guard just because we checked the action status for WFC, but now we check nothing else ? This potentially could lead to inproper action state  - WFC but I have never enabled it.

strailov (changes requested):

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaConfirmationManagement.java
herdt-michael: Good catch, will disallow revoking actions in canceling state.

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaConfirmationManagement.java
herdt-michael: I'll add a condition to verify that the confirmation flow is enabled.

strailov (approved):

PR #3253, opened 2026-08-05
Auto assignments now don't get a status reset

kaloyanam:
When a target filter query gets only a name update, the auto assignment linked to it now doesn't get reset, but gets just an update. Also an empty update of a target filter query doesn't recreate the auto assignment


hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaTargetFilterQueryManagement.java
```diff
@@ -173,22 +173,29 @@ private void validate(final Update update) {
     }
 
     private void updateAutoAssignment(Update update) {
-        findLinkedAutoAssignment(update.getId()).ifPresent(autoAssignment -> {
+        final Optional<AutoAssignment> autoAssignment = findLinkedAutoAssignment(update.getId());
```
avgustinmm: Keep processing in ifPresent - cleaner

avgustinmm (changes requested):

avgustinmm (approved):

PR #3267, opened 2026-08-11
Hawkbit UI Auto Assignment view

kaloyanam:
## What 
Created an Auto Assignment View for the Hawkbit UI, removed obsolete fields from the Target Filter Query View. Also updated Rollouts View to directly update the status on action performed

## Changes
- Created an `AutoAssignmentView` based on the `RolloutsView`
- Updated `TargetFilterQueryView` so that it no longer has a link to auto assignments and does not use any deprecated methods
- Updated `RolloutView` so that on start/pause/resume the status as well as the icon changes without needing a refresh
- Added a permission for reading auto assignments
- Updated the `MgmtAutoAssignmentMapper` so it now includes createdAt/By and lastModifiedAt/By and maps them


avgustinmm (approved):
LGTM

PR #3287, opened 2026-08-13
Fix nondeterministic ordering with postgres on artifacts

strailov:


avgustinmm:
AFAIR, hawkbit we don't, in general, sort list/set collections. Sorting just this one - seems a little bit awkward. 
Lists are just collections. I think the tests shall be fixed to compare lists as sets - not as ordered lists. We don't have any sort contracts. Having MySQL and H2 sorting anyway doesn't imply the hawkbit does it.
Just maybe using Set instead of List will match better.

strailov:
Changed the tests, however not sure we should target Set instead of List here - not worth it IMHO, because there are more files to touch with potential bug introduction.

avgustinmm (approved):

PR #3304, opened 2026-08-18
Imrpoveme Assignments

vasilchev:
- Read DistributionSet type id directly (getTypeId()) instead of materializing the lazy DistributionSetType + its eager module-type elements
  - Retrieve confirmation-flow config once per assignment batch instead of per target/action — avoids an isConfirmationFlowEnabled() (tenant-config) lookup per item, which isn't cached for tenants with default/null values.
  - Drop the redundant per-target actions-quota check in the assignment path; already enforced upstream (batched) by enforceMaxActionsPerTarget()
  - JpaTarget.autoConfirmationStatus: remove CascadeType.DETACH (no need to load just to detach).


hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/model/JpaDistributionSet.java
```diff
@@ -71,6 +71,11 @@ public class JpaDistributionSet
     @NotNull
     private DistributionSetType type;
 
+    // read-only mapping of the ds_type FK, so the type id can be read without materializing the (LAZY) type entity
+    // and its EAGER elements collection - see getTypeId()
+    @Column(name = "ds_type", insertable = false, updatable = false)
+    private Long typeId;
```
avgustinmm, line 77: wouldn't lazy DistributionSetType type with static weaving + just getId work?

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/OnlineDsAssignmentStrategy.java
```diff
@@ -88,12 +82,13 @@ public JpaAction createTargetAction(final TargetWithActionType targetWithActionT
      */
     @Override
     public JpaActionStatus createActionStatus(final JpaAction action, final String actionMessage) {
+        return createActionStatus(action, actionMessage, isConfirmationFlowEnabled());
+    }
+
+    @Override
+    public JpaActionStatus createActionStatus(final JpaAction action, final String actionMessage, final boolean confirmationFlowEnabled) {
         final JpaActionStatus result = super.createActionStatus(action, actionMessage);
```
avgustinmm: AFAIS the createActionStatus without confirmation flow enabled is called only here - why not call with available confirmationFlowEnabled  and just remove shorter method?

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/AbstractDsAssignmentStrategy.java
```diff
@@ -272,12 +263,4 @@ private void cancelAssignDistributionSetEvent(final List<Action> actions) {
                 new CancelTargetAssignmentEvent(tenant, actions)));
     }
 
-    private void assertActionsPerTargetQuota(final Target target) {
```
avgustinmm, line 275: who now checks quota?

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/model/JpaDistributionSet.java
vasilchev: It seems that works only in Hibernate, EclipseLink loads the type entity to getId() to work

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/AbstractDsAssignmentStrategy.java
vasilchev: JpaDeployemntManagement#assignDistributionSets performs validateAndFilterRequestForAssignments --> enforceMaxActionsPerTarget -> checkMaxAssignmentQuota

So it was doing it twice per assignment

avgustinmm: yes, it seems so, please add commemt to createActions method that the caller is responsible for quota check

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/model/JpaDistributionSet.java
vasilchev: just tested it again, hibernate works fine.
Eclipse link even does 2 selects when getType() is called (it loads the type and its 'eager' elements dstype-to-smtypes)

- SELECT ..FROM sp_distribution_set_type WHERE ((id = ?)
- SELECT .. FROM sp_ds_type_element WHERE (distribution_set_type = ?)

avgustinmm (approved):

PR #3305, opened 2026-08-20
Add optional externalRef to distribution set assignment requests

herdt-michael:
MgmtDistributionSetAssignment and MgmtTargetAssignmentRequestBody were extended with an optional externalRef field. The value is propagated through TargetWithActionType/DeploymentRequest down to JpaAction creation in AbstractDsAssignmentStrategy, where it is set if provided. Instead of adding new tests, the existing weightValidation tests in MgmtDistributionSetResourceTest and MgmtTargetResourceTest were extended to verify that the field is persisted on the resulting action. Two direct DeploymentRequest constructor calls in JPA tests were adjusted to the new signature.


vasilchev:
Looks overall good. - couple of questions though

JpaDeploymentManagament#assignDistributionSets / JpaDeploymentManagement#offlineAssignedDistributionSets


AutoAssign/Rollout - does not set externalRef  :? 
offlineAssignedDistributionSets (used by MgmtTargetRestApi) - also does not seem to use the externalRef from the RestResource ?

Just notes (schedulers seem to be ok without, but offlineAssignedDistributionSets seem to be a valid case - maybe for a separate PR)

vasilchev (approved):

PR #3314, opened 2026-08-24
Fix dependencies and remove old overrides

vasilchev:


pom.xml
```diff
@@ -226,7 +221,6 @@
             <dependency>
```
avgustinmm, line 221: if we don't define here version - isn't this dependency management entry to be removed at all?

avgustinmm (changes requested):

pom.xml
vasilchev: why should be removed ? version is just not overriden it will take whatever (1.6.0) version comes from openfeign. (in 5.0.3 and also 5.0.2 i think) it is already 1.6.0

avgustinmm (approved):

pom.xml
vasilchev: yes, it is not needed anymore in dependency management, will remove

PR #3337, opened 2026-09-03
Target Group Quota introduction

strailov:
### Introduction

Limits the number of distinct target groups a tenant may create.
Quota check on each write case for target group - create/update (+bulk case) targets if groups is present and assigning group to target(s).
Default quota set to 100 with options to be increased/decreased. 

### Behaviour

  - `null` group never consumes quota — unassignment always allowed
  - Limit `<= 0` means unlimited
  - Assigning an **existing** group is always allowed, even for a tenant already over the limit,
    so introducing a limit never breaks existing tenants
  - Batches are asserted once against their distinct new groups, so `n + k` is checked rather
    than `n + 1` repeatedly
  - Rejection throws `AssignmentQuotaExceededException`
  - Existence is checked before the quota rejection is surfaced in `update`, so a bad id still
    reports `EntityNotFoundException`

### DIFF
--- a/hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaTargetManagement.java
+++ b/hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaTargetManagement.java
@@ -19,6 +19,8 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
+import java.util.SortedSet;
+import java.util.TreeSet;
 import java.util.function.BiFunction;
 import java.util.function.Consumer;
 import java.util.function.Function;
@@ -34,6 +36,7 @@ import jakarta.persistence.metamodel.MapAttribute;
 import jakarta.validation.constraints.NotEmpty;
 
 import lombok.extern.slf4j.Slf4j;
+import org.eclipse.hawkbit.context.AccessContext;
 import org.eclipse.hawkbit.ql.jpa.QLSupport;
 import org.eclipse.hawkbit.repository.QuotaManagement;
 import org.eclipse.hawkbit.repository.TargetManagement;
@@ -100,6 +103,50 @@ public class JpaTargetManagement
         this.quotaManagement = quotaManagement;
         this.targetTypeRepository = targetTypeRepository;
         this.targetTagRepository = targetTagRepository;
+    }
+
+    @Override
+    @Transactional
+    @Retryable(includes = ConcurrencyFailureException.class, maxRetriesString = Constants.RETRY_MAX, delayString = Constants.RETRY_DELAY)
+    public JpaTarget create(final TargetManagement.Create create) {
+        assertTargetGroupQuota(Collections.singletonList(create.getGroup()));
+        return super.create(create);
+    }
+
+    @Override
+    @Transactional
+    @Retryable(includes = ConcurrencyFailureException.class, maxRetriesString = Constants.RETRY_MAX, delayString = Constants.RETRY_DELAY)
+    public List<JpaTarget> create(final Collection<TargetManagement.Create> create) {
+        assertTargetGroupQuota(create.stream().map(TargetManagement.Create::getGroup).toList());
+        return super.create(create);
+    }
+
+    @Override
+    @Transactional
+    @Retryable(includes = ConcurrencyFailureException.class, maxRetriesString = Constants.RETRY_MAX, delayString = Constants.RETRY_DELAY)
+    public JpaTarget update(final TargetManagement.Update update) {
+        try {
+            assertTargetGroupQuota(Collections.singletonList(update.getGroup()));
+        } catch (final Exception ex) {
+            // target existence check in order to throw EntityNotFound instead of AssignmentQuotaException if both applicable
+            getValid(update.getId());
+            throw ex;
+        }
+        return super.update(update);
+    }
+
+    @Override
+    @Transactional
+    @Retryable(includes = ConcurrencyFailureException.class, maxRetriesString = Constants.RETRY_MAX, delayString = Constants.RETRY_DELAY)
+    public Map<Long, JpaTarget> update(final Collection<TargetManagement.Update> update) {
+        try {
+            assertTargetGroupQuota(update.stream().map(TargetManagement.Update::getGroup).toList());
+        } catch (final Exception ex) {
+            // target existence check in order to throw EntityNotFound instead of AssignmentQuotaException if both applicable
+            get(update.stream().map(TargetManagement.Update::getId).toList());
+            throw ex;
+        }
+        return super.update(update);
     }
 
     @Override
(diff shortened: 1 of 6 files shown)


hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaTargetManagement.java
```diff
@@ -102,6 +105,50 @@ protected JpaTargetManagement(
         this.targetTagRepository = targetTagRepository;
     }
 
+    @Override
+    @Transactional
+    @Retryable(includes = ConcurrencyFailureException.class, maxRetriesString = Constants.RETRY_MAX, delayString = Constants.RETRY_DELAY)
+    public JpaTarget create(final TargetManagement.Create create) {
+        assertTargetGroupQuota(Collections.singletonList(create.getGroup()));
```
avgustinmm, line 112: why here we don't check for group == null, but on some places we do check>?
same on update

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaTargetManagement.java
```diff
@@ -552,4 +608,36 @@ private void throwEntityNotFoundExceptionIfTagDoesNotExist(final Long tagId) {
             throw new EntityNotFoundException(TargetTag.class, tagId);
         }
     }
+
+    private void assertTargetGroupQuota(final Collection<String> requested) {
+        final long limit = quotaManagement.getMaxTargetGroups();
```
avgustinmm: if wanted is empty - no need to getMaxTargetGroups, which at some point could become db managed

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaTargetManagement.java
```diff
@@ -552,4 +608,36 @@ private void throwEntityNotFoundExceptionIfTagDoesNotExist(final Long tagId) {
...
+        if (limit <= 0) {
+            return;
+        }
+
+        final SortedSet<String> wanted = requested.stream()
+                .filter(Objects::nonNull)
+                .collect(Collectors.toCollection(() -> new TreeSet<>(String.CASE_INSENSITIVE_ORDER)));
```
avgustinmm: db could be case sensitive?

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaTargetManagement.java
```diff
@@ -552,4 +608,36 @@ private void throwEntityNotFoundExceptionIfTagDoesNotExist(final Long tagId) {
...
+
+        if (wanted.isEmpty()) {
+            return; // no group(s), skip findDistinctGroups db call
+        }
+
+        // one group, already present -> allowed
+        if (wanted.size() == 1 && jpaRepository.existsByGroup(wanted.first())) {
+            return;
+        }
+        final List<String> existing = jpaRepository.findDistinctGroups(AccessContext.tenant());
```
avgustinmm, line 626: it could be the case that updating a group other is removed, e.g.
update group x -> y, if device is the only in x - then x is removed y is created - no need to check.
I wonder if it will be better if we do a "relaxed" check - i.e. if there are already 100 groups - we don't allow not existing, though - this also could be wrong if we "remove" groups

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaTargetManagement.java
strailov: We actually do on later point - on wanted.isEmpty() in the assertTargetGroupQuota. It is possible to directly check in the create method, but I have decided the code would become more complex to read idk ...

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaTargetManagement.java
strailov: Fair point, will switch the order

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaTargetManagement.java
strailov: Ah, agreed - I'll let the db collation do the job without specifying in java for it

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaTargetManagement.java
strailov: Didn't quite get the second one - don't we actually currently do that ? We don't allow not existing groups (currently) when the limit is reached . 
Fair point for the first case with x and y, however not quite sure currently how could we handle it painlessly. 
Most of those cases some devices are left with group x which could lead to hitting the limit, however if exactly those devices that has x goes to y this is like more renaming and we won't work in that case, yes. Not sure if we should mitigate this - workaround is present, just unassign X and assign Y.

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaTargetManagement.java
strailov: So I guess the two guards in assign methods are redundant here ... Will remove them in order to comply with all the others.

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaTargetManagement.java
```diff
@@ -552,4 +604,34 @@ private void throwEntityNotFoundExceptionIfTagDoesNotExist(final Long tagId) {
...
+
+        existing.forEach(wanted::remove);
+        if (wanted.isEmpty()) {
+            return; // no growth -> allowed
+        }
+
+        QuotaHelper.assertAssignmentQuota(
```
avgustinmm, line 633: why use method with parrentId (when no such) instead of:
public static void assertAssignmentQuota(
            final long requested, final long limit,
            @NotNull final Class<?> type, @NotNull final Class<?> parentType) {
i.e. parent null?

avgustinmm (approved):
LGTM, but see comment - maybe better to call method without parentId

hawkbit-repository/hawkbit-repository-jpa/src/main/java/org/eclipse/hawkbit/repository/jpa/management/JpaTargetManagement.java
strailov: Because we won't get into : 
if (parentId != null && countFct != null) in the assertAssignmentQuota where we want to evaluate currentCount + requested > limit . Didn't want to modify the method just for one parameter, which just needs to be non null (and why not the parent to be the Tenant in the exception message).

PR #3344, opened 2026-09-09
Feature/action updated event extension

strailov:
Add status mark on action updated event creation + downloadOnly value in cases where status must be processed by event listeners and reloading the entity from database could cause collision.


vasilchev (approved):

denislavprinov (approved):
LGTM

PR #3350, opened 2026-09-11
fix: add OAuth2 to OpenAPI global security requirement

flobz:
securityRequirement.addList was calling BEARER_AUTH_SEC_SCHEME_NAME instead of OAUTH2_AUTH_SEC_SCHEME_NAME inside the tokenUrl block, so the OAuth2 scheme appeared in securitySchemes but not in the top-level security requirement — making it invisible in Swagger UI.

Do I need to create a PR to backport the fix to 1.1 branch ?


avgustinmm (approved):
LGTM, thanks for contribution

avgustinmm:
No need to create PR for 1.1. On next release we will update it with master commits.

PR #3362, opened 2026-09-17
Fix SM Metadata Duplicate values insertion

### LINKED ISSUE #3361: Can't store or delete metadata with same value
After upgrade to hawkbit 1.0.3 from 0.9.0, we can't remove metadata from software if it has the same value as any other metadata for same software. We use it to store a timestamp and we can store same timestamp for many keys. But those values were stored before the update of hawkbit, cause now we can't save metadata with same values.
The requests for both DELETE and POST responds with success but the metadata is not deleted or posted even though it responds with success. 
My guess that the constraint that should be on the metadata key is on the metadata value instead. Cause why shouldn't it be possible to have metadatas with same values?

Also if updating a metadata's value to the same value as any other metadata for same software, the metadata that should have been updated is deleted.

strailov:
For an @ElementCollection of an @Embeddable keyed by @MapKeyColumn, EclipseLink derives each element's identity from the embeddable's own mapped columns and excludes the map key. Two entries under different keys whose embeddable columns are equal therefore collapse onto one identity, and:

    adding a second such entry in a later transaction emits no INSERT — the commit succeeds silently and the entry is lost;
    removing one of two such entries emits no DELETE — the removal is silently discarded.

The same mapping with a basic value type (Map<String, String>, DirectMapMapping) behaves correctly, so this is specific to the AggregateCollectionMapping path.

Severity: silent data loss. The transaction commits without error and the application has no way to detect the loss except by re-reading.

This PR is a workaround for this problem by embedding the key into the mapping.


avgustinmm (approved):
LGTM, maybe it could however add comment in:
```java
public static class JpaMetadataValue implements MetadataValue {

        @Column(name = "meta_key", length = SoftwareModule.METADATA_KEY_MAX_SIZE, updatable = false)
        private String key;
``` 
why this field was added, with a ref to eclipselink issue (if one is created)

